PR #22, opened 2018-07-06
Partially fix #15.

### LINKED ISSUE #15: Pull in updates from sse2neon project
@hasindu2008 [recently contributed](https://github.com/jratcliff63367/sse2neon/pull/14) a bunch of NEON implementations to sse2neon.

SIMDe had many NEON implementations not in sse2neon, but it's quite possible the recent additions to sse2neon include functions for which we only have a portable implementation. It's also possible their implementations are faster.

Just need to go through the changes and see what we can use. Both projects are MIT.

jeffdaily:
Only looked at additions in https://github.com/jratcliff63367/sse2neon/pull/14/files.


nemequ:
Awesome. Merged, thank you!

I'm pretty sure this actually resolves #15; sse2neon hasn't had any activity other than that patch set.

PR #24, opened 2018-07-09
implement _mm_blendv_epi8

jeffdaily:
The SSE4.1 intrinsic only uses the sign bit to blend.  The neon bitwise select uses all bits.  The given mask in this case is compared less than zero to effectively copy the sign bit across all bits in each 8-bit lane.


jeffdaily:
The failed travis tests are unrelated to the code changed by this PR.

nemequ:
Indeed; I had Travis re-run that build and it worked :/

Merged, thanks.

I invited you to join the project so you don't need to use PRs (or wait for me to approve them).  The master branch is protected so you need to push to a branch first, but once CI is successful you should be able to merge into master.  If you're more comfortable using PRs feel free to do so; whatever you prefer.

PR #26, opened 2018-11-28
sse4.1: Add missing explicit cast to __m128i*

betajippity:
Explicitly casting to __m128* on this line appears to be necessary to compile using clang or gcc when SIMDE_SSE4_1_NATIVE is defined.

### DIFF
--- a/simde/x86/sse4.1.h
+++ b/simde/x86/sse4.1.h
@@ -987,7 +987,7 @@ SIMDE__FUNCTION_ATTRIBUTES
 simde__m128i
 simde_mm_stream_load_si128 (const simde__m128i* mem_addr) {
 #if defined(SIMDE_SSE4_1_NATIVE)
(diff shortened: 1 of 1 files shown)


nemequ:
Hmm… are you using a C++ compiler?

I don't mind the change (and C++ is supported), just trying to understand why the existing code is a problem.

nemequ:
Merged, thanks!

betajippity:
> Hmm… are you using a C++ compiler?
> 
> I don't mind the change (and C++ is supported), just trying to understand why the existing code is a problem.

Yup, this was with a C++ compiler on a C++ codebase that I was playing with porting over. So alas, the compiler complained about implicit casts from void*. Thanks for taking a look!

nemequ:
Okay, that explains it.  I'll look into compiling the tests with a C++ compiler in CI to avoid this in the future. Thanks for the info!

PR #28, opened 2018-12-18
sse2+sse3: fix some gcc casts away qualifiers warnings

marmeladema:
fix some gcc casts away qualifiers warnings when neon mode is enabled


nemequ:
Merged, thanks!

PR #32, opened 2018-12-21
sse4.1: add NEON implementation of _mm_extract_epi64

marmeladema:
This commit includes a fix when the selector is greater than 1.


nemequ:
Pushed as d69c330b2604c785a8c5ab9f8d4ca25e7a2ec6ca, thanks!

PR #34, opened 2018-12-21
sse4.1: add NEON implementation of _mm_extract_epi8

marmeladema:
This commit includes a fix when the selector is greater than 15.


nemequ:
Pushed as 422ed9cc55c88feaa0412e9f1511e10e60fd39ba, thanks!

Sorry it took me so long to get to all these… holidays :(

PR #35, opened 2018-12-22
sse2: add NEON implementation of _mm_srli_epi64 + fix emulated version

marmeladema:


nemequ:
Pushed as 5a0da0d4339cecdc3fedb68673254829f1ce7606, thanks!

PR #44, opened 2019-08-04
avx: add simde_mm256_mul_ps

sabarishbollapragada:


nemequ:
Looks good.  Merged, thanks!

I also added a test (8005d58a26edd62feeb8f6a885b1b887b4b0e67d).  It's was just a copy & paste (from [test/x86/skel.c:979](https://github.com/nemequ/simde/blob/6e3ba90605361069cf3422c44242c79171b87275/test/x86/skel.c#L979), then copy the generated case over from the output); no big deal, but if you're planning on adding more code moving forward (which would be awesome!) it would be great if you could include tests for each function. 

Sorry about CI failing; it looks like PGI got a bit more strict about alignment with 19.4, which is actually pretty nice since it helps catch bugs.  I improved the alignment code a bit (including adding a new `simde_assert_aligned`), it should work much better moving forward.

PR #49, opened 2019-11-29
Add some more AVX2 ops

mr-c:


nemequ:
Merged, thanks!

I rewrote some commit messages so the IDs won't match, but the code is the same.

mr-c:
Thank you @nemequ !

Any guidance for adding AVX-512 operations?

mr-c:
Oh, and could you use [`co-authored-by:`](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) if you re-write commits? Would be nice for future contributors to see their username in the contributors list :-)

mr-c:
nevermind, I was looking at the wrong commits :-D

PR #55, opened 2019-12-13
c++ compat for std::isnan;

mr-c:


nemequ:
FWIW, this explains why I don't want to do this: https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

PR #70, opened 2020-01-10
travis: use YAML block format for a better visualization.

junaruga:
Let me send pull-request about `.travis.yml`.
I think that the YAML block format makes us see `.travis.yml` like a real shell script.

Reference: https://en.wikipedia.org/wiki/YAML#Basic_components

How do you think?


nemequ:
Great idea.  Merged as cb13779d7a21e444cf37715c681b6b014d01223d, thanks!

PR #79, opened 2020-01-20
include libstdc++.so.6 for armhf

mr-c:
And make arm32 mandatory to pass, now that it does


nemequ:
Wow, awesome!  I can't believe arm32 was so close… last time I looked it was failing a bunch of tests, especially the conversion ones.  I'll have to bisect.

Thanks for putting the ctest stuff back in; I'll work on fixing the Windows build in a few minutes and merge this as soon as it's done.

Issue #72, opened 2020-01-14
Native types (or GCC-style vectors) for args/retval instead of unions


nemequ:
I'm thinking that it might be a good idea to pass native types (e.g., `__m128`, `int8x4_t`, etc.) or GCC-style vectors when possible instead of SIMDe's union.  This would cause a bit of cruft in most SIMDe functions since we'd need to convert to our union, but I think it would make things a bit more reliable.

* Compilers are less likely to move stuff in and out of vector registers.  Inlining takes care of most of this issue, but doesn't solve it completely for all code.
* Some compilers (*cough* MSVC) don't like passing around aligned unions, but seem to do okay with the native vectors.
* I'm hoping this also gets rid of some warnings from GCC that I don't see another way to silence (the ones about the ABI changing in some GCC version), but I'd need to verify that.

If I decide to go ahead this is going to be a pretty noisy change, but the sooner it happens the better.

nemequ:
I've started working on this in the `wip/native` branch.

So far I've only done MMX.  The patch is massive, but it actually wasn't as much work as I'd feared; most of it was done automatically with search and replace.

Unfortunately, one thing MMX had going for it is that I was using a different style for most of the functions that looked a bit like

```c
#if defined(SIMDE_MMX_NATIVE)
  return do_native_stuff();
#else
  simde__m64 r;

  #if defined(SIMDE_MMX_NEON)
    r.neon_iX = do_neon_stuff();
  #else 
    // Portable implementation
  #endif

  return r;
#endif
```

Whereas most of the other files use

```c
simde__m64 r;

#if defined(SIMDE_MMX_NATIVE)
  r.n = do_native_stuff();
#elif defined(SIMDE_MMX_NEON)
  r.neon_iX = do_neon_stuff();
#else
  // Portable implementation
#endif

return r;
```

The old MMX style is much closer to what we need going forward.  Converting is easily going to be the most time-consuming part of this issue since I can't think of a good way to automate it.

The good news is that the public API doesn't change unless people were accessing the union members.

nemequ:
MMX, SSE, and SSE2 are pretty much done; I need to figure out a few issues on MSVC (and s390x, but that has never worked; I'll look at that after everything else is done).

If anyone wants to play around with it be aware that ISA extensions I have not updated (i.e., currently everything other than MMX, SSE, and SSE2) will be completely unusable.

nemequ:
[...]

PR #81, opened 2020-01-28
Add missing parenthesis in _mm_bsrli_si128 and _mm_srli_si128 defines.

bddinechin:
Error showed up while using -DSIMDE_ENABLE_NATIVE_ALIASES .

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -698,8 +698,8 @@ simde_mm_bsrli_si128 (simde__m128i a, const int imm8) {
 #endif
 #define simde_mm_srli_si128(a, imm8) simde_mm_bsrli_si128(a, imm8)
 #if defined(SIMDE_SSE2_ENABLE_NATIVE_ALIASES)
-#  define _mm_bsrli_si128(a, imm8) SIMDE__M128I_TO_NATIVE(simde_mm_bsrli_si128(SIMDE__M128I_FROM_NATIVE(a), (imm8))
-#  define _mm_srli_si128(a, imm8) SIMDE__M128I_TO_NATIVE(simde_mm_bsrli_si128(SIMDE__M128I_FROM_NATIVE(a), (imm8))
+#  define _mm_bsrli_si128(a, imm8) SIMDE__M128I_TO_NATIVE(simde_mm_bsrli_si128(SIMDE__M128I_FROM_NATIVE(a), (imm8)))
+#  define _mm_srli_si128(a, imm8) SIMDE__M128I_TO_NATIVE(simde_mm_bsrli_si128(SIMDE__M128I_FROM_NATIVE(a), (imm8)))
 #endif
 
 SIMDE__FUNCTION_ATTRIBUTES


mr-c:
Thanks @bddinechin !

Though I'm surprised that the testing didn't get this. Which compiler do you use and what the version?

nemequ:
Merged, thanks!

The native aliases don't currently receive any testing regular testing; we'd have to duplicate all the tests cases and modify them to call the unprefixed versions, and we can't really just do something like

```c
#if defined(SIMDE_TEST_NATIVE_ALIASES)
#define SIMDE_FUNC(func) func
#else
#define SIMDE_FUNC(func) simde#func
#endif
```

Then just compile the same file once with `SIMDE_TEST_NATIVE_ALIASES` defined and once without because the native aliases will assume Intel's original types (e.g., `char` instead of `int8_t`) which would break a bunch of tests.

That said, I think a lot of the issues with the native aliases will disappear once #72 is finished, since we're getting rid of all the `SIMDE__M128I_TO_NATIVE` and `SIMDE__M128I_FROM_NATIVE` cruft; this should actually change to something like

```c
#  define _mm_bsrli_si128(a, imm8) simde_mm_bsrli_si128(a, imm8)
#  define _mm_srli_si128(a, imm8) simde_mm_bsrli_si128(a, imm8)
```

Which will make it a lot easier to spot stuff like this.

Anyways, please let us know if you find anything other problems with the native aliases; since we don't have tests for them bug reports like this are really valuable and deeply appreciated!

bddinechin:
I use a gcc 7.5 compiler targeted to the Kalray VLIW core (https://github.com/kalray). I contribute to this gcc target in the areas of instruction scheduling, if-conversion, and vector support. The SIMDE_ENABLE_NATIVE_ALIASES could be especially useful to compile x86 code with builtins on this core.

nemequ:
That's very cool!

Is there a native API for that, or are you supposed to use GCC-style vector extensions, OpenMP SIMD, OpenCL SIMD, etc.?

If there is something I can do in SIMDe to improve support (beyond using closing parenthesis when appropriate) I'd definitely be interested.

bddinechin:
For the Kalray VLIW core compiler using SIMDe, the current plan is to rely GCC extensions for vector types and operations and avoid architecture-specific builtins. In the coming days, I will compile representative simde/x86/{sse,avx}*.h functions and to check if our GCC retargeting is good at removing overhead (currently it seems not). This will help focusing on the most important areas of GCC vector support to retarget on our vliw architecture.

However, Kalray has plans to switch to LLVM as the C/C++ application compiler (today LLVM is only used to compile OpenCL-C), while keeping GCC for OS (Linux, FreeRTOS) porting.

So, to answer your question about SIMDe improvements, after we get positive results with GCC compiling, we will do the same with our LLVM port. Fortunately, LLVM understands the GCC vector types and operations.

For your information, I came to SIMDe while trying to compile and run the following implementation of Smith-Waterman (https://github.com/mengyao/Complete-Striped-Smith-Waterman-Library). It only uses 20 or so SSE2 builtins.

nemequ:
That's all very good to know, thanks.

> For the Kalray VLIW core compiler using SIMDe, the current plan is to rely GCC extensions for vector types and operations and avoid architecture-specific builtins.

If you don't mind me asking, what about OpenMP SIMD?  It's supported by GCC, clang, ICC, PGI, and even [MSVC](https://devblogs.microsoft.com/cppblog/simd-extension-to-c-openmp-in-visual-studio/) (as well as others), which makes it very attractive from a portability standpoint.  It's also much more expressive than GCC-style vectors, which is great for users but I assume means it's also significantly more effort to support.

> In the coming days, I will compile representative simde/x86/{sse,avx}*.h functions and to check if our GCC retargeting is good at removing overhead (currently it seems not).

I'm just landed a massive patch which may help eliminate a lot of that.  You might want to try the latest version.

> So, to answer your question about SIMDe improvements, after we get positive results with GCC compiling, we will do the same with our LLVM port. Fortunately, LLVM understands the GCC vector types and operations.

FWIW, if it's possible to get the compiler up and running on CI (we use Travis now, but I don't have any problem adding other providers) I'd be happy to add a build to at least make sure everything compiles.  If there is an emulator we could make sure all the tests pass, too.

bddinechin:
About OpenMP SIMD, I agree this is clearly the way to go. More generally, maturing OpenMP support in our compilers is a main objective, altough current plans are to do this on LLVM. I will double-check OpenMP SIMD on our GCC 7.5 and let you know.

About CI, I'm interested in knowing what code base can be used for testing, specifically for our case of interest: run sse- or avx- tuned C codes on non x86 core (like ours) with minimal/no changes to to the application source code. (in particular, without changing the builtin names). Finding such benchmarks appears more challenging than using industry performance benchmarks which are architecture-neutral.

nemequ:
> I will double-check OpenMP SIMD on our GCC 7.5 and let you know.

I'm less concerned with the *current* state.  We prefer GCC vector extensions when possible, but since they aren't nearly as expressive as OpenMP SIMD there are a lot of cases where we hit the OpenMP SIMD fallbacks.

It just seems like everyone keeps pushing for me to use GCC-style vectors in SIMDe and ignoring OpenMP SIMD, so I was worried OpenMP SIMD was being ignored on the compiler side, too.  Based on your comment that seems not to be the case.

> Finding such benchmarks appears more challenging than using industry performance benchmarks which are architecture-neutral.

FWIW, I've been thinking it might be nice to put together a meta-repository of projects which are using SIMDe (or forks of them which do) and a script to build them, then set it Travis CI to build them all weekly (or when manually triggered) with a fully up-to-date copy of SIMDe as a regression test.

I'm not sure exactly how you would need it be structured for your use case, but if something like that might be useful for you maybe we could talk about it?  Feel free to e-mail me (check the git log for my address) if you'd rather do that somewhere other than on this issue.

bddinechin:
Just a few questions:
- file x86/sse.h seems unbalanced because of extra ')' at line 2717:

  define _mm_sfence_ps()) simde_mm_sfence_ps()

- it seems _mm_alignr_epi8 is not seen under -DSIMDE_ENABLE_NATIVE_ALIASES:

__m128
  alignr(__m128 a, __m128 b)
  {
   return _mm_alignr_epi8(a, b, 4); 
  }

bddinechin:
Also seems unbalanced at sse3.h lines 249 and 328

  #  define _mm_alignr_epi8(a, b, count) simde_mm_alignr_epi8(a, b, count))

  #  define _mm_shuffle_pi8(a, b) (simde_mm_shuffle_pi8(a, b)

nemequ:
Thanks!

693af88911c150f85c90db2037cfefdc15666936 should fix the unbalanced parenthesis.  I also ran a quick (and very ugly) script that to catch similar issues:

```bash
TMPIFS="$IFS";
IFS=$(echo -en "\n\b");
for macro in `grep -oP '(?<=define )_mm.+$' *.h`; do
  nopen=$(echo "$macro" | grep -o '(' | wc -l);
  nclose=$(echo "$macro" | grep -o ')' | wc -l);
  if [ $nopen != $nclose ]; then
    echo $macro;
  fi;
done;
IFS="$TMPIFS";
```

Based on the fact that it found the same macros you mentioned I'm guessing you did something similar.  Hopefully that was all of them.

As for `_mm_alignr_epi8`, I think the problem is that it takes an `__m128i` not an `__m128`.  If I switch your example to `__m128i` it works for me.  Looking at the function I did see something weird, though, and removed that with fa9e50ee32ff6c5c65f28904f75c638f36335c02 but I don't think it had anything to do with your problem.  If it's still not working for you let me know.

bddinechin:
For _mm_alignr_epi8, I had to include "ssse3.h", not "sse3.h". Indeed the argument was __m128i too, I made the mistake while reducing to a test code.

I have tried compiling with -DSIMDE_ENABLE_OPENMP -fopenmp-simd with our GCC 7.5. The assembly source is the same except for label names and some static data.

PR #88, opened 2020-02-27
Define terms the first time we use them

mr-c:


nemequ:
Good idea.  Merged, thanks!

PR #89, opened 2020-02-29
Added compile time checks for sse.h

masterchef2209:


nemequ:
Sorry, it looks like they are failing due to problems with CodeCov.io, which has been extremely unreliable (and inaccurate) lately.  I've made some changes to the CodeCov YAML file (29eb028899fe5b7a1822275523bb20f221f11cdd) so hopefully this doesn't happen again in the future.  If it does I'll probably just remove CodeCov.io support.

As for the patch, it looks good.  I've rebased it on top of the CodeCov.io YAML changes and merged it as c90c7436e81294258f40e6c899754dee15c39ce6.

One very minor tweak: the first line of the commit messages should be in the form of "component: description of changes".  For example, I changed the commit message on your patch to "sse: add compile time range checks".

masterchef2209:
Thank You, i will keep the commit messages in correct format from next time.

PR #91, opened 2020-03-01
Added compile time checks

masterchef2209:
Added compile time checks in file avx2.h, sse2.h and sse4.1.h


nemequ:
This looks great, thanks!  I rebased and merged it.

Sorry about CodeCov still not working.  I'm pretty sure the problem is https://github.com/codecov/codecov-action/issues/29 and it sounds like that should be fixed soon.  I've added 97643fd4da14c83bc378976e84f7ddd8b97967b8 as a temporary work-around, so hopefully your next PR will either work correctly (if that issue is fixed), or at least not cause the whole build to break.

PR #92, opened 2020-03-03
Added implementation for _mm256_sub_epi64

masterchef2209:
added implementation of avx2 mm256_sub_i64. Files modified: avx2.h avx2.c.


nemequ:
Awesome, thanks!  The code is exactly right, great job.  It seems like you're picking this all up pretty quickly.

I changed the formatting a bit before merging; I think your editor is set to use 4 spaces for indentation, but SIMDe uses 2.  You can usually configure that on a per-project basis, but the process varies depending on what editor you're using.  Obviously that's less important than the code being correct, but if you could try to use 2 spaces for future patches I'd appreciate it :)

Merged as 59f5089e4b2194961534efabca027e1666f49e4f.

masterchef2209:
I didn't notice that, I will make sure it is 2 spaces from next time.
Thank You

Issue #93, opened 2020-03-04
document expectations in a CONTRIBUTING file


mr-c:
https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

- [ ] coding standard
- [ ] How to test locally (Is `-march=native` set?)
- [ ] link to relevant wiki pages

PR #94, opened 2020-03-04
Create .editorconfig to document coding standard

mr-c:
https://editorconfig.org/

Related to #93 

We could also add a shortcut using a tool that people could run locally or as part of a git submit hook

Options for that local reformatting include:
 1. `astyle -s2 $(find test simde -type f -name '*[ch]')` (also makes the brace style consistent, see http://astyle.sourceforge.net/astyle.html#_Basic_Brace_Styles for more options)
 2. [`uncrustify`](https://github.com/uncrustify/uncrustify) So many options!
 3. [`indent`](https://www.gnu.org/software/indent/)
 4. ['clang-format'](http://clang.llvm.org/docs/ClangFormat.html)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# Configuration file for EditorConfig
+# More information is available under http://EditorConfig.org
+
+# Ignore any other files further up in the file system
+root = true
+
+# Configuration for all files
+[*]
+# Enforce Unix style line endings (\n only)
+end_of_line = lf
+# Always end files with a blank line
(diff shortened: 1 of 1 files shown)


nemequ:
How did I not know about EditorConfig?!?  Merged, thanks!

I've never been able to get any formatter to do exactly what I want.  clang-format seems to get the closest, but there is still a huge diff which makes the code quite a bit uglier (IMHO).  At some point it might be worth it, but I don't think it would be now.

Issue #86, opened 2020-02-21
WASM SIMD implementations of SSE


nemequ:
We should add WASM simd128 implementations of as many SSE/SSE2/etc. functions as possible.

Some of the functions won't see much, if any, improvements since we already have GCC-style vector extension and OpenMP SIMD support.  The real benefit will be for the functions that can't use GCC-style vectors.  For example, saturated operations, min/max, etc.  And of course there will be a lot of cases where maybe the compiler can autovectorize and maybe it can't (especially if operating at -O2 instead of -O3), and explicitly calling the simd128 versions is more likely to result in vectorized versions.

The implementation shouldn't be too difficult.  We'll need to add a `v128_t` entry to the `simde_m128_private`, `simde_m128i_private`, `simde_m128d_private`, and of course properly detect WASM SIMD in simde-arch.h then use that in the headers like sse.h and sse2.h.  Other than that it should generally just be a matter of adding an `#elif defined(SIMDE_ARCH_WASM_SIMD128)` and an implementation.

Right now we don't have an emscripten build in CI; we'll have to figure out how to add it back in.  I removed it because emscripten was running out of memory (it works fine on my desktop with 24 GiB RAM, but kept getting killed on the little VM workres).  Adding in some optimization flags may help reduce the memory requirement; I hadn't done that because emscripten was generating an internal compiler error, but that may be fixed now (at least using tot, hopefully there will be a release soon); if not, I should be reporting them to emscripten so they can be fixed.

nemequ:
I just pushed a commit (fb6dca9804bef6861fe3a7f82be3d2a6d88fb509) which adds the necessary bits to SSE and implements one function (`_mm_add_ps`) as an example.

nemequ:
Emscripten is working again in CI; I think it was just a buggy release (1.39.7 probably) that was causing the memory issues.

Unfortunately attempting to enable WASM SIMD support (with `-msimd128 -s SIMD=1`) causes [an error](https://travis-ci.org/nemequ/simde/jobs/653760153#L620) when running the code.  I've [reported the issue](https://github.com/emscripten-core/emscripten/issues/10541) to the emscripten people.

nemequ:
The error has been fixed in more recent versions of V8 than Node.js uses.  Using latest d8 instead of nodejs to run the tests fixed the problem, and `-msimd128 -s SIMD=1` is now used on Travis for the emscripten build.

Now that they'll be properly tested I guess it's time to start adding implementations.

[...]

PR #95, opened 2020-03-05
Added implementation for _mm256_adds_epu16

masterchef2209:
added implementation for _mm256_adds_epu16


nemequ:
LGTM, thanks!  Merged as 4e033cab3dbe2172420b68a29f2eaa09445bbbf2.

Can you please rebase your branch on top of master?  That way I can merge your patches without rebasing, which would let us skip a CI build (the master branch is protected, so the exact commit ID has to run successfully in CI before a patch can appear there, and rebasing changes the ID).

It seems like you've got this down pretty well, maybe next time you'd like to try a more challenging function?  One the permute/permutevar ones, perhaps?  After that, maybe a WebAssembly implementation or two (#86)?  After that I think you're pretty much ready for GSoC (though, to be honest, I think you're ready now).

PR #97, opened 2020-03-06
Added implementation of mm256_permute4x64_epi64 + one minor correction

masterchef2209:
Added implementation of mm256_permute4x64_epi64 + one minor correction in compile time check of other function, imm8 should be [0,255] instead of [0,15]


nemequ:
Merged, thanks!

PR #99, opened 2020-03-07
Reuse of simde_mm_move_sd to simplify simde_mm_*_sd functions.

himanshi18037:


nemequ:
You're almost there.  The failures are just because you're using [`vceqq_u64`](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vceqq_u64) everywhere; you just need to change that to the right function for the operation you're implementing; for example, in `simde_mm_comige_sd` you should be using `vcgeq_f64`, `simde_mm_comigt_sd` should use `vcgtq_f64`, etc.

The only complication is that NEON doesn't have a not-equal function.  So you could either use do something like `return vgetq_lane_u64(vreinterpretq_u64_u32(vmvnq_u32(vreinterpretq_u32_u64(vceqq_f64(a_.neon_f64, b_.neon_f64)))), 0) ? 1 : 0;` or just `return !vgetq_lane_u64(vceqq_f64(a_.neon_f64, b_.neon_f64), 0);`.  I'd suggest the latter.

nemequ:
Merged as 09324165a2c35e7cf50c8c9a716da75dfdff77e6 and 85bc31f2659304a92f915d4c879aa1c15afa09bc.  Thanks for all your work on this!

PR #100, opened 2020-03-07
Added WASM implementation for mm_sub_ps and mm_mul_ps

masterchef2209:
added WASM implementation for mm_sub_ps and mm_mul_ps


nemequ:
Merged, thanks!

PR #103, opened 2020-03-09
Implementing avx2 functions _mm256_max_epu16 and _mm256_srli_epi32

himanshi18037:


mr-c:
Huh, I don't see why the code coverage dropped, do you @nemequ ?

nemequ:
CodeCov has been extremely unreliable lately.  The osx build has been unable to upload for weeks (I finally filed a support request this morning), and others seem to fail sporadically.

Even when it is working the coverage numbers are way lower than they should be. I think OpenMP SIMD messes things up because there are a lot of cases where a for loop's conditions aren't hit but the contents are, even with --coverage.  And I think NEON intrinsics aren't counted (due to lack of an "artificial" attribute).

nemequ:
Merged, thanks!

@mr-c, I'll try to shore up coverage analysis this week.. If I can't get codecov to behave maybe coveralls will work better…

PR #105, opened 2020-03-10
Added WASM implementation for SSE functions mm_add_ps and mm_andnot_ps

masterchef2209:
Added WASM implementation for SSE functions mm_add_ps and mm_andnot_ps

### DIFF
--- a/simde/x86/sse.h
+++ b/simde/x86/sse.h
@@ -292,6 +292,8 @@ simde_mm_and_ps (simde__m128 a, simde__m128 b) {
 
 #if defined(SIMDE_SSE_NEON)
   r_.neon_i32 = vandq_s32(a_.neon_i32, b_.neon_i32);
+#elif defined(SIMDE_SSE_WASM_SIMD128)
+  r_.wasm_v128 = wasm_v128_and(a_.wasm_v128, b_.wasm_v128);
 #elif defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
   r_.i32 = a_.i32 & b_.i32;
 #else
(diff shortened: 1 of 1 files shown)


nemequ:
The `_mm_and_ps` patch works perfectly, thanks!  Merged as afd9331cdabb0c4bce8e6a3857aea12c0ec0b152

`wasm_v128_andnot` is hidden behind an `#ifdef __wasm_unimplemented_simd128__` macro in the emscripten header.  AFAICT that's used for functions which are just missing SIMD implementation in LLVM, but they still work by doing basically the same things we do in the portable versions.

I assume the primary reason they are ifdef'd out is to prevent people from accidentally using slow functions, in which case I don't see any reason to disable them in SIMDe… there is no advantage, and if we define `__wasm_unimplemented_simd128__` at least SIMDe automatically gets the faster implementation when it becomes available.

I added 4296375d32ff14a3ed762b44c2fa154c7114d550 to define `__wasm_unimplemented_simd12__` prior to including `wasm_simd128.h`, so if you rebase on top of that `wasm_v128_andnot` will be available.

That said, the patch isn't quite right.  Intel's andnot is defined as `~a & b`, but WASM is `a & ~b` (which seems much more sensible to me… it's "andnot" not "notand"), so the tests will fail.  I think the easiest way to fix that would be to change the patch to just swap the arguments (`wasm_v128_andnot(b_.wasm_v128, a_.wasm_v128)` instead of `wasm_v128_andnot(a_.wasm_v128, b_.wasm_v128)`).

For what it's worth, if you want to test these locally in the future, the procedure is

 * [Install emscripten](https://emscripten.org/docs/getting_started/downloads.html)
 * Install d8 (I use [jsvu](https://github.com/GoogleChromeLabs/jsvu))
 * Build SIMDe with something like `emcmake cmake .. -DCMAKE_{C,CXX}_FLAGS='-msimd128 -s SIMD=1 -O3' && make -j && ~/.jsvu/v8 --experimental-wasm-simd run-tests.js` (assuming you're in path/to/simde/test/build)

PR #106, opened 2020-03-10
sse2: fix "comparison of integer expressions of different signedness" warning

gh2o:


nemequ:
Nice.  Merged, thanks!

FWIW, I fixed a bunch of similar warnings yesterday (the patch was sitting in staging), so there should be much fewer diagnostics tripped going forward.

PR #107, opened 2020-03-11
Fix link to NEON wiki

Un1Gfn:


Un1Gfn:
Sorry but is [ci skip] goint to work if specified in commit edit?

nemequ:
Merged, thanks!

I'm not sure if [ci skip] works or not (or on which CI providers; we use Travis, AppVeyor, and GitHub Actions), but the master branch is protected so we can't actually merge anything into it without verifying it in CI first, so it needs to be checked anyways.

PR #109, opened 2020-03-11
Added implmentation of _mm256_slli_epi16 and _mm256_slli_epi64 in sim…

himanshi18037:


nemequ:
NIce, it looks like you found a bug in all of the `simde_mm256_slli_*` functions.

If the number of bits specified was greater than the width of the types, we were shifting by 0 bits (effectively a noop).  However, according to Intel's documentation it just shifts by the least significant 8 bits of `imm8` without regard to the width of the type.  In other words, shifts of more than 16 bits should result in zeros, not the original value.

The other tests were shifting by values inside of the expected range, so we didn't catch this.

I fixed the issue in all of the functions, and modified the `simde_mm256_slli_epi16` test case to shift by all values in the range [0,17], which is where things get interesting… compilers behave differently.  Some of them seem to just mask off everything outside of the expected range, some will return 0 (gcc and clang, for example, just generate an xor of the register with itself).  Intel's Intrinsics Guide is somewhat ambiguous.  I'm just going to require input in the range all compilers can handle.

The portable version is problematic anyways since shifting a signed integer by more bits than it contains is undefined behavior.  We would have to choose an implementation (either 0, or mask off the more significant bits), and it's really not clear which is correct.

After digging into the [Intel® 64 and IA-32 ArchitecturesSoftware Developer’s Manual](https://software.intel.com/en-us/articles/intel-sdm) a bit, I think the operation may actually be undefined.  It is defined as 0 for PSLLW, but I don't see anything for VPSLLW.  SAL says the range is limited to [0,31] for 32-bit mode and [0,63] for 64-bit mode, but since there are no 16-bit registers that's not really much help, either.  I think the right thing is to just treat it as undefined behavior.

Anyways, thanks for calling my attention to an interesting problem :)

nemequ:
Oh, and I merged your patch before tweaking the behavior.  Thanks :)

himanshi18037:
> 
> Anyways, thanks for calling my attention to an interesting problem :)

Even  I am amused because I was not able to think that it would be the reason . Thanks for making me understand it.

PR #117, opened 2020-03-24
enhanced README

FaresSalem:


nemequ:
Merged as d47deb23aeeeeef40011826f9967a92b3172cfd3, thanks!

I slightly changed the bit about `-mfpu=neon` since that hasn't really been working for me lately.  I'm not sure if GCC changed the flag, if that's just for armv7, etc., but I've had to use `-march=armv8-a+simd` lately.  It was definitely a good idea to put in that link, though, thanks for that :)

FaresSalem:
When I have commits on my fork and open a PR but like this one you made some edits to the same lines, which created a conflict when I tried synchronizing my fork, should that be solved by some way like reviewing or adding comments to my PR before closing it so I can change them?

nemequ:
There are two schools of thought on this.  The more git-like way is to use `git merge branch-to-merge` to merge upstream changes into your branch.  When you do this and there there is a conflict git will walk you through resolving it; basically you edit the conflicting files to look like you want them to going forward, then do a `git add conflicted-file && git merge --continue`, and git creates a "merge commit" with those changes.

However, I prefer a more linear history without merge commits, so for SIMDe what you should do is use `git rebase` instead of `git merge`.  The procedure is basically the same, but instead of creating a merge commit you remove conflicts from each commit, and git will actually modify your commits instead of leaving the history completely intact.  If you use the interactive mode (`git rebase -i branch-to-rebase-on-top-of`) You can also use this to do more advanced stuff, such as squashing commits (combining multiple commits into a single one), rewording commit messages, reordering commits, etc.

The process tends to be a bit different each time, so you it's hard to teach someone; you really have to learn by doing.  I try to keep contributing as easy as possible, so I usually just handle it myself instead of asking people to rebase (or merge).  If all your changes are full merged, it's generally best to just create a new branch, or do a hard reset and a forced push.

If you want to learn, I'd be willing to walk you through any examples that come up.

PR #119, opened 2020-03-24
Updating _mm512_xxx_epu8 and _mm512_xxx_epu16 functions in test/x86/s…

himanshi18037:
Updating _mm512_xxx_epu* functions in test/x86/skel.c.
Adding x_mm512_set_epu* implementations in simde/x86/avx512f.h
Adding mm512_adds_epu* implementations in simde/x86/avx512bw.h and test cases in test/x86/avx512bw.h


nemequ:
As we discussed, I spent a bit of time looking into that ICC error.  Maybe I'm missing something, but it looks like a bug in their compiler.  I've reported the issue, hopefully they fix it soon.

Meanwhlie, I've merged your patches and added a few of my own on top to work around the ICC bug.

PR #123, opened 2020-03-28
Drop an executable bit of simde/x86/avx2.h.

junaruga:
I found a header file that has a executable bit.

```
$ find simde/ -type f -executable | xargs ls -l
-rwxrwxr-x 1 jaruga jaruga 59236 Mar 28 15:44 simde/x86/avx2.h
```

I dropped it. :)

```
$ chmod -x simde/x86/avx2.h

$ ls -l simde/x86/avx2.h
-rw-rw-r-- 1 jaruga jaruga 59236 Mar 28 15:44 simde/x86/avx2.h
```


nemequ:
Merged, thanks!

PR #125, opened 2020-03-29
Added implementation of mm_unpacklo_epi32 and mm_unpacklo_epi64 in avx2

masterchef2209:
Added implementation of mm_unpacklo_epi32 and mm_unpacklo_epi64 in avx2


simde/x86/avx2.h
```diff
@@ -1844,6 +1844,32 @@ simde_mm256_unpacklo_epi16 (simde__m256i a, simde__m256i b) {
...
+#if defined(SIMDE_AVX2_NATIVE)
+  return _mm256_unpacklo_epi32(a, b);
+#else
+  simde__m256i_private
+    r_,
+    a_ = simde__m256i_to_private(a),
+    b_ = simde__m256i_to_private(b);
+
+#if defined(SIMDE__SHUFFLE_VECTOR)
+  r_.i32 =SIMDE__SHUFFLE_VECTOR(32, 32, a_.i32, b_.i32,
```
nemequ: There should be a space after the =

simde/x86/avx2.h
```diff
@@ -1844,6 +1844,32 @@ simde_mm256_unpacklo_epi16 (simde__m256i a, simde__m256i b) {
...
+                                0, 8, 1, 9, 4, 12, 5, 13);
+#else
+  r_.m128i[0] = simde_mm_unpacklo_epi32(a_.m128i[0], b_.m128i[0]);
+      r_.m128i[1] = simde_mm_unpacklo_epi32(a_.m128i[1], b_.m128i[1]);
```
nemequ: These two lines should line up.

simde/x86/avx2.h
```diff
@@ -1870,6 +1870,32 @@ simde_mm256_unpacklo_epi32 (simde__m256i a, simde__m256i b) {
...
+#if defined(SIMDE_AVX2_NATIVE)
+  return _mm256_unpacklo_epi64(a, b);
+#else
+  simde__m256i_private
+    r_,
+    a_ = simde__m256i_to_private(a),
+    b_ = simde__m256i_to_private(b);
+
+#if defined(SIMDE__SHUFFLE_VECTOR)
+  r_.i64 =SIMDE__SHUFFLE_VECTOR(64, 32, a_.i64, b_.i64,
```
nemequ: There should be a space after the =.

simde/x86/avx2.h
```diff
@@ -1870,6 +1870,32 @@ simde_mm256_unpacklo_epi32 (simde__m256i a, simde__m256i b) {
...
+                                1, 5, 3, 7);
```
nemequ: These values are wrong, which is why CI is failing.  Perhaps you forgot to add `-march=native` to your CFLAGS/CXXFLAGS, so only the fallbacks were running on your machine?  The easiest way to do that is to call cmake like `cmake .. -DCMAKE_{C,CXX}_FLAGS='-march=native'`.

You'll also need to regenerate the test cases, and make sure you copy the output from `/x86/avx2/mm256_unpacklo_epi64/native/c` or `/x86/avx2/mm256_unpacklo_epi64/native/cpp`, *not* `/x86/avx2/mm256_unpacklo_epi64/emul/c` or `/x86/avx2/mm256_unpacklo_epi64/emul/cpp`.  The former are the result of calling `_mm256_unpacklo_epi64`, the latter are the result of calling the portable version.

Also, this function call is short enough that it can go on one line.  That's really not a big deal; I'd normally probably just accept the patch, but since it needs some work anyways it would be nice if you could change that, too :)

simde/x86/avx2.h
```diff
@@ -1870,6 +1870,32 @@ simde_mm256_unpacklo_epi32 (simde__m256i a, simde__m256i b) {
...
+#else
+  r_.m128i[0] = simde_mm_unpacklo_epi64(a_.m128i[0], b_.m128i[0]);
+      r_.m128i[1] = simde_mm_unpacklo_epi64(a_.m128i[1], b_.m128i[1]);
```
nemequ: These two lines should line up.

simde/x86/avx2.h
masterchef2209: I will make sure to check proper indentation from next time. Usually my IDE collapse parts of code, so I may have missed it.

simde/x86/avx2.h
masterchef2209: I will be more careful from next time.

simde/x86/avx2.h
masterchef2209: Yes, I the flags native flag was not check and I checked the didn't notice that the tests which were passing were emul and not native. Also I made a bug and used indexes of unpackhi instead of unpacklo. I will be more careful from next time.

masterchef2209:
I have resolved all bugs in new commit. Please review.
Thank You

nemequ:
Merged, thanks!

Great job on this, BTW.  The shuffle stuff can be a bit disorienting until you get used to it, but you obviously figured it out pretty quickly.

PR #126, opened 2020-03-29
Add trivial SSE* fallbacks to avx.h

rosbif:
Add trivial SSE2 fallback for simde_mm256_setzero_si256.
Add trivial SSE4.1 fallback for simde_mm256_testz_si256.

Newbie warning: my first GitHub pull request.

My application makes heavy use of certain functions and runs very slowly on old machines without these fallbacks and other AVX2, AVX512F and AVX512BW ones that I shall send if all goes well.

### DIFF
--- a/simde/x86/avx.h
+++ b/simde/x86/avx.h
@@ -427,11 +427,15 @@ simde_mm256_setzero_si256 (void) {
   return _mm256_setzero_si256();
 #else
   simde__m256i_private r_;
-
+#if defined(SIMDE_ARCH_X86_SSE2)
+  r_.m128i[0] = simde_mm_setzero_si128();
+  r_.m128i[1] = simde_mm_setzero_si128();
+#else
   SIMDE__VECTORIZE
   for (size_t i = 0 ; i < (sizeof(r_.i32f) / sizeof(r_.i32f[0])) ; i++) {
     r_.i32f[i] = 0;
   }
+#endif
 
   return simde__m256i_from_private(r_);
 #endif
@@ -5165,17 +5169,21 @@ simde_mm256_testz_si256 (simde__m256i a, simde__m256i b) {
 #if defined(SIMDE_AVX_NATIVE)
   return _mm256_testz_si256(a, b);
 #else
-  int_fast32_t r = 0;
   simde__m256i_private
     a_ = simde__m256i_to_private(a),
     b_ = simde__m256i_to_private(b);
 
+#if defined(SIMDE_ARCH_X86_SSE4_1)
+  return simde_mm_testz_si128(a_.m128i[0], b_.m128i[0]) && simde_mm_testz_si128(a_.m128i[1], b_.m128i[1]);
+#else
+  int_fast32_t r = 0;
   SIMDE__VECTORIZE_REDUCTION(|:r)
   for (size_t i = 0 ; i < (sizeof(a_.i32f) / sizeof(a_.i32f[0])) ; i++) {
     r |= a_.i32f[i] & b_.i32f[i];
   }
 
   return HEDLEY_STATIC_CAST(int, !r);
+#endif
 #endif
 }
 #if defined(SIMDE_AVX_ENABLE_NATIVE_ALIASES)


rosbif:
I guess that my use case doesn't fit in with yours (or at least with your test philosophy).

Let me explain my use case:-
I have SSE4.2, AVX2 and AVX512 versions of my software.
I do much of my work on an old Arrandale laptop which only supports up to SSE4.2.
I wish to test my AVX2 and AVX512 versions efficiently on this laptop.
Without these and other fallbacks for AVX2 and AVX512, my software runs a good order of magnitude more slowly.

mr-c:
Hello @rosbif , thank you for your PR. It looks fine to me, might just be a shortcoming of our testing infrastructure. I'll let @nemequ do the final evaluation to confirm

rosbif:
Thank you Michael for your thumbs up.
I also have modified versions of avx2.h, avx512f.h and avx512.h.
I sent avx.h first as it is by far the simplest.
If it is accepted I shall prepare the other PRs.
They are more complicated as they contain 
- fallbacks for AVX2 and SSE
- typo corrections
- additional functions

nemequ:
Hi @rosbif!  Thanks for the patch!  I've committed it with some minor modifications for `simde_mm256_testz_si128`; I wanted to keep the return value outside of that ifdef so it's available when we start adding other optimizations.  I also made some minor improvements to the test case to exercise some additional paths, and modified the commit message to match [SIMDe's style](https://github.com/nemequ/simde/blob/master/CONTRIBUTING.md#commit-messages).

What compiler version and flags are you using?  I don't actually see [any difference](https://godbolt.org/z/vbEi_C) at `-O3` (which you should really be using for SIMDe, along with `-fopenmp-simd -DSIMDE_ENABLE_OPENMP`. See the ["Usage"](https://github.com/nemequ/simde#usage) section in the README).  

Developing for an architecture or ISA extension your hardware doesn't natively support is very much a use case SIMDe is intended to address; I'm glad someone is using it for this!  Please let me know if you run into any problems.

I'm actually planning on making some changes to SIMDe that would conflict with patches like this, so please submit your other patches ASAP and I'll merge them first.  Note that AVX-512 should call 256-bit versions, not directly call 128-bit versions unless they aren't equivalent.  Also, Intel's compiler is buggy here; you can't reliably use a loop.

Historically I didn't want to just have 256-bit versions call 128-bit versions twice since there is a good chance it will break autovectorization if the compiler actually supports 256-bit vectors but not the ISA in question (for example, ARM SVE).  This shouldn't really be a problem with your patch since it will only call the SSE2/SSE4.1 versions if the compiler has enabled thos extensions, the problem is that if we do it this way we can't take advantage of optimized 128-bit versions for other targets (NEON, AltiVec, WASM SIMD, etc.).

My plan is to add a preprocessor macro which is defined to the length of the longest vector type your CPU supports natively.  So, for Intel CPUs without AVX that would be 128, with AVX it would be 256 (or 512 with AVX-512).  Then, instead of checking to see if `SIMDE_ARCH_X86_SSE2` is defined, we check to see if `SIMDE_NATIVE_VEC_LENGTH <= 128` and, if so, call the SSE2 version.  This would let us take advantage of manual optimizations in SSE2 for other other architectures, but still give the compiler room to autovectorize when the native vector length.

Merging conflicts wouldn't be too difficult, but if you already have the work done I'd rather just get that in place first and avoid the conflicts altogether.

rosbif:
Thank you Evan for your reply and for your great work.

I use the following compilers:-
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)

I don't have Intel's compiler and I haven't been tempted to find it as I have had very bad experiences with it in the past. But that was about 25 years ago. If you don't have confidence in your compiler you end up suspecting it for all your bugs. Do you have to pay for it?

I prefer clang which always gives faster code with identical flags, generally up to 50% but in some cases nearly three times as fast!
I generally use -Ofast which includes all of -O3 and more.
I have tried with -fopenmp-simd -DSIMDE_ENABLE_OPENMP but I have observed no improvement.

The intrinsics which really kill performance without explicit hand coded fallbacks are _mm256_shuffle_epi8() and _mm512_shuffle_epi8().
I use these very heavily.
I don't see how any compiler could find these from the C equivalent.
Also _mm512_mask_cmpeq_epu8_mask().
But I found that explicit hand coded fallbacks always improve performance.

I shall try to submit my other PRs today.
(Unfortunately my laptop has failed with a PSU problem which I haven't yet resolved.)
Can I include new functions as I probably shan't have time to generate test vectors?

nemequ:
Ah yes, the shuffle intrinsics are tough to autovectorize.  However, if you really want to speed that up the best thing would be to use SIMDE__SHUFFLE_VECTOR (which uses __builtin_shufflevector on clang, or __builtin_shuffle on gcc).

> Can I include new functions as I probably shan't have time to generate test vectors?

Can you make those a separate PR?  I don't want to merge them until the have tests, but if you submit the implementations I can add tests to go with them.

Issue #75, opened 2020-01-16
Add clang on arm32 to CI


nemequ:
Clang tends to do a much better job of warning (well, erroring) about out-of-range values being passed.  This can be a bit annoying because the checks are done before dead code elimination (so you can't do something like `(imm8 & ~15) ? vdup_n_s16(0) : vshl_s16(imm8)`), but it also catches a *lot* of bugs.

For now clang-9 is working on arm32, but I'd definitely like to get a CI build going before I start working on more ARM implementations.

@junaruga, maybe you happen to know of an example we could use?

junaruga:
> @junaruga, maybe you happen to know of an example we could use?

Hi, I am happy to assist this.
Thank you for explaining the context to use clang.

I do not know the exact example for "clang-9 on arm32".
But I know some clues.

## clang-9 and clang-8

clang-8 on amd64 (x86_64) is example with "llvm-toolchain-xenial-8" repository.
https://github.com/ruby/ruby/blob/master/.travis.yml#L99

And it seems that `llvm-toolchain-bionic-9` repository is available on [llvm web site](https://apt.llvm.org/). Maybe we can use the repository.

> deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main
> deb-src http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main

## arm32

And for arm32 case, I introduced you the example for "gcc on arm32" you.

I will try to run "clang-9 on arm32 to CI" on my forked repository.

nemequ:
Thanks.

I used that repo for a clang-9 for amd64 build, it works great.

I'm just not sure how to get it to cross-compile to arm32 in a way that will be usable to run tests; it's not like GCC where you install an `arm-linux-gnueabihf-gcc` and just use that.  Clang is [natively a cross-compiler](https://clang.llvm.org/docs/CrossCompilation.html), I just have no idea what values to pass.

All the examples I can find want me to DL a sysroot for a board (usually a RPi) to target, but this needs to target Ubuntu 18.04 arm32.  Seems like it should be easier, but I guess nobody is bothering.

junaruga:
Perhaps `- CFLAGS="-target armv8l-unknown-linux-eabihf"` might be used for the `clang-9 on arm32`.

I am still working in progress [here](https://github.com/junaruga/ci-multi-arch-native-test/blob/wip/travis-arm32-clang/.travis.yml).

[...]

PR #128, opened 2020-03-30
suggestion: drone: add arm32 clang-N cases.

junaruga:
This PR is related to https://github.com/nemequ/simde/issues/75 .
I want to suggest using Drone CI to test clang arm32 cases for now.

As a initial setting, I added clang-9 and clang-7 arm32 cases.
Here is the result of my forked repository. You see some errors by `make -j` on the clang-7 case.
https://cloud.drone.io/junaruga/simde/10/2/2

```
+ make -j
...
/drone/src/test/x86/../../simde/x86/sse2.h:2344:54: error: no member named 'neon_f64' in 'simde__m128d_private'
...
```

This case is for clang-7.
But I observed the similar errors on clang-10 and clang-9 arm32 cases in Fedora project build system.
Later I will share it too.

For Drone CI, if you are interested in starting Drone CI, the way is just to go to https://drone.io/ , and enable it for this repository.

Drone CI has 2 type of the setting files `.drone.yml` and `.drone.star`.
You can switch the used file in the setting page. The default setting is `.drone.yml`.
But you see this PR's 2 cases has common parts. Because `.drone.yml` does not support like a `matrix` feature in Travis.
If we want to do like a `matrix` feature, we need to use `.drone.star` that has Starlark language syntax. Starlark is a Python-like script language. [1]

How do you think?

[1] https://docs.drone.io/pipeline/scripting/starlark/

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,57 @@
+---
+kind: pipeline
+type: docker
+name: "clang-9 arm32"
+platform:
+  os: linux
+  arch: arm
+steps:
+  - name: test
+    image: ubuntu:bionic
+    environment:
+      CC: clang-9
+      CXX: clang++-9
+    commands:
+    - uname -m
+    - apt-get -yq update
+    - apt-get -yq install clang-9 cmake git-core make
+    - git submodule update --init --recursive
+    - mkdir test/build
+    - cd test/build
+    - cmake ..
+    - make -j
+    - ./run-tests
+trigger:
+  branch:
+    exclude:
+    - wip/*
+
+---
+kind: pipeline
+type: docker
+name: "clang-7 arm32"
+platform:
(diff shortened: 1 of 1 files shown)


nemequ:
This sounds like a good idea to me.  I actually had it enabled already, but hadn't added the configuration; notice that it actually built your PR already.

The starlink stuff looks really nice, and it would make it really easy to keep all the configuration stuff in sync.  I'd like to switch to that, but for now I'll go ahead and merge this as-is… we can always switch over later.

junaruga:
> The starlink stuff looks really nice, and it would make it really easy to keep all the configuration stuff in sync. I'd like to switch to that, but for now I'll go ahead and merge this as-is… we can always switch over later.

Okay. Thank you for merging the PR.

PR #131, opened 2020-03-30
Avx2 blend

mr-c:
`mm256_blend_epi32` is needed for https://tracker.debian.org/pkg/kalign


nemequ:
Awesome.  Merged, thanks.

Some compilers have trouble if the imm8 isn't an integer constant expression, so I added a trivial patch on top to move the the native definitions to macros.

PR #134, opened 2020-03-31
Correct typos in avx512f.h

rosbif:
I think that SIMDE_AVX512_ENABLE_NATIVE_ALIASES and SIMDE_AVX2_ENABLE_NATIVE_ALIASES should be SIMDE_AVX512F_ENABLE_NATIVE_ALIASES.
Or am I being stupid again?

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -352,7 +352,7 @@ simde_mm512_castpd_ps (simde__m512d a) {
     return r;
   #endif
 }
-#if defined(SIMDE_AVX512_ENABLE_NATIVE_ALIASES)
+#if defined(SIMDE_AVX512F_ENABLE_NATIVE_ALIASES)
 #  define _mm512_castpd_ps(a) simde_mm512_castpd_ps(a)
 #endif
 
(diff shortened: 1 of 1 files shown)


nemequ:
Oops.  No, you're absolutely right.  Merged (as 8a0a1627db81455cd155b8aac66716811c33abe8), thanks!

Unfortunately native aliases aren't really tested; we would have to copy all the test cases, add a way to map a function to either the prefixed or native version, and still there would be some minor differences we would need to ifdef around in the tests.  I'm not sure if you found these because you were trying to use them or just reading the code, but either way it's really valuable, so thank you!  Please don't hesitate to let us know if you find any more issues like this.

rosbif:
I have many local additions to avx512f but you asked me to hold off before submitting them.
Can I submit them now?

As a question of style (and maybe efficiency) would you prefer, for example:-
  r_.m256i[0] = simde_mm256_set1_epi8(a);
  r_.m256i[1] = simde_mm256_set1_epi8(a);
or
  r_.m256i[1] = r_.m256i[0] = simde_mm256_set1_epi8(a);
or something completely different e.g. memset()?

I haven't looked into generating test vectors yet, but testing my software pushes tens of gigabytes of quasi-random data through these functions.
Apart from rare special cases (e.g. the (stupid?) sign bit setting bytes to zero in byte shuffles), I think that my functions are well tested.
IIRC you said that you would generate test cases, but would you like me to try?

To answer your implicit questions:-
I found the first error when I C&P'd an existing function to use as a template for mine.
I thought WTF and so I grepped for ENABLE_NATIVE_ALIASES and found the others.

I started off using the simde_ versions but I recently tried SIMDE_ENABLE_NATIVE_ALIASES.
After a little effort this worked fine so I switched to using the native versions.

PR #135, opened 2020-03-31
Define SIMDE_AVX512BW_ENABLE_NATIVE_ALIASES

rosbif:
Shouldn't SIMDE_AVX512BW_ENABLE_NATIVE_ALIASES be conditionally defined and used something like this?


nemequ:
Yep.  Merged (6f7307e826abb290d0c4e531a087e52dbc7b83ad), thanks!

PR #137, opened 2020-04-01
Add simde_mm512_shuffle_epi8 and simde_mm512_cmpeq_epi8_mask to avx512bw.h

rosbif:
Add simde_mm512_shuffle_epi8() and simde_mm512_cmpeq_epi8_mask() functions.

Remarks:-
The shuffle_epi8 intrinsics are my favourite SIMD instructions.
You probably won't like the double AVX and SSE fallbacks.
They are essential for me as without the SSE fallback for shuffle my software runs **FIFTY** times more slowly (AVX512 code on an SSE4.2 capable laptop).


rosbif:
The coverage decrease seems logical.
IIRC you said not to worry about generating the test vectors but I am quite prepared to do so.
Also I have lots of AVX512F intrinsics on hold.

mr-c:
@rosbif If you've got the time, feel free to add the test vectors!

nemequ:
Yes, that would be great.  Here is some documentation: https://github.com/nemequ/simde/wiki/Implementing-a-New-Function#adding-a-test

rosbif:
I have prepared the tests.
Should I close this PR and open a new one with both commits together?

mr-c:
That's fine, we can squash the commits when we merge.

test/x86/avx512bw.c
```diff
@@ -1491,6 +1491,761 @@ test_simde_mm512_adds_epu16(const MunitParameter params[], void* data) {
...
+                           INT8_C(-119), INT8_C( -69), INT8_C(  62), INT8_C( -90),
+                           INT8_C(   2), INT8_C(  38), INT8_C(  40), INT8_C(   9),
+                           INT8_C(  13), INT8_C(  75), INT8_C( -86), INT8_C(-119),
+                           INT8_C(  96), INT8_C( -43), INT8_C(  33), INT8_C(  18),
+                           INT8_C(  69), INT8_C(  80), INT8_C(  24), INT8_C( -32),
+                           INT8_C(  23), INT8_C( -28), INT8_C( -52), INT8_C( -60)),
+      -4721616227963928955 },
+  };
+
+#if 0
```
mr-c, line 2174: oops, this section should be removed

rosbif:
I deliberately left, but disabled, my test vector generation code for you to check.
I had to add code to randomly make the two vectors equal about half the time.
Otherwise the coverage was virtually zero.
Please remove it if you agree.

nemequ:
Awesome, thanks!

In addition to @mr-c's comment about squashing the commits, it's also worth noting that pull requests are about merging one branch into another, so updating your branch updates the pull request.  You can even do forced pushes if you want to squash or amend a commit yourself.  Just know that GitHub doesn't send out e-mails when you update the commits, so please try to remember to comment whenever you want a review.

As for the patch, it looks pretty good, I just have a couple minor requests.  For the returned masks, the type should be `__mmask64`, not `int64_t`.  `__mmask64` is actually unsigned, so the format specifier should be switched, too.  "%lu` instead of `%ld`should do the trick, though I tend to use the portable `PRIu64` macro (it doesn't really matter here since it's only running on your computer, but it's probably a good habit to get into).  The generated code also needs to be enclosed in a `UINT64_C()` to work reliably and without warnings, so something like

```
printf("      UINT64_C(%20" PRIu64 ") },\n", r);
```

For the issue of making them equal half the time, I usually do something like this right after the `munit_rand_memory()` calls:

```
/* You can hardcode the length as 64 instead of the sizeof stuff  */
for (size_t j = 0 ; j < (sizeof(a_.i8) / sizeof(a_.i8[0])) ; j++) {
  if (munit_rand_int_range(0, 1))
    b_.i8[j] = a_.i8[j];
}
```

Of course it needs to be tweaked slightly for different types and situations, but that's the general idea.

I'm highly sympathetic about using #if 0 instead of getting rid of the test vector generation stuff.  I'm always really tempted to do that, but the test files are already gigantic.

If you want me to go ahead and make the changes when I merge this just let me know.  Basically, you should feel free to do as much or as little of the coding as you want; I'm just grateful for any help you're willing to provide :)

rosbif:
Thank you Evan and Michael for your help.
Please go ahead with suppressing my #if 0 and making the minor changes you suggest while merging.
I shall continue with my many additions to avx512f, although some have been implemented by others in the meantime.
Concerning the int64_t mask: Intel has a silly habit of returning signed masks rather than unsigned ones (c.f. the movemask intrinsics).  This caused me some problems with sign extension during my work.
I figured that _mm512_cmpeq_epi8_mask's return value would also be signed.

rosbif:
I replied too quickly; I didn't realize that these corrections weren't just a few minor retouches.
If you haven't already started the corrections, let me know and I'll do them to save your precious time

rosbif:
Remarks:-
__mmask64 wasn't defined so I used simde__mmask64.
I prefer to print bit masks in hexadecimal not decimal.

PR #140, opened 2020-04-01
Fix type for argument 3 in _mm_maskmoveu_si128

jsbache:
Fix type for argument 3 in simde_mm_maskmoveu_si128 when using native aliases. The Intel intrinsic is declared as taking a 'char*' and we must therefore cast in order to redirect to the simde version

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -2893,7 +2893,7 @@ simde_mm_maskmoveu_si128 (simde__m128i a, simde__m128i mask, int8_t mem_addr[HED
 #endif
 }
 #if defined(SIMDE_SSE2_ENABLE_NATIVE_ALIASES)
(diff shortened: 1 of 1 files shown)


nemequ:
I merged this, but switched the cast to use [`HEDLEY_REINTERPRET_CAST`](https://github.com/nemequ/hedley/blob/master/hedley.h#L727) instead of a C-style cast (just to get rid of a potential warning; -Wold-style-cast in clang, and I know IAR has a similar warning).

I'll try to go through all the functions which accept a `mem_addr` later today, they'll probably all have this problem.

mr-c:
Thank you @jsbache!

nemequ:
Indeed, thank you for the patch, @jsbache.  This type of stuff is really helpful, and greatly appreciated.

I went through the other `mem_addr` parameters and added similar casts (87742887a0792f2f0408cff8d4ba194801d27c0c), so hopefully we won't see this pop up with other functions, too.  I'll also try to go through the rest of the API at some point soon to add casts for integer parameters, especially `int8_t`/`char`.

PR #141, opened 2020-04-01
Fixes for native alias mode of avx

jsbache:
Add defines for native alias mode for all of the SIMDE_CMP defines.
Fixed an incorrect native alias for _mm256_cvtepu8_epi16


nemequ:
Split into two patches (7c28e6b8c195e8dc39fe813e8069034392e31c11 & 91255b91d2999857e2fbdbb1ef982eff9b2c8675) and merged, thanks!

I'd probably have left it as a single commit if the changes were in the same file, but I'd generally like to keep unrelated changes in separate commits.  I'm completely fine with them being in the same PR, though.

PR #142, opened 2020-04-01
avx512f : Implement simde_mm512_cvtepi8_epi*

himanshi18037:


nemequ:
Thanks for the patch!

The problem on s390x is that the input vector for the tests doesn't match the "from" type, and since s390x is big-endian the layout of the memory is i bit different.  I went ahead and pushed a fixed version (5b1723577ea6e9052c8dbe98b5367bac01fd785f).

PR #143, opened 2020-04-01
Change fundamental types for simde__m128i

jsbache:
Previously typeof(m128) == typeof(m128i) and this is a problem for code that use mmx types to specialize templates. Specifically the following compiles for Intel, but fails for Arm:
template<>
inline void Foo<__m128>::Bar()
{ . . .}
template<>
inline void Foo<__m128i>::Bar()	<— fails because the underlying type float32x4_t has already been specialized.
{ . . .}

The Intel defines are:
	typedef float __m128 __attribute__((__vector_size__(16), __aligned__(16)));
	typedef long long __m128i __attribute__((__vector_size__(16), __aligned__(16)));

This proposal switches __m128i to using int64_t

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -244,9 +244,9 @@ typedef union {
   typedef __m128i simde__m128i;
   typedef __m128d simde__m128d;
 #elif defined(SIMDE_SSE2_NEON)
-  typedef __attribute__((neon_vector_type(2))) int64_t simde__m128i;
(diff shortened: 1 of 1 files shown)


jsbache:
The failures are in the " Install APT Dependencies" step and I don't see any useful information in the failure report

nemequ:
Good catch.  Merged, thanks!

Sorry about all the issues :(

PR #148, opened 2020-04-03
Lots more C&P errors in avx512f.h

rosbif:
avx512f.h: Correct new C&P errors

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -1013,7 +1013,7 @@ simde_mm512_set1_epi8 (int8_t a) {
   #endif
 }
 #if defined(SIMDE_AVX512F_ENABLE_NATIVE_ALIASES)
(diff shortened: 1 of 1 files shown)


rosbif:
I noticed a few days ago that my code would no longer compile with SIMDE_ENABLE_NATIVE_ALIASES so I went back to simde prefixes.
As I intend to submit modifications to avx512f.h I thought I should correct these problems first.
I corrected those that prevent my code from compiling but there may be others.

Silly question: why does "git status" say the master branch of my nemequ/simde clone is up to date when "git pull" has things to update?

mr-c:
> Silly question: why does "git status" say the master branch of my nemequ/simde clone is up to date when "git pull" has things to update?

Not silly :-) 
`git status` is about your local checkout

`git pull` is really a `git fetch` (contact the server and learn about the latests commits) followed by a `git merge`

PR #150, opened 2020-04-04
Add avx512f intrinsics

rosbif:
avx512f: add the following intrinsics:-
_mm512_broadcast_i32x4
_mm512_load_si512
_mm512_loadu_si512
_mm512_srli_epi64
_mm512_xor_si512


rosbif:
I tried to keep alphabetical order but I didn't rearrange those that were already disordered.
I found that I don't need explicit fallbacks for these intrinsics with simple pure C implementations.
However my compilers obviously don't efficiently vectorize "complicated" intrinsics like those I added to avx512bw.

I may not need some of the explicit SSE fallbacks I added to avx.h and avx2.h previously.
Do you want me to check and remove them if possible?

rosbif:
The failed check seems to be another APT timeout :-(

nemequ:
Thanks for the patches!

Travis and AppVeyor are failing, too, and not because of APT failures.  I restarted GH Actions and it looks like that succeeds, but there are problems on other compilers and architecutres.

> I tried to keep alphabetical order but I didn't rearrange those that were already disordered.

Most of the functions in AVX-512 that are out of order need to come first because they're used to implement other functions.

> I may not need some of the explicit SSE fallbacks I added to avx.h and avx2.h previously.
> Do you want me to check and remove them if possible?

Unfortunately not all compilers behave the same, especially not for all versions.  It's generally best to have fallbacks in place when possible since it's not feasible to check all versions of all compilers on all architectures :(

test/x86/avx512f.c
```diff
@@ -1309,6 +1534,214 @@ test_simde_mm512_setone_si512(const MunitParameter params[], void* data) {
...
 }
 
+static MunitResult
+test_simde_mm512_srli_epi64(const MunitParameter params[], void* data) {
+  (void) params;
+  (void) data;
+
+  const struct {
+    simde__m512i a;
+    uint32_t     b;
```
nemequ, line 1544: This type should match the parameter, which in this case is `unsigned int` not `uint32_t`.  `unsigned int` isn't 32-bits on all platforms; IIRC it's 16 bits on AVR (though I think that violates the C standard), and 64 bits on some platforms (ILP64 and SILP64).

test/x86/avx512f.c
```diff
@@ -1309,6 +1534,214 @@ test_simde_mm512_setone_si512(const MunitParameter params[], void* data) {
...
+    simde__m512i r;
+  } test_vec[8] = {
+    { simde_x_mm512_set_epu64(UINT64_C(0x3d7f1878959e7823), UINT64_C(0xdb19c9e23de5574c),
+                              UINT64_C(0x41f761e519d11af5), UINT64_C(0x87862bc5e638d1f3),
+                              UINT64_C(0x48d3b224fa800458), UINT64_C(0xa540b9aff87ed332),
+                              UINT64_C(0x171283b07eae579a), UINT64_C(0xf871de8b316995f7)),
+      UINT32_C(0),
```
nemequ, line 1551: Again, the type should be `unsigned int` not `uint32_t`, so these should just be `xU` instead of `UINT32_C(x)`.

test/x86/avx512f.c
```diff
@@ -1309,6 +1534,214 @@ test_simde_mm512_setone_si512(const MunitParameter params[], void* data) {
...
+                              UINT64_C(0x016582685c8964d4), UINT64_C(0x01c7fa22e30a4cf5)) },
+    { simde_x_mm512_set_epu64(UINT64_C(0xef4891f108bd3572), UINT64_C(0x70ed11eef2e11aa0),
+                              UINT64_C(0xef3984671f62c16d), UINT64_C(0x00cdcea30b923238),
+                              UINT64_C(0x44ef5f23e94bd07e), UINT64_C(0x61a977d25ae9865b),
+                              UINT64_C(0xbe912feb87b529cc), UINT64_C(0x6a55790501f02cd4)),
+      UINT32_C(1),
+      simde_x_mm512_set_epu64(UINT64_C(0x77a448f8845e9ab9), UINT64_C(0x387688f779708d50),
+                              UINT64_C(0x779cc2338fb160b6), UINT64_C(0x0066e75185c9191c),
+                              UINT64_C(0x2277af91f4a5e83f), UINT64_C(0x30d4bbe92d74c32d),
+                              UINT64_C(0x5f4897f5c3da94e6), UINT64_C(0x352abc8280f8166a)) },
```
nemequ, line 1618: Hex can be nicer to look at for bitwise operations, but all the assertion functions (like `simde_assert_m512i_u64`) print errors in decimal.  Using hex here makes it hard to tell which case is failing.

simde/x86/avx512f.h
```diff
@@ -605,6 +625,38 @@ simde_mm512_castsi512_si256 (simde__m512i a) {
...
+#endif
+
+SIMDE__FUNCTION_ATTRIBUTES
+simde__m512i
+simde_mm512_loadu_si512 (simde__m512i const * mem_addr) {
+#if defined(SIMDE_AVX512F_NATIVE)
+  return _mm512_loadu_si512((__m512i const*) mem_addr);
+#else
+  simde__m512i r;
+  r = *mem_addr;
```
nemequ, line 652: This needs to use `simde_memcpy`.  Unlike `_mm512_load_si512`, `_mm512_loadu_si512` has no alignment requirements, and `*mem_addr` requires `_Alignof(simde__m512i)` alignment, which is usually 64-byte.  x86 and AArch64 tends to be pretty forgiving about unaligned accesses, but other architectures (such as arm32) will result in errors (usually a SIGBUS, IIRC).

simde/x86/avx512f.h
```diff
@@ -1243,6 +1295,59 @@ simde_mm512_setone_pd(void) {
...
+simde_mm512_srli_epi64 (simde__m512i a, unsigned int imm8) {
+#if defined(SIMDE_AVX512F_NATIVE)
+  return _mm512_srli_epi64(a, imm8);
+#else
+  simde__m512i_private
+    r_,
+    a_ = simde__m512i_to_private(a);
+
+#if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
+  r_.u64 = a_.u64 >> HEDLEY_STATIC_CAST(int32_t, imm8);
```
nemequ, line 1309: The cast here is unnecessary, but the [documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_srli_epi64&expand=5530) says this function only looks at the least significant 8 bits, so you'll want to mask `0xff`.

You might want to use `int_fast16_t` for the result of the mask operation (in which case you would need a cast).  I would say `int_fast8_t`, but that's actually `sizeof(char)` in a lot of places (including my x86_64 desktop).  That should give the compiler the most room to choose the fastest type.

nemequ: Also, to be clear, the loop below suffers from the same issue, so the mask should be common to both cases.

test/x86/avx512f.c
```diff
@@ -1309,6 +1534,214 @@ test_simde_mm512_setone_si512(const MunitParameter params[], void* data) {
...
+  };
```
nemequ, line 1619: It's probably a good idea to include at least one value outside of [0, 255] for `b`.  For example, 1066, which should shift by 42 bits (`1066 & 0xff == 42`).

rosbif:
> Travis and AppVeyor are failing, too, and not because of APT failures. I restarted GH Actions and it looks like that succeeds, but there are problems on other compilers and architecutres.

Oops, sorry.
> > I may not need some of the explicit SSE fallbacks I added to avx.h and avx2.h previously.
> > Do you want me to check and remove them if possible?
> 
> Unfortunately not all compilers behave the same, especially not for all versions. It's generally best to have fallbacks in place when possible since it's not feasible to check all versions of all compilers on all architectures :(

I must have misunderstood you: I had the impression that you wanted to avoid fallbacks and let the compiler do the job of vectorizing. I have AVX and SSE fallbacks for virtually all the intrinsics that I sent. I carefully removed them before submitting. Do you want me to put them back in?

test/x86/avx512f.c
rosbif: OK, I'll correct.

test/x86/avx512f.c
rosbif: OK, I'll correct.
I didn't understand the purpose of these macros.
When I sent a raw value in another PR, someone told me to add a UINT64_C() so I thought they were mandatory.

test/x86/avx512f.c
rosbif: OK I'll change.
I deliberately went to the trouble of using hex for most bitwise operations as it is more understandable for humans :-(

test/x86/avx512f.c
rosbif: You're good at logical operations on decimals ;-)
The example I based this on only used a fixed shift value of 7 (IIRC) :-(
(Maybe this was because some functions like this need the shift value to be a compile-time constant but I don't think this was the case here.)
I decided that this wasn't sufficient coverage.
I deliberately forced 64 because sometimes shift operators take the shift count modulo the number of bits.
I'll go further if you wish.

simde/x86/avx512f.h
rosbif: OK, I'll correct.
To me all this HEDLEY stuff is a mystery.
What is its purpose?
The function I pinched the code from had this HEDLEY_STATIC_CAST so I thought it was necessary.

nemequ:
> > Travis and AppVeyor are failing, too, and not because of APT failures. I restarted GH Actions and it looks like that succeeds, but there are problems on other compilers and architecutres.
> 
> Oops, sorry.
> 
> > > I may not need some of the explicit SSE fallbacks I added to avx.h and avx2.h previously.
> > > Do you want me to check and remove them if possible?
> > 
> > 
> > Unfortunately not all compilers behave the same, especially not for all versions. It's generally best to have fallbacks in place when possible since it's not feasible to check all versions of all compilers on all architectures :(
> 
> I must have misunderstood you: I had the impression that you wanted to avoid fallbacks and let the compiler do the job of vectorizing. I have AVX and SSE fallbacks for virtually all the intrinsics that I sent. I carefully removed them before submitting. Do you want me to put them back in?

No, it's my fault, I should have been much clearer.

Falling back on a half-length vector is probably a good idea more often than not.  For example, implementing `__m512i` functions using `__m256i`.  However, falling back on an `__m128i` implementation should only be done if `__m256i` isn't possible; the `__m256i` version will fall back on an `__m128i` implementation, so adding a `__m128i` implementation to an `__m512i` function is just repeating yourself unnecessarily (modulo compiler bugs, of course).  In other words, AVX fallbacks would *usually* be good, SSE fallbacks *and* AVX fallbacks are probably a bad idea.

In some situations the shorter-length fallbacks may make things worse, especially on non-x86 platforms.  In other situations, such as on x86, without `-O3`/`-fopenmp-simd, and/or on older compilers.  Overall it's probably a win, though.

What really needs to happen is for me to go through and make some changes in how the fallbacks work so we do a better job of choosing whether to use the fallback or rely on the compiler.  Just checking to see if we have a native implementation isn't a good test.  It's relatively high on my list; I was actually hoping to get to it last week.

For example, consider `_mm_min_epi32` (sse4.1) when implementing `_mm256_min_epi32` (avx2).  If we just test for `#if defined(SIMDE_SSE4_1_NATIVE)` we are never going to hit that anywhere except for x86.  However, all versions of NEON have `vminq_s32` which should be used to implement `_mm_min_epi32`, so if we don't use it there is a good chance the result will be significantly slower.  Another (probably better) example is saturated math.  I've tried really hard to get compilers to generate something decent, and they just don't cooperate.  However, a lot of SIMD ISA extensions (including NEON and I think WASM) have saturated math functions.

Eventually we may start tweaking the priority of fallbacks; for example, it may make sense to prioritize GCC-style vector extension implementations, then use AVX.  However, this will be a lot easier if we have the code in place to reorganize.

Basically, the answer to the question of "are shorter-length fallbacks good or bad?" is "it's complicated."

I think the source of the misunderstanding is that I was trying to say that in the past I wasn't overly concerned with the shorter-length fallbacks for various reasons, which explains why a lot of the functions in AVX don't have them.  All I was really trying to say was that I didn't think it was quite as vital as you seemed to think; you're looking at relatively old compilers on x86, and there is more to it than that.

simde/x86/avx512f.h
rosbif: Sorry, I'll correct.
I realized for the alignment: I put a simde_assert_aligned in the non-u version.

On a similar subject:-
Should I use simde_memset for intrinsics which set all bytes to the same value (sometimes even all zero)?

In avx.h why does simde_mm256_lddqu_si256 call _mm256_loadu_si256 in native?
Particularly as, although equivalent, the Intel Intrinsics Guide says that _mm256_lddqu_si256 may sometimes perform better.
(Why the hell then did Intel put both in the same AVX version?)
Maybe the lddqu version isn't supported by some compilers.
How do you handle this in general?
I note that e.g. _mm256_set_m128i isn't supported by my gcc version but is OK with clang.
Are there many like this?

test/x86/avx512f.c
nemequ: They're from the C standard, for portability.  See § 7.20.4.1 of C17, but they go back to C99.  They are mandatory for fixed-length types, but should not be used for int/short/long/char/long long/etc.

Think of them as adding, say, U or L suffixes.  When declaring an `unsigned long` with a value of 1729 in your code, you should use `1729UL` (it's more important for larger values, but you get the idea).  However, a `uint32_t` may be mapped to unsigned long, unsigned int, unsigned long long, etc., so for portability we need a macro which appends the correct suffix.  If `uint32_t` is compatible with `unsigned int`, `UINT32_C(1729)` is equivalent to `1729U`.  If it's compatible with `unsigned long`, then it's equivalent to `1729UL`, and so on.

rosbif:
> > > > I may not need some of the explicit SSE fallbacks I added to avx.h and avx2.h previously.
> > > > Do you want me to check and remove them if possible?
> > > 
> > > 
> > > Unfortunately not all compilers behave the same, especially not for all versions. It's generally best to have fallbacks in place when possible since it's not feasible to check all versions of all compilers on all architectures :(
> > 
> > 
> > I must have misunderstood you: I had the impression that you wanted to avoid fallbacks and let the compiler do the job of vectorizing. I have AVX and SSE fallbacks for virtually all the intrinsics that I sent. I carefully removed them before submitting. Do you want me to put them back in?
> 
>Falling back on a half-length vector is probably a good idea more often than not. For example, implementing `__m512i` functions using `__m256i`. However, falling back on an `__m128i` implementation should only be done if `__m256i` isn't possible; the `__m256i` version will fall back on an `__m128i` implementation, so adding a `__m128i` implementation to an `__m512i` function is just repeating yourself unnecessarily (modulo compiler bugs, of course). In other words, AVX fallbacks would _usually_ be good, SSE fallbacks _and_ AVX fallbacks are probably a bad idea.

I has hoped that it would work like this (AVX512 falling back to AVX which in turn falls back to SSE if necessary) but unfortunately it doesn't :-(
For me the killer intrinsic is _mm512_shuffle_epi8. Without the double fallback I get a 50x performance penalty running on SSE.
I think the reason is that on SSE, SIMDE_ARCH_X86_AVX2 isn't defined so it won't even try to fallback to AVX.

test/x86/avx512f.c
nemequ: Yeah, I agree, it's easier to look at for that.  If you feel strongly about it you could always add macros like `simde_assert_m128i_i32x` (note the x suffix) for a hex version.  I'd accept a patch, but I'm not going to bother writing it myself ;)

simde/x86/avx512f.h
nemequ: > Should I use simde_memset for intrinsics which set all bytes to the same value (sometimes even all zero)?

IIRC I usually do that for one version (like *_ps), then implement the others with a castps_*(setzero_ps()).  But yeah, memset would work.

> Particularly as, although equivalent, the Intel Intrinsics Guide says that _mm256_lddqu_si256 may sometimes perform better.

I don't remember that being intentional; it should probably be reversed.

> (Why the hell then did Intel put both in the same AVX version?)

IIRC the syntax in assembly is a bit different, but I'd have to look.  Even if it were, I don't understand why there would be a performance difference; it seems like they should be able to compile to the same microcode…

> How do you handle this in general?

Case-by-case basis.  Luckily there aren't many.

> I note that e.g. _mm256_set_m128i isn't supported by my gcc version but is OK with clang.
> Are there many like this?

A few, yes.  That's one reason our CI is so extensive.  The real fun part is trying to figure out when a compiler added a function so we can support it natively in newer versions of the compiler.  IIRC you said you were using fairly old compilers; I suggest using [Compiler Explorer](https://www.godbolt.org) to try to figure that out, and hope you don't run into issues with compilers not on Compiler Explorer.  If you do, let me know; I have a lot of compilers installed locally and might be able to help.

simde/x86/avx512f.h
nemequ: [Hedley](https://github.com/nemequ/hedley) doesn't really have one purpose; it's a collection of stuff to do different things.  You might want to skim through [the web site](https://nemequ.github.io/hedley/).

`HEDLEY_STATIC_CAST` is pretty straightforward; in C++, the code you have above gets preprocessed to `static_cast<uint32_t>(imm8)`, and in C it gets preprocessed to `((uint32_t) imm8)`.  Basically, it helps keep C++ compilers happy while preserving compatibility with C.  Several C++ compilers actually have a warning (`-Wold-style-cast` in clang, I know IAR has one, too) if you use a C-style cast in your C++ code.

test/x86/avx512f.c
rosbif: Thanks.
C99 ??? C17 ??? I'm still on K&R ;-) No only joking !

test/x86/avx512f.c
nemequ: Not really, but my calculator is.  I'm too lazy to do it by hand :)

Yeah, 64 (or really anything in [64, 255]) is good to have, but you're still not really checking to make sure the upper 3 bytes are ignored.  Actually, you might want to do one where `b & 0x3f == b & 0xff` (like 1066; maybe generate it with `munit_rand_memory(sizeof(b), &b); b &= 0xc0;`), and one where `b & 0x3f != b & 0xff`?  That way you also cover the case of having to ignore high bits *and* make sure the shift isn't modulo n bits.

nemequ:
> I think the reason is that on SSE, SIMDE_ARCH_X86_AVX2 isn't defined so it won't even try to fallback to AVX.

Yes, the changes I'm talking about would fix that.  Basically, instead of looking to see wether `SIMDE_ARCH_X86_AVX2` (or `SIMDE_AVX2_NATIVE`) is set, it would fall back on the AVX2 function whenever the largest natively-supported vector size is < 512 bits. 

FWIW, if you can use one of shuffle functions with an immediate-mode argument you should get good performance on pretty much any architecture with gcc or clang since we can use `__builtin_shuffle`/`__builtin_shufflevector`.

test/x86/avx512f.c
nemequ: MSVC still doesn't fully support C99.  They only added inttypes.h and stdint.h in 2013, and there are actually a lot of people still using old MSVC versions… upgrading a proprietary compiler which costs a significant amount of money isn't a very attractive idea for a lot of people.

test/x86/avx512f.c
rosbif: > It's probably a good idea to include at least one value outside of [0, 255] for `b`. For example, 1066, which should shift by 42 bits (`1066 & 0xff == 42`).

Houston we have a problem :-
As you say, according to the IIG, your 1066 (0x42a) should be equivalent to 42 (0x2a).
In fact shifting by 1066 always gives a zero result :-(
So it looks as if more than 8 bits are taken into account.
(Values > 16 bits are also taken into account.)

This is in /x86/avx512f/mm512_srli_epi64/native/c when running under the Intel SDE.
Or am I being stupid again?

rosbif: I attach quick and dirty test code to prove the IIG is wrong.
It is very slow under the Intel SDE so I only go up to 24 bits.
You can just change MAX if you are lucky enough to have real AVX512 hardware.
I thought I could attach it but it complained.
The formatting suffers a bit in "Preview", sorry.
So here it is inline:-

#include <stdio.h>
#include <stdlib.h>
#include <limits.h>
#include <immintrin.h>

#if 0
#define MAX ((unsigned long) UINT_MAX + 1)
#else
#define MAX (1UL << 24)
#endif

int main(void)
{
  __m512 ff;
  unsigned long i;
  unsigned int index, last_index = 0;
  int last = 0, is_zero, last_is_zero = 0;

  ff = _mm512_set1_epi8(0xff);

  for (i = 0; !last; ++i)
    {
      last = (i >= MAX);
      index = (unsigned int) i;

      is_zero = ! _mm512_test_epi64_mask(_mm512_srli_epi64(ff, index), ff);

      if ((is_zero != last_is_zero) || last)
	{
	  printf("%s from %u to %u\n", last_is_zero ? "zero" : "non zero", last_index, index - 1);
	  last_is_zero = is_zero;
	  last_index = index;
	}
    }

  return EXIT_SUCCESS;
}

test/x86/avx512f.c
rosbif: Is it realistic to run AVX512 on a machine with 16-bit ints?
I haven't worked on a 16-bit machine for at least 30 years.
I agree that while K&R permitted almost any size of int, recent standards impose a minimum of 32 bits. In K&R IIRC the only restriction was sizeof(long) >= sizeof(int) >= sizeof(short) >= sizeof(char) plus a char must be at least 8 (or 7 for ASCII?) bits. (I have even worked on a machine with 32-bit chars.)
Anyway for the shifts (see below) I wanted to check the effect of setting the MSBit which would generate a negative value if cast to int.
But in doing this I assumed 32-bit ints :-(
Can I issue a PR with this potential problem?

test/x86/avx512f.c
nemequ: It looks like the Intrinsics Guide is wrong.  I just used this:

```c
#include <immintrin.h>
#include <stdint.h>
#include <inttypes.h>
#include <stdlib.h>
#include <stdio.h>

int main(void) {
  __m512i v = _mm512_set1_epi64(~INT64_C(0));
  __m512i r1 = _mm512_srli_epi64(v, 0x042a);
  __m512i r2 = _mm512_srli_epi64(v, 0x042a & 0xff);

  printf("0x%016" PRIx64 "\n", *((int64_t*) (&r1)));
  printf("0x%016" PRIx64 "\n", *((int64_t*) (&r2)));

  return EXIT_SUCCESS;
}
```

To [test on AVX-512 hardware](https://travis-ci.org/github/nemequ/ci-noise/builds/671320307) (Travis CI often runs on hardware which supports AVX-512).

The good news is you don't need to worry about testing those values, or masking the input :)

Do you want me to take care of notifying Intel, or would you like to do it?

rosbif:
This commit should be applied after my previous one.
I hope that is OK, otherwise I'll combine them.
I hope I have addressed all of Evan's remarks.
I have limited the shift count values to 16 bits.

test/x86/avx512f.c
rosbif: Yes, I imagine that you have more contacts at Intel than I do.
The problem is worse with the 256 and 128 versions of these instructions.
Not only do they have this problem but they also take an int as the shift count and the pseudo-code doesn't test for negative values. They should change the shift count to unsigned int.
There are probably heaps of other intrinsics with similar problems and I just stumbled on this one.

nemequ:
Excellent, the changes look good, thank you!  The MSVC errors were latent issues, cbc60dc3c195dcdb3352efd9389b14f7e3d3fd93 fixes them.

I made a few changes prior to merging.  It's all trivial stuff, normally I'd just silently "fix" it and thank you for the patch but since it seems like you're interested in contributing going forward I thought it would be a good idea to let you know what I changed and why.

 * For the preprocessor stuff, I changed the indentation to basically treat it just like the C code.  I know there isn't any consistency on this in SIMDe, but I've been trying to slowly move to this style; it makes it easier to generate code programmatically which is formatted correctly.
 * In the xor function, I switched the portable fallback to use the i32f member (`int_fast32_t`), which may be slightly faster than `int64_t`.
 * Removed the MODIF_CM checks for the fallbacks.  I'm guessing this is an artifact from some other changes.
 * Used `*mem_addr` instead of `simde_memcpy` in `simde_mm512_load_si512`.  This function is for aligned loads, so it's okay.  The `simde_assert_aligned` call theoretically makes this unnecessary since it uses the [`HEDLEY_ASSUME`](https://github.com/nemequ/hedley/blob/d9261eeb6907b502dd06581932a6738ce4f65012/hedley.h#L1113) to tell the compiler that the data is aligned, but not all compilers support it.
 * Switched from C++ style comments to C style comments (`/* ... */` instead of `// ...`).  Again, this shouldn't really matter since we already require C99 (or at least the subset MSVC supports) and C++ style comments are included in C99, but it does keep our C99 dependency weaker.

Anyways, thanks for all the work on this patch, and sorry again for not being clear about the fallbacks.

test/x86/avx512f.c
rosbif: > I made a few changes prior to merging. It's all trivial stuff, normally I'd just silently "fix" it and thank you for the patch but since it seems like you're interested in contributing going forward I thought it would be a good idea to let you know what I changed and why.
> 
>     * For the preprocessor stuff, I changed the indentation to basically treat it just like the C code.  I know there isn't any consistency on this in SIMDe, but I've been trying to slowly move to this style; it makes it easier to generate code programmatically which is formatted correctly.
> 

I don't really agree with these preprocessor modifications.
Unless this has changed in the last 40 years, strictly the # of the preprocessor directive **MUST** be in column 1 although few preprocessors enforce this. However the # can be separated from the keyword.

```
// These are OK
#         define TEST 1
#         if TEST

// but these are not
         #define TEST 1
         #if TEST
```

>     * In the xor function, I switched the portable fallback to use the i32f member (`int_fast32_t`), which may be slightly faster than `int64_t`.
>

OK

>     * Removed the MODIF_CM checks for the fallbacks.  I'm guessing this is an artifact from some other changes.
>

So sorry about this.
I thought I had removed them all.
I obviously didn't look at the final diff carefully enough.
They were for me to easily test the speed difference with and without the fallbacks.

>     * Used `*mem_addr` instead of `simde_memcpy` in `simde_mm512_load_si512`.  This function is for aligned loads, so it's okay.  The `simde_assert_aligned` call theoretically makes this unnecessary since it uses the [`HEDLEY_ASSUME`](https://github.com/nemequ/hedley/blob/d9261eeb6907b502dd06581932a6738ce4f65012/hedley.h#L1113) to tell the compiler that the data is aligned, but not all compilers support it.
>

OK

>     * Switched from C++ style comments to C style comments (`/* ... */` instead of `// ...`).  Again, this shouldn't really matter since we already require C99 (or at least the subset MSVC supports) and C++ style comments are included in C99, but it does keep our C99 dependency weaker.
>

Noted

nemequ: > Unless this has changed in the last 40 years, strictly the # of the preprocessor directive MUST be in column 1 although few preprocessors enforce this. However the # can be separated from the keyword.

C99 and up pretty clearly allow it.  [§ 6.10 ¶ 2 of the C99 spec](http://port70.net/~nsz/c/c99/n1256.html#6.10p2) (emphasis added):

> A preprocessing directive consists of a sequence of preprocessing tokens that satisfies the following constraints: **The first token in the sequence is a # preprocessing token that** (at the start of translation phase 4) is either the first character in the source file (optionally after white space containing no new-line characters) **or that follows white space containing at least one new-line character**. The last token in the sequence is the first new-line character that follows the first token in the sequence. A new-line character ends the preprocessing directive even if it occurs within what would otherwise be an invocation of a function-like macro.

AFAIK C89 was ambiguous about this (and pretty much everything else, really), so I think there were some compilers which didn't allow leading whitespace, but I don't think I've ever actually run into one in the wild, though.

rosbif: Sorry, as I said in my edit, you are right.
However at work, post 2000, I had an unforgiving compiler on one of the historic platforms for our multi-platform software.
I think it was probably the compiler for the DEC Alpha on their Tru64 Unix system V4.0F.
(Incidentally, at the time, the Alpha was a wonderful processor, blowing away anything from Intel/AMD.)
However it just could have been the SGI compiler on Irix.

nemequ: Sorry, I didn't see the edit, just read the e-mail.  That makes sense; I've only been using C since the late 90s, and then I was really only using GCC (IIRC I actually started with egcs, which should give you a good idea of the timeframe), so while I heard people say the # had to be at the beginning of the line it was never a problem for me personally.

Anyways, back to the masking stuff.  I was looking into this a bit further, and it looks like it's actually the compilers optimizing anything > 63.  See <https://godbolt.org/z/2thNFh>.  MSVC is the only compiler that doesn't optimize anything greater than 63 to an xor of itself (i.e., all zeros).  And what MSVC does generate for values > 63 is insane; instead of just using the value as an immediate mode argument to vpsrlq they move it into a vector register then use that.

rosbif: Thanks for pointing me to Compiler Explorer: I didn't know it.
I tried the following code where it cannot possibly know the shift count and I can see no > 63 test.

```
#include <immintrin.h>
#include <stdlib.h>

__m512i foo(__m512i bar) {
    unsigned int j = rand();
    return _mm512_srli_epi64(bar, j);
}

```

nemequ: Enjoy Compiler Explorer; it's a fantastic tool.  FWIW, you can use SIMDe on it: http://simde.netlify.com/godbolt/simple is the basic one, or http://simde.netlify.com/godbolt/demo to see a bunch of compilers. 

Good point about making sure the compiler can't know the value; I'm used to the intrinsic requiring that argument be a compile-time constant, so I didn't even think of it.  If I do

```c
#include <immintrin.h>
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>

__m512i foo(__m512i bar, unsigned int j) {
    return _mm512_srli_epi64(bar, j);
}

int main(int argc, char** argv) {
  __m512i v = _mm512_set1_epi64(~UINT64_C(0));
  __m512i r = _mm512_srli_epi64(v, atoi(argv[1]));
  printf("0x%llx\n", ((uint64_t*) &r)[0]);

  return 0;
}
```

It uses all 64 bits, which makes sense; it matches the behavior of `_mm512_srl_epi64`.  I'm now convinced that the intrinsics guide is wrong.

PR #162, opened 2020-04-06
avx512f: added implementation for mm512_mul_ps, mm512_mul_pd, mm512_div_pd, mm512_div_ps, mm512_or_si512

masterchef2209:
avx512f: added implementation for mm512_mul_ps, mm512_mul_pd, mm512_div_pd, mm512_div_ps, mm512_or_si512


simde/x86/avx512f.h
```diff
@@ -2445,6 +2445,144 @@ simde_mm512_cvtepi64_epi16 (simde__m512i a) {
...
+
+#if defined(SIMDE_ARCH_X86_AVX2)
+  r_.m256i[0] = simde_mm256_or_si256(a_.m256i[0], b_.m256i[0]);
+  r_.m256i[1] = simde_mm256_or_si256(a_.m256i[1], b_.m256i[1]);
+#elif defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
+  r_.i32f = a_.i32f | b_.i32f;
+#else
+  SIMDE__VECTORIZE
+  for (size_t i = 0 ; i < (sizeof(r_.i32) / sizeof(r_.i32[0])) ; i++) {
+    r_.i32[i] = a_.i32[i] | b_.i32[i];
```
nemequ, line 2575: This should use i32f (which is `int_fast32_t`) instead of i32 (which is `int32_t`).  For bitwise operations the width of the lane doesn't matter, so it's best to give the compiler a bit more room to optimize, just like you did above with the GCC-style vector extension version.

masterchef2209: thanks I will keep that in mind

nemequ:
Merged, thanks!

PR #165, opened 2020-04-07
avx512f: added implementation for mm512_mask_add_x functions

masterchef2209:
avx512f: added implementation for mm512_mask_add_epi32, mm512_mask_add_epi64, mm512_mask_add_ps, mm512_mask_add_pd


nemequ:
The error on Travis is due to a problem in skel.c.  `simde__mmask16` is a typedef to `uint_fast16_t` not `uint16_t`.  Since `simde__mmask16` could be larger than a `uint16_t` we need to make an effort to only print 16 bits.  I've made some modifications to skel.c to fix this.

Other than that, which was my fault, there isn't anything wrong with this.  Nice work.

That said, there is an easier way to implement these: just do the non-masked operation and use `simde_mm512_mask_mov_*` to blend them together.  I added implementations of `simde_mm512_mask_sub_*` so you can see what I'm talking about, see 736d1b3578d6e365ddcb4feb37a38af0ca4d7580.

Doing it this way lets us reuse a *lot* of code which would otherwise be duplicated.  The drawback is it may be a bit slower if compilers naïvely have to extract, perform the non-masked operation, insert the result, extract the result, perform the mask_mov, and insert again.  I haven't actually looked to see if they're capable of optimizing out the extra insert then extract operation, but that's usually the type of thing they're really good at.  Even if they're not already doing it, I feel pretty good about the odds that at least gcc and clang will add such an optimization in the relatively near future.

masterchef2209:
Thank You, I will remember to minimize duplication from next time. I will fix it for mm_mask_add_x soon.

Apart from that I have another doubt about error which I was facing, I see int_fast32_t is defined as simde__m512i in SIMDE_VECTOR_SUBSCRIPT in avx512f.h and in stdint.h if _WORDSIZE==64 I see that long int is defined as int_fast32_t but when I was trying to implement mm512_max_epi32 I was getting compilation error that mm512_max_epi32 expects long long int as argument but is receiving a long int type, what should be the fix for this?

masterchef2209:
Hi, I have changed the implementation and used mm512_mask_mov_x for all functions.

nemequ:
Awesome, thank you!

I changed the function order a bit, so it goes `_mm512_add_epi32`, `_mm512_mask_add_epi32`, `_mm512_add_epi64`, `_mm512_mask_add_epi64`, etc.  I think it makes sense to keep these together; their implementations are obviously closely tied together, and this makes it easier to copy and paste the functions to use them as a basis for new functions.

I know the ordering in avx512f is a mess.  That's my fault, sorry, I'll try to clean it up soon.  Basically, there should be a group at the top of the file for functions stuff that's likely to be used in other functions, like `_mm512_mask_mov_*` and `_mm512_set{zero,one}_*`.  Next, each group (add, sub, mul, div, etc.) should be in alphabetical order with the `mask_foo_*` functions coming immediately after each corresponding `foo_*` function.

For the test cases, the same thing except let's put the mask functions after the entire group of non-mask functions.

> Apart from that I have another doubt about error which I was facing, I see int_fast32_t is defined as simde__m512i in SIMDE_VECTOR_SUBSCRIPT in avx512f.h and in stdint.h if _WORDSIZE==64 I see that long int is defined as int_fast32_t but when I was trying to implement mm512_max_epi32 I was getting compilation error that mm512_max_epi32 expects long long int as argument but is receiving a long int type, what should be the fix for this?

It depends on exactly what error you were getting, but generally the answer is to use a cast.  On most compilers, just doing for example `((__mmas16) k)` (where `k` is a `simde__mmask16`) is enough.  Since we only ever care about the lower 16 bits in `simde__mmask16` we're okay with narrowing the type to 16 bits.

However, a simple C-style cast like that can actually trigger warnings in some compilers in C++ mode.  For clang it's `-Wold-style-cast`.  IAR has one too, but I don't remember the number off the top of my head.  Since we want to make sure people can use SIMDe in either C or C++ compilers without warnings, we use `HEDLEY_STATIC_CAST(__mmask16, k)` instead; that will be preprocessed to `((__mmask16) k)` in C, but `static_cast<__mmask16>(k)` in C++.

With all that said, it's very important to note that you should be careful with this.  Diagnostics like this from the compiler are its way of telling your code might not do what you think it does, and casting is basically telling the compiler to shut up and do as it's told.  The thing is, compiler warnings are usually pretty good… so when you get a warning like this don't just assume the correct answer is to cast to get around it.  Sometimes the problem will be that you're accidentally using the wrong type, or there is some other issue.  Only cast when you're confident that it's really the right thing to do.

This is actually important for warnings in general.  They can be annoying, but they are there to help you find problems in your code early, which is *much* better than finding them later.  Only work to silence the warning once you're convinced it's the right thing to do.

masterchef2209:
I will try to work on it again, that was very informative.
Also, I will make sure to keep the order of functions as per your guidelines from next time :)

PR #166, opened 2020-04-08
sse.h: AltiVec optimisation in _mm_cmpneq_ps

himanshi18037:


nemequ:
Merged, thanks!

I was able to reproduce it; I just forgot to add -qaltivec when we were chatting earlier.  I removed that second cast and added 641d90fa342b7fb5f4ad29996be5cc07c49c715d so it wouldn't try to use that function on xlc.

Issue #167, opened 2020-04-08
aarch64 tests fail with clang-10.0.0 with -O2 and some flags 


junaruga:
On the latest master branch commit: `29b911059dd06bc1e4c30d8922871a4b239279e3`, I executed the unit tests with Fedora's recommended flags on Fedora rawhide (= 33) aarch64. The I got following 4 errors.

The flags are similar with gcc's one used in https://github.com/nemequ/simde/issues/156 .
But `-fstack-clash-protection` and `-specs=*` flags are filtered out by followng logic.

```
optflags_clang=$(echo "%{optflags}" | \ 
  sed -e 's| -fstack-clash-protection||' -e 's| -specs=[^ ]*||g')
```

```
$ clang --version
clang version 10.0.0 (Fedora 10.0.0-1.fc33)
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ clang++ --version
clang version 10.0.0 (Fedora 10.0.0-1.fc33)
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

```
$ optflags_clang='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables'

$ mkdir test/build-clang-with-flags

$ pushd test/build-clang-with-flags

CC=clang CXX=clang++ cmake \
  -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
  -DCMAKE_C_FLAGS="${optflags_clang}" \
  -DCMAKE_CXX_FLAGS="${optflags_clang}" \
  ..

$ /usr/bin/make -O -j5 V=1 VERBOSE=1
...
/x86/sse4_1/mm_cvtepu8_epi16/emul/c  [ ERROR ]
/x86/sse4_1/mm_cvtepu8_epi32/emul/c  Error: /builddir/build/BUILD/simde-29b911059dd06bc1e4c30d8922871a4b239279e3/test/x86/sse4.1.c:1193: assertion failed: r.i16[0] == test_vec[i].r.i16[0] (26627 == 3)
Error: child killed by signal 6 (Aborted)
[ OK    ] [ 0.00001405 / 0.00001225 CPU ]
...
[...]

PR #170, opened 2020-04-09
suggestion: drone: add Fedora rawhide clang build.

junaruga:
This PR is to add Fedora rawhide aarch64 (arm64) clang (10.0.0) case to Drone CI.
As the rawhide is a development environment, while it's good to check it on the latest compilers, it might be good to build with `failure: ignore`.

And it reproduces https://github.com/nemequ/simde/issues/167 .


nemequ:
Good idea.  Merged, thanks!

PR #176, opened 2020-04-11
drone: replace .drone.yml with .drone.star.

junaruga:
I replaced `.drone.yml` with `.drone.star`.
The both contents are same.

To enable `.drone.star`, you need to change Settings - Configuration `.drone.yml` to `.drone.star`.

The useful way to verify `.drone.star` and the result is to run [Drone CI cli command](https://docs.drone.io/cli/install/) on your local.

Following command outpus YAML file from current directory's `.drone.star`.
Drone CI system internally actually converts `.drone.star` to YAML format.

```
$ drone starlark convert --stdout > convert.yml
```

You can compare`convert.yml` with `.drone.yml`.

## About the implementation

Here is [the specification](https://github.com/google/starlark-go/blob/master/doc/spec.md
) of Starlark. Starlack has some differences from Python. For example, the variable defined on the top level is recognized as immutable object.

The object design is inspired from `.travis.yml`. You can see `jobs`, `before_install`, `install`, `before_script`, `script` object. The logic is merging each job setting based on a default job setting.

When the default job setting and jobs variables are declared as a top level, it is recognized as a immutable variable. 
As it prevents the job object to be merged to the default job object in the logic, `def get_default_job()`, `def get_jobs()` are used to get the objects.

We can do refactoring the `job_` variables later. This PR is a first step to use `.drone.star`.

Though `job_fedora` is commented out aligning current `.drone.yml`, I confirmed it worked.


junaruga:
By the way, as there are `meson` [deb package](https://packages.ubuntu.com/bionic-updates/meson
) in Ubuntu and [RPM package](https://src.fedoraproject.org/rpms/meson).

As when running pip3 install `meson`, the source compile is executed.
When installing `meson` from distro package, the source compile is not executed, as the package has compiled already. And maybe we can save a little time.

junaruga:
Here is the result of Drone CI on my forked repository.
https://cloud.drone.io/junaruga/simde/46

nemequ:
Awesome! This should provide a really good basis for expanding use of Drone. Merged, thanks!

> Maybe we can save a little time when using the distro package.

That would be fine with me. In the past I've had some problems with the packaged Meson from Ubuntu Bionic not being new enough, but I'm not sure SIMDe requires any features only available in newer versions of Meson. Definitely worth a try.

junaruga:
I am glad you like it!

> Definitely worth a try.

Okay. We can try it later.

PR #186, opened 2020-04-14
avx512f: implement _mm512_cvtsepi32_epi*

himanshi18037:


nemequ:
Excellent, thank you!  Merged as 4472fe2ce0f07dd44ba1f27d28b72ffb77c92912

I reorganized the functions in both the header and the test (and parts of skel.c) a bit prior to merging this.  In the future please try to maintain the order, and if you have any questions just ask.  That should also help reduce conflicts during merging since hopefully multiple people won't be trying to add code to the end of the file.

Additionally, I cleaned up the formatting for the preprocessor directives in avx512f.h to match the style I've been trying to switch to.  Basically, preprocessor directives are now indented just like blocks in normal C code.  I know there is a lot of code in other parts of SIMDe that doesn't follow this convention (I only started doing it recently), but if you could try to match that style for new code I'd appreciate it.  I'll try to keep an eye on it during reviews, too.

himanshi18037:
Thanks, I will focus on matching the style convention of code in further contributions.

PR #187, opened 2020-04-14
avx512f: added the implementation for mm512_maskz_div_x functions

masterchef2209:
avx512f: added the implementation for mm512_maskz_div_x functions

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -2812,6 +2812,19 @@ simde_mm512_mask_div_ps(simde__m512 src, simde__mmask16 k, simde__m512 a, simde_
 #endif
 
 SIMDE__FUNCTION_ATTRIBUTES
+simde__m512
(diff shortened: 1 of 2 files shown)


nemequ:
Great.  Merged, thanks!

I reorganized the file functions a bit.  In the future, please try to order things the same way.  I'll try to do a better job of keeping an eye on the order during review.

Since I was touching pretty much every line in avx512f.h anyways, I also cleaned up the formatting for the preprocessor directives a bit to match the style I've been trying to shift to.  Basically, preprocessor directives are now indented just like blocks in normal C code.  I know there is a lot of code in other parts of SIMDe that doesn't follow this convention (I only started doing it recently), but if you could try to match that style for new code I'd appreciate it.

Thanks again for the patch!

masterchef2209:
Sure, I will keep that in mind and follow the new convention.
Thank You

PR #195, opened 2020-04-16
avx512bw: implement _mm*_cvtsepi16_epi8

himanshi18037:


simde/x86/avx512bw.h
```diff
@@ -356,6 +356,85 @@ simde_mm512_cvtepi8_epi16 (simde__m256i a) {
...
+#if defined(SIMDE_AVX512BW_NATIVE)
+  return _mm_cvtsepi16_epi8(a);
+#else
+  simde__m128i_private r_;
+  simde__m128i_private a_ = simde__m128i_to_private(a);
+
+  r_ = simde__m128i_to_private(simde_mm_setzero_si128());
+  SIMDE__VECTORIZE
+  for (size_t i = 0 ; i < (sizeof(r_.i16) / sizeof(r_.i16[0])) ; i++) {
+     r_.i16[i] =
```
nemequ, line 371: Should be `r_.i8[i]` not `r_.i16[i]`.  Sorry, I really should have seen that when we were talking earlier.

It would also probably be more appropriate to use `sizeof(a_.i16) / sizeof(a_.i16[0])` for calculating the loop bounds, since we're not actually using `r_.i16`.  It doesn't really matter, I just think it would be a bit easier to read.

I know I was the one who suggested the above code, but how about moving the initialization code for r_ to the declaration (remove line 368, change 365 to `simde__m128i_priavte r_ = simde__m128i_to_private(simde_mm_setzero_si128());`?  Again, it doesn't really matter, it just seems a bit easier to read.

himanshi18037: Thanks, I will work on the changes suggested.

Issue #199, opened 2020-04-18
centos x86_64 clang-3.4.2: 1 test failure


junaruga:
I got the following 1 error with clang 3.4.2 on CentOS 7 x86_64 environment.

```
+ CC=clang
+ CXX=clang++
+ meson ../..
The Meson build system
Version: 0.47.2
Source dir: /builddir/build/BUILD/simde-29b911059dd06bc1e4c30d8922871a4b239279e3
Build dir: /builddir/build/BUILD/simde-29b911059dd06bc1e4c30d8922871a4b239279e3/test/build-clang
Build type: native build
Project name: SIMDe
Project version: 0.5.0
Native C compiler: clang (clang 3.4.2 "clang version 3.4.2 (tags/RELEASE_34/dot2-final)")
Native C++ compiler: clang++ (clang 3.4.2 "clang version 3.4.2 (tags/RELEASE_34/dot2-final)")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Checking for function "clock_gettime" : YES
Library rt found: YES
Compiler for C supports arguments -fopenmp-simd: NO
Compiler for C++ supports arguments -fopenmp-simd: NO
Compiler for C supports arguments -qopenmp-simd: NO
Compiler for C++ supports arguments -qopenmp-simd: NO
Dependency openmp found: NO
Build targets in project: 9
Found ninja-1.7.2 at /usr/bin/ninja-build

+ ninja-build -v
...
+ ./test/run-tests
Running test suite with seed 0x914d8d00...
...
/x86/sse/mm_cvtps_pi16/native        [ ERROR ]
/x86/sse/mm_cvtps_pi32/native        Error: ../../test/x86/sse.c:2128: assertion failed: r.i16[0] ≈ test_vec[i].r.i16[0] (127 ≈ 352) 
Error: child killed by signal 6 (Aborted)
[ OK    ] [ 0.00001398 / 0.00001298 CPU ]
...
1324 of 1325 (100%) tests successful, 0 (0%) test skipped.
```

junaruga:
As a information, the test failure doe not happen with the clang's `-O2` and some flags.

```
+ CC=gcc
+ CXX=g++
+ CFLAGS='-O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches   -m64 -mtune=generic'
[...]

PR #200, opened 2020-04-18
drone: add CentOS 7 clang 3 build.

junaruga:
This PR is to add CentOS 7 clang build in Drone CI.
It's also to reproduce https://github.com/nemequ/simde/issues/199 issue.
Here is [the CI result](https://cloud.drone.io/junaruga/simde/61/10/2) on my forked repository.

```
/x86/sse/mm_cvtps_pi16/native        [ ERROR ]
Error: ../test/x86/sse.c:2128: assertion failed: r.i16[0] ≈ test_vec[i].r.i16[0] (127 ≈ 352)
...
380 of 1381 (100%) tests successful, 0 (0%) test skipped.
```

Thanks!

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -6,8 +6,8 @@ def get_test_commands():
   return [
     "mkdir -p build",
     "cd build",
-    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson ..',
-    "ninja -v",
+    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson .. || (cat meson-logs/meson-log.txt; false)',
+    '"$(command -v ninja || command -v ninja-build)" -v',
     "./test/run-tests",
   ]
 
@@ -22,6 +22,12 @@ def get_dnf_install_commands(extra_pkgs = []):
   return [
     "dnf install -y %s ninja-build git-core python3-pip" % " ".join(extra_pkgs),
     "pip3 install meson",
+  ]
+
+def get_yum_install_commands(extra_pkgs = []):
+  return [
+    "yum install -y epel-release",
+    "yum install -y %s meson ninja-build git-core" % " ".join(extra_pkgs),
   ]
 
 def get_default_job():
@@ -250,6 +256,24 @@ def get_jobs():
     }
   }
 
+  job_centos7_clang3 = {
+    "name": "centos7 clang3",
+    "steps": [
+      {
+        "image": "centos:7",
+        "environment": {
+          "CC": "clang",
+          "CXX": "clang++",
+        },
+        "failure": "ignore"
+      }
+    ],
+    "custom": {
+      # gcc, gcc-c++ are necessary to build on clang.
+      "install": get_yum_install_commands(["clang", "gcc", "gcc-c++"])
+    }
+  }
+
   return [
     job_clang9_armv7,
     job_clang9_armv8,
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -6,8 +6,8 @@ def get_test_commands():
   return [
     "mkdir -p build",
     "cd build",
-    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson ..',
-    "ninja -v",
+    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson .. || (cat meson-logs/meson-log.txt; false)',
+    "ninja -v || ninja-build -v",
```
junaruga, line 10: For this line, I wanted to write like this to set the value as default: `ninja`. When centos7, the value is set as `ninja-build` by the environment variable. But I recognized the value `${FOO}` is not evaluated for the `{ environment: {"FOO": "something"}` in `.drone.star`. So, I could not adapt the way.

```
"${NINJA:-ninja} -v",
```

.drone.star
```diff
@@ -6,8 +6,8 @@ def get_test_commands():
   return [
     "mkdir -p build",
     "cd build",
-    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson ..',
-    "ninja -v",
+    'CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson .. || (cat meson-logs/meson-log.txt; false)',
```
junaruga, line 9: This `cat` command is useful to see the error log when `meson` fails.
This error actually happens on CentOS7 when building with clang without installing `gcc`, `gcc-c++`.
Here is the error log.

```
+ CFLAGS="$ARCH_FLAGS" CXXFLAGS="$ARCH_FLAGS" meson .. || (cat meson-logs/meson-log.txt; false) 
...
Sanity check compile stderr:
/usr/bin/ld: cannot find crt1.o: No such file or directory
/usr/bin/ld: cannot find crti.o: No such file or directory
/usr/bin/ld: cannot find crtbegin.o: No such file or directory
/usr/bin/ld: cannot find -lgcc
/usr/bin/ld: cannot find -lgcc_s
/usr/bin/ld: cannot find -lc
/usr/bin/ld: cannot find -lgcc
/usr/bin/ld: cannot find -lgcc_s
/usr/bin/ld: cannot find crtend.o: No such file or directory
/usr/bin/ld: cannot find crtn.o: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
...
```

.drone.star
junaruga: Perhaps we might be able to just replace `ninja -v` with `ninja-build -v` rather than `ninja -v || ninja-build -v`.

nemequ: With `"ninja -v || ninja-build -v"`, if the build fails because of a compilation problem it will try again with `ninja-build -v`.  How about something like `"\"$(which ninja || which ninja-build)\" -v"` instead?

Just using ninja-build would probably be okay, but if it's not an "official" alias (i.e., an alias installed by ninja not just distros) I'd prefer not to just because I'd like it to work if we ever add a configuration that doesn't support `ninja-build`.  That said, it wouldn't be a big deal.

junaruga: > "\"$(command -v ninja || command -v ninja-build)\" -v" instead?

This looks great! I might want to use the single quote to avoid the escape. `'"$(command -v ninja || command -v ninja-build)" -v'`. Let me fix it.

> i.e., an alias installed by ninja not just distros

You are right. It is not official alias.

nemequ:
Merged as b4eb6a9191741f6cc88f4fe51bb94ebed0f25bb4, thanks!

PR #208, opened 2020-04-20
Native alias test

mr-c:
Script used for testing: https://gist.github.com/mr-c/90dbca3383037914dbdd5abf935231d9

The following functions are implemented inside GCC 10 but not GCC 9, which causes problems when using the NATIVE_ALIASES

```
_mm256_loadu2_m128
_mm256_loadu2_m128d
_mm256_loadu2_m128i
_mm256_storeu2_m128d
_mm256_storeu2_m128
_mm256_storeu2_m128i
_mm256_zextps128_ps256
_mm256_zextpd128_pd256
_mm256_zextsi128_si256
```

The following functions need to be manually unrolled, ideally (but are probably fine as is)

```
AVX2:
_mm256_extract_epi8
_mm256_extract_epi16
AVX:
_mm256_extract_epi32
_mm256_extract_epi64
_mm256_insert_epi8
_mm256_insert_epi16
_mm256_insert_epi32
_mm256_insert_epi64
```

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -31,6 +31,9 @@
 HEDLEY_DIAGNOSTIC_PUSH
 SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
 
+#  if defined(SIMDE_AVX512F_NATIVE)
+#    undef SIMDE_AVX512F_NATIVE
+#  endif
 #  if defined(SIMDE_ARCH_X86_AVX512F) && !defined(SIMDE_AVX512F_NO_NATIVE) && !defined(SIMDE_NO_NATIVE)
 #    define SIMDE_AVX512F_NATIVE
 #  elif defined(SIMDE_ARCH_ARM_NEON) && !defined(SIMDE_AVX512F_NO_NEON) && !defined(SIMDE_NO_NEON)
@@ -274,7 +277,7 @@ typedef union {
 #if !defined(SIMDE_AVX512F_NATIVE) && defined(SIMDE_ENABLE_NATIVE_ALIASES)
   #define SIMDE_AVX512F_ENABLE_NATIVE_ALIASES
   typedef simde__m512 __m512;
-  typedef simde__m512i __m512i;
+  //typedef simde__m512i __m512i;
   typedef simde__m512d __m512d;
 #endif
 
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -3321,7 +3321,7 @@ simde_mm512_mask_test_epi32_mask (simde__mmask16 k1, simde__m512i a, simde__m512
   #endif
 }
 #if defined(SIMDE_AVX512F_ENABLE_NATIVE_ALIASES)
-  #define _mm512_mask_test_epi32_mask(a, b) simde_mm512_mask_test_epi32_mask(a, b)
+  #define _mm512_mask_test_epi32_mask(k1, a, b) simde_mm512_mask_test_epi32_mask(k1, a, b)
 #endif
 
 SIMDE__FUNCTION_ATTRIBUTES
@@ -3344,7 +3344,7 @@ simde_mm512_mask_test_epi64_mask (simde__mmask8 k1, simde__m512i a, simde__m512i
   #endif
 }
 #if defined(SIMDE_AVX512F_ENABLE_NATIVE_ALIASES)
-  #define _mm512_mask_test_epi64_mask(a, b) simde_mm512_mask_test_epi64_mask(a, b)
+  #define _mm512_mask_test_epi64_mask(k1, a, b) simde_mm512_mask_test_epi64_mask(k1, a, b)
 #endif
 
 SIMDE__FUNCTION_ATTRIBUTES
--- a/native-calls.c
+++ b/native-calls.c
@@ -1,0 +1,6 @@
+#define SIMDE_ENABLE_NATIVE_ALIASES
+#include "simde/x86/sse2.h"
+
+# _mm_stream_si64(__int64* mem_addr, __int64 a)
+
+# _mm_maskmove_si64(__m64 a, __m64 mask, char* mem_addr)
--- a/native-calls.c
+++ b/native-calls.c
@@ -1,6 +1,0 @@
-#define SIMDE_ENABLE_NATIVE_ALIASES
-#include "simde/x86/sse2.h"
-
-# _mm_stream_si64(__int64* mem_addr, __int64 a)
-
-# _mm_maskmove_si64(__m64 a, __m64 mask, char* mem_addr)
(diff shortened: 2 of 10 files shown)


junaruga:
This is a great catch!

It looks great if we can add your script or something to verify the alias in the current testing workflow with `CMakeLists.txt`.

After that, we might be able to remove the following note `it's recommended to only do this for testing`  in `README.md` if we can use it safely.

> If you define `SIMDE_ENABLE_NATIVE_ALIASES` before including SIMDe
> you can use the same names as the native functions.  Unfortunately,
> this is somewhat error-prone due to portability issues in the APIs, so
> it's recommended to only do this for testing.

mr-c:
@junaruga Thanks!

In my SIMDe writeup for Debian I recommend the use of the native alias to keep the patches more maintainable: https://wiki.debian.org/SIMDEverywhere

And then I found https://github.com/nemequ/simde/commit/b30129b3b48a6823013da2b309c50a081177b6b8 the hard way so I went looking for a more comprehensive solution :-D

nemequ:
This is awesome, thank you!

> After that, we might be able to remove the following note it's recommended to only do this for testing in README.md if we can use it safely.

Unfortunately this is still an issue.  The easiest example is all the functions that operate on 8-bit signed lanes.  Intel uses `char` instead of `int8_t`, but on most ARM platforms `char` is *unsigned*.  Code which uses SSE is likely to assume `char` is signed (since without SIMDe it's restricted to platforms where `char` is signed).  There are lots of small portability issues with the Intel APIs which make it unsafe to assume code will work as intended with SIMDe.

That said, we could probably soften the language a bit, and maybe add some documentation explaining some of the issues.

> In my SIMDe writeup for Debian I recommend the use of the native alias to keep the patches more maintainable: https://wiki.debian.org/SIMDEverywhere

I think that's the right call for Debian.  If upstream projects are willing to adopt SIMDe IMHO it would be better to use the `simde_*` functions since that will enlist the type system to help identify portability issues, but for patches carried by Debian (or any distro) I think it's more important to keep the delta small.

simde/x86/avx512f.h
```diff
@@ -274,7 +277,7 @@ typedef union {
 #if !defined(SIMDE_AVX512F_NATIVE) && defined(SIMDE_ENABLE_NATIVE_ALIASES)
   #define SIMDE_AVX512F_ENABLE_NATIVE_ALIASES
   typedef simde__m512 __m512;
-  typedef simde__m512i __m512i;
+  //typedef simde__m512i __m512i;
```
nemequ, line 280: ?

mr-c:
> I don't understand why 5522b21 is necessary. Are you running the tests after stripping off the simde prefix?

Yes, the script I linked to strip out the `simde` or `SIMDE` prefix 

Without 5522b21 I got failures in CPP mode (emul vs. native calls)

simde/x86/avx512f.h
mr-c: I got an error about about conflicting definitions that this fixed.

The AVX512* includes are very confusing to me

mr-c:
@nemequ The testing script is pretty gross, so tips on cleaning it up are welcome: https://gist.github.com/mr-c/90dbca3383037914dbdd5abf935231d9

Not quite ready to be fully automated yet

mr-c:
> I think that's the right call for Debian. If upstream projects are willing to adopt SIMDe IMHO it would be better to use the simde_* functions since that will enlist the type system to help identify portability issues, but for patches carried by Debian (or any distro) I think it's more important to keep the delta small.

Since my testing script generates a list of all the functions that have been implemented in SIMDe, a variation on that could be used to create a safer bulk renamer

nemequ:
I just created a `wip/native-aliases-test` branch.  It has an updated version of your patch to add the helpers (which, AFAICT, is all of this PR that hasn't been committed yet), and a rewritten script which uses the Intel Intrisics Guide to generate a list of functions.  I also added a test on GitHub Actions.

Let me know what you think.

mr-c:
@nemequ Thanks! They look good, modulo my comments about `SIMDE_*` on https://github.com/nemequ/simde/commit/abfca1951a158292a8631be9886a7078526c669f

mr-c:
@nemequ Adding `-march=native` to that test script raises a lot of issues: https://github.com/nemequ/simde/runs/642316073#step:7:50

nemequ:
> @nemequ Thanks! They look good, modulo my comments about `SIMDE_*` on [abfca19](https://github.com/nemequ/simde/commit/abfca1951a158292a8631be9886a7078526c669f)

Good point.  I think it's probably worth merging this even without those changes, but I'll try to find a way to cover the macros, too.

> @nemequ Adding `-march=native` to that test script raises a lot of issues: https://github.com/nemequ/simde/runs/642316073#step:7:50

Oh, good point.  I'll take a look in a few minutes; it's probably more stuff I broke with the simde-features.h changes.  I'm going to need to get an SDE build running on CI for stuff like GFNI anyways, so maybe I'll just run this is SDE, too; IIRC GH Actions don't even have AVX-512, much less GFNI, and it's going to be a while before they migrate to ICL+.

nemequ:
I just merged the rest of this, finally.  The `SIMDE_*` macros aren't done yet, but that can be addressed later.  I wanted to get this in ASAP since this test catches a lot of bugs which I obviously can't be trusted to spot when reviewing patches.

I found out what was forcing you to comment out the __m512* typedefs, see 030b05f2209f1464babc221063358ca86952b404 for an explanation. The good news is that it works, and all the tests pass with -mavx512f or -mavx. Unfortunately you'll get a ton of warnings right now about redefining macros; one for each "intrinsic" that the compiler implements as a macro.

I think all we can do here is add an undef before defining each native alias.  For example,

```diff
diff --git a/simde/x86/gfni.h b/simde/x86/gfni.h
index d1b494e..2e25a2a 100644
--- a/simde/x86/gfni.h
+++ b/simde/x86/gfni.h
@@ -121,6 +121,7 @@ simde_x_mm_gf2p8matrix_multiply_epi64_epi8 (simde__m128i x, simde__m128i A) {
   #define simde_mm_gf2p8affine_epi64_epi8(x, A, b) simde_mm_xor_si128(simde_x_mm_gf2p8matrix_multiply_epi64_epi8(x, A), simde_mm_set1_epi8(HEDLEY_STATIC_CAST(uint8_t, b)))
 #endif
 #if defined(SIMDE_X86_GFNI_ENABLE_NATIVE_ALIASES)
+  #undef _mm_gf2p8affine_epi64_epi8
   #define _mm_gf2p8affine_epi64_epi8(x, A, b) simde_mm_gf2p8affine_epi64_epi8(x, A, b)
 #endif
 
```

ISA extensions prior to up to and including AVX wouldn't require this, but all the extensions which are declared by including immintrin.h (except for AVX), including AVX2, FMA, AVX-512, GFNI, etc. would.

mr-c:
Thanks for taking this to completion, Evan!

PR #233, opened 2020-04-24
avx512bw, avx512f: test cases for _mm512_add_epi*,_mm512_castpd*_pd512

himanshi18037:
avx512bw, avx512f:  adding test cases for _mm512_add_epi*,_mm512_castpd*_pd512


himanshi18037:
> You're assuming that the upper bits (384 for castpd128_pd512, 256 for castpd256_pd512) are zero, but in reality they are undefined; they could be anything.
> 
> I'd suggest using `_mm512_maskz_mov_pd` to zero out the undefined bits in the test case before checking for equality.

Thanks, Hopefully, it should be correct after using _mm512_maskz_mov_pd.

PR #251, opened 2020-04-29
avx2: added copyright information in avx2.h

masterchef2209:
avx2: added copyright information in avx2.h


mr-c:
I'll merge this after updating it. Since this branch is on the main repo, it will be faster if we don't use a PR for things that don't need review as the CI runs twice otherwise

PR #253, opened 2020-04-29
avx512f.[hc]: add _mm512_slli_epi32 and _mm512_slli_epi64

rosbif:
In doing this I found that there are probably errors in the 128-bit and 256-bit versions.
They don't seem to handle out of range shift values correctly.
Should I correct them?


rosbif:
It is strange that there are two occurrences of -2147483648 (-2**31) in the test vectors.
I shall try to find out why.
But I don't agree with the compiler as -2147483648 is -2**31 which is (just) in range for an INT32.
The range for an INT32 is -2**31 to 2**31 - 1.

rosbif:
I plead "not guilty" of these failures ;-)

simde/x86/avx512f.h
```diff
@@ -3780,6 +3780,92 @@ simde_mm512_maskz_sub_pd(simde__mmask8 k, simde__m512d a, simde__m512d b) {
...
+  simde__m512i_private
+    r_,
+    a_ = simde__m512i_to_private(a);
+
+  /* The Intel Intrinsics Guide says that only the 8 LSBits of imm8 are
+   * used.  In this case we should do "imm8 &= 0xff".  However in
+   * practice all bits are used. */
+
+  /* The 128-bit and 256-bit versions don't seem to test for this and run-tests fails.
+   * I think that this is an error and that they should perform this test. */
```
nemequ, line 3838: I haven't tested this specific function, but in general the problem seems to be a mismatch in expectations.

In assembly, values outside the expected range ([0, 63] in this case) are either a compile-time error or are masked off by the CPU, I'm not entirely sure which.

The IIG specifies that the high bits are masked off, but that seems to just be a shorthand to say that the values must fall in the expected range.

Some compilers generate errors if the values don't fall in the expected range; clang tends to be more aggressive about this, but I've seen GCC do it, too.

Some compilers also allow values outside the expected range as sort of an extension.  The problem here is that there is no consistency across compilers.  I've run into situations where MSVC generates a different result from GCC/clang for these.

My position in SIMDe has basically been that values outside the expected range are undefined behavior, and we should add `HEDLEY_REQUIRE_MSG` macros to the functions to try to generate a compile time warning.

nemequ:
> But I don't agree with the compiler as -2147483648 is `-2**31` which is (just) in range for an INT32.
> The range for an INT32 is `-2**31` to `2**31 - 1`.
> Or is there something strange in the `INT32_C()`decoration?

Ah, I've fallen into this trap before.  The problem is that according to the C standard the constant isn't actually `-2147483648`, the constant is `2147483648`, which is then negated.  The relevant part of the C17 is § 6.4.4.1.  If you actually look at the definition for `INT32_MIN`, you'll see it is not `-2147483648` but `(-2147483647-1)` for this reason.

When test cases generate a value like this I just do a search and replace to substitute `INT32_MIN` (or `INT64_MIN`).

> I plead "not guilty" of these failures ;-)

Yeah, free CI can be flaky.  I restarted them, looks like they're all passing now.

FWIW, you should have permission to restart failures on GitHub Actions if you want.  For Travis I'm not sure being listed as a collaborator on the project is enough or not, but I don't see a way to add additional users so hopefully it is.

I'll go ahead and merge this, but I think I'm going to add [`HEDLEY_REQUIRE_MSG`](https://nemequ.github.io/hedley/api-reference.html#HEDLEY_REQUIRE_MSG) first.

nemequ:
I just did a bit more research on this, and it seems like as of avx512 there is finally some consistency here.  And [MSVC actually generates](https://godbolt.org/z/27pju3) a `vpslld zmm, zmm, imm8` for values up to 127 (and `vpslld zmm, zmm, xmm` for values > 127) and the result is correct, so I think for AVX-512 we can actually rely on this behavior.  No need to restrict the range, though I am going to remove the comments about previous instruction sets not testing this being wrong; unfortunately there is a reason :(

Thanks for the patch, it should be hitting master soon!

simde/x86/avx512f.h
rosbif: > Some compilers generate errors if the values don't fall in the expected range; clang tends to be more aggressive about this, but I've seen GCC do it, too.
> 
> Some compilers also allow values outside the expected range as sort of an extension. The problem here is that there is no consistency across compilers. I've run into situations where MSVC generates a different result from GCC/clang for these.
> 

While compilers may test in the case of compile-time constants I doubt that they would insert a run-time test. This would slow down code just to handle a case that probably shouldn't happen (if it does it probably indicates a bug in the code).

> My position in SIMDe has basically been that values outside the expected range are undefined behavior, and we should add `HEDLEY_REQUIRE_MSG` macros to the functions to try to generate a compile time warning.
> 

Yes, I agree. Not only are there differences between compilers but it is also possible that Intel and AMD processors treat these grey areas differently (or even different generations of Intel processors).
However in our previous discussion around here https://github.com/nemequ/simde/pull/150#discussion_r403509435 you suggested that I introduce tests for out of range values.
If we now agree to treat these as undefined, should I remove these tests?
In fact if you add the Hedley macro to generate a warning, this will cause failure on compilations which treat warnings as errors and so I MUST remove the tests.
However there is one out of range value, the exact size of the type, which I feel is useful (see below). Should I at least test this?

> _Edit_: it's also worth noting that right shifting a signed integer by more than the width of the type is undefined behavior in C.
> 

IIRC in K&R it was even worse than this:
- ANY right shift of an signed type was undefined. I guess that this may have been because the DEC PDP11 (for which they developed C IIRC) may only have had instructions for logical right shifts. However luckily all machines I have used generate an arithmetic right shift.
- Any shift by the exact width of the type was also undefined. Indeed many machines take the shift modulo the width of the type. This means that a UINT32 shifted by 32 bits doesn't disappear to zero but is unchanged.

I remember being once bitten by this. I wrote a left rotate of a UINT32 u by n bits as:
r = (u << n) | (u >> (32 - n));
This fails for n = 0 as it generates a 32-bit right shift which generally leaves u unchanged instead disappearing to zero.

I hope that this has been improved since K&R because arithmetic right shifts are very useful and a shift by the exact word size can be useful too.

rosbif:
@nemequ : Thank you Evan for your help.
Does the style of my additions to avx512f.h (indentation of pre-processor directives and general layout) correspond to what you had in mind?

simde/x86/avx512f.h
nemequ: > However in our previous discussion around here #150 (comment) you suggested that I introduce tests for out of range values.

You're right, that does seem inconsistent.  The issues you encountered helped convince me to treat these as undefined, which is why I suggested what I did here.  However, as I mentioned in [another comment](https://github.com/nemequ/simde/pull/253#issuecomment-621489223), after looking into this a bit more I think that for this particular function it actually makes sense to support values outside the expected range.

Take a look at <https://godbolt.org/z/_5Cku9>.

 * If `imm8` is within the expected ([0, 63]) range it's what you'd expect the instruction for this intrinsic to be: `vpslld zmm, k, zmm, i8`.  
 * If `imm8` is in [64, 127], MSVC will continue to generate the same code as the version in the expected range.  That's fine; on x86/x86_64 we know the result will always be 0.
 * If `imm8` is > 127, MSVC will load the int into the first lane of an `__m128i` and do a `_mm512_sll_epi32`.  For values > 64 the result will always be zero.
 * Knowing that the result will alwasy be 0, Clang, GCC, and ICC will all just xor the input with itself (the same thing `_mm512_setzero_si512` does) if `imm8` > 63.

So basically, *nobody*, not even Intel, actually does what the Intrinsics Guide says.  But everyone does generate the same results.

I think we're just going to have to treat these on a case-by-case basis where if compilers actually agree we can operate that way in SIMDe.

> While compilers may test in the case of compile-time constants I doubt that they would insert a run-time test. This would slow down code just to handle a case that probably shouldn't happen (if it does it probably indicates a bug in the code).

They don't really test.  That Compiler Explorer link above also has a "unknown" version if you want to see for yourself.  If they don't know the value at compile time the compiler will just use `_mm512_sll_epi16(a, _mm_cvtsi32_si128(imm8))` instead.

The problem is that the immediate version is *way* faster.  On Icy Lake, just for the shift we're talking about a latency of up to 8 cycles vs 1 for the immediate mode.

> IIRC in K&R it was even worse than this:

Ouch!  I didn't realize it was that bad :(

nemequ:
Yep, the coding style looked good.  IIRC the only thing I changed was removing those two comments, and that's only because there is actually a reason for the behavior in other functions; if it weren't for compilers generating inconsistent results I'd completely agree with that, too.  Thanks again for the patch!

PR #254, opened 2020-04-29
avx512dq: implement mm512_{mask,maskz}_andnot_{ps,pd}

himanshi18037:


nemequ:
I'm pretty sure it was a NaN propagation issue, so I ended up having to just finish implementing the integer versions, then used them to implement the floating point ones.  37440b4e910f46359560871be8862e4103fdf37b has a short explanation, but let me know if you have any questions.

I kept your test cases and merged them as 228cac4f9d5cf827bb9aff8ad798588b34c7850c.

Thanks for the patch!

PR #257, opened 2020-05-01
avx512f: test cases for mm512_set4_{epi32,epi64,ps,pd}

himanshi18037:


nemequ:
I already fixed it, but test_simde_mm512_set4_epi{32,64} were missing the UINT32_C/UINT64_C macros on the inputs.

PR #259, opened 2020-05-02
avx512f: added the implementation for mm512_{mask,maskz}_cvtsepi64_epi32

masterchef2209:
I don't understand why this fails on appveyor, please review this.


test/x86/avx512f.c
```diff
@@ -10493,6 +10493,176 @@ test_simde_mm512_cvtsepi64_epi32(const MunitParameter params[], void* data) {
...
+    simde__m256i r;
+  } test_vec[8] = {
+    { simde_mm256_set_epi32(INT32_C( -783872818), INT32_C(  224826276), INT32_C(  833953142), INT32_C(  704393899),
+                            INT32_C(-1282792525), INT32_C(-1682931810), INT32_C( 1715663188), INT32_C(-1736532826)),
+      UINT8_C(184),
+      simde_mm512_set_epi64(INT64_C( -304424330925858839), INT64_C( 6920061078002113788),
+                            INT64_C( 5392365268250659625), INT64_C( 7451707948021880068),
+                            INT64_C(-8094918606566868518), INT64_C( 4039990350021559518),
+                            INT64_C(-5079608809355355713), INT64_C( 4544449552448075830)),
+      simde_mm256_set_epi32(INT32_C(-2147483648), INT32_C(  224826276), INT32_C( 2147483647), INT32_C( 2147483647),
```
nemequ, line 10515: -2147483648 is -2³¹, which is `INT32_MIN`.

According to § 6.,4,.,1 of [the C standard](https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf), the unary minus isn't actually part of the integer constant.  What's really going on here is that you have an integer constant with a value of 2147483648, which is then negated.

However, `2147483648` is actually outside the range of a signed 32-bit integer (`INT32_MAX` is 2147483647).  Since it doesn't fit in a signed 32-bit integer the compiler is treating it as an *unsigned* 32-bit integer*.  You then negate the unsigned 32-bit integer, but that doesn't change the underlying type; the negative of an unsigned integer is still unsigned (and therefore still positive), which is what MSVC is warning you about.

The solution here is to replace every occurrence  of `INT32_C(-2147483648)` in the generated code with `          INT32_MIN `.  Interestingly, if you look at a copy of `stdint.h` you can see that `INT32_MIN` is typically defined as `(-2147483647-1)`, not `-2147483648`, to avoid this issue.

\* Depending on the type of `int32_t`, another compiler may have actually treated it as a signed 64-bit integer, but on Windows `int32_t` is an `int` not a `long int`, so `INT32_C()` doesn't add a suffix, which means the type is the first which can contain the value from `int`, `long int`, and `long long int`.  On platforms where `int32_t` is a `long` `INT32_C` will add an "L" or "l" suffix, so a different cell in that table would apply and you'd end up with a `long long int`.

masterchef2209: This is very interesting, I will remember to use defined values for this case. 
I have made the changes in new commit but it is still failing on icc, please review it.
Thank You

nemequ: Looks like it was just a temporary problem installing ICC; I restarted the build and it worked.

FWIW, you should have access to restart builds on Travis if you log-in with GitHub.  My understanding is that as long as you have write access to the repo Travis should let you restart builds.  I don't mind doing it if you prefer, but this way you don't have to wait.

Issue #261, opened 2020-05-04
Big-endian tests


rosbif:
This is not really an issue but maybe someone can help.
I am developing some code which I suspect may have endian problems.
I want to test this code locally on a big-endian platform without creating a PR and generating the noise associated with likely s390x failures.
I am running Ubuntu 20.04 and  I have tried its qemu 4.2.0 on four platforms.
Unfortunately it dumps core on both the big-endian platforms I tried, even on a classic hello.c :-(
```
On s390x qemu dumps core
+ s390x-linux-gnu-gcc hello.c
+ qemu-s390x -L /usr/s390x-linux-gnu a.out
./testHello: line 6:  3304 Segmentation fault      (core dumped) qemu-s390x -L /usr/s390x-linux-gnu a.out
+ set +x

On aarch64 OK
+ aarch64-linux-gnu-gcc hello.c
+ qemu-aarch64 -L /usr/aarch64-linux-gnu a.out
Hello world!
+ set +x

On mips64 qemu dumps core
+ mips64-linux-gnuabi64-gcc hello.c
+ qemu-mips64 -L /usr/mips64-linux-gnuabi64 a.out
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
./testHello: line 18:  3322 Segmentation fault      (core dumped) qemu-mips64 -L /usr/mips64-linux-gnuabi64 a.out
+ set +x

On mips64el OK
+ mips64el-linux-gnuabi64-gcc hello.c
+ qemu-mips64el -L /usr/mips64el-linux-gnuabi64 a.out
Hello world!
+ set +x

```
Has anyone managed to get this to work?
(Remark: I also tried docker but ran into a newbie problem and didn't insist.)

nemequ:
I had a similar problem with s390x on Debian for a while, but it works now.  Maybe the fixes have yet to percolate out to Ubuntu?

For what it's worth, I run Fedora but I have a Debian Unstable VM with all the cross compilers installed (Debian's support for cross-compilation is much better); if you have spare RAM it's a decent way to go.  I just mount the home directory using sshfs and ssh in, so it feels very native.

Another option is that you could sign up for Travis (it's free) for your fork.  I actually have [https://github.com/nemequ/ci-noise](a repo) which I use when I want to run stuff through CI without generating a ton of noise in SIMDe (or whatever I'm working on).  You can even temporarily comment out builds other than the one you're interested in (e.g., s390x) in .travis.yml so it doesn't take forever.

[...]

Issue #169, opened 2020-04-09
Add Galois Field New Instructions


rosbif:
I would like to add GFNI to SIMDe.

When I first looked at GFNI I thought that these intrinsics were useless for me as many are hardwired for the (IMHO stupid) reduction polynomial x^8 + x^4 + x^3 + x + 1 (0x11B) probable inherited from AES, whereas I need 0x11D and 0x187.
(I say stupid because, although irreducible, their polynomial is not primitive. This causes problems for some use cases.)
Yesterday I decided to have a closer look at these intrinsics and found out how I could use them profitably.

According to the IIG some intrinsics depend on GFNI + AVX512F and others on GFNI + AVX512VL.
I figure I should put them in gfni.[hc].
What do you think?
Would you accept patches?

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#avx512techs=GFNI

Also requires AVX512F:
 
 - [ ] _mm512_gf2p8affine_epi64_epi8
 - [ ] _mm512_mask_gf2p8affine_epi64_epi8
 - [ ] _mm512_maskz_gf2p8affine_epi64_epi8
 - [ ] _mm512_gf2p8affineinv_epi64_epi8
 - [ ] _mm512_mask_gf2p8affineinv_epi64_epi8
 - [ ] _mm512_maskz_gf2p8affineinv_epi64_epi8
 - [ ] _mm512_gf2p8mul_epi8
 - [ ] _mm512_mask_gf2p8mul_epi8
 - [ ] _mm512_maskz_gf2p8mul_epi8
 
Also requires AVX512VL:
 
 - [ ] _mm_gf2p8affine_epi64_epi8
 - [ ] _mm_mask_gf2p8affine_epi64_epi8
 - [ ] _mm_maskz_gf2p8affine_epi64_epi8
 - [ ] _mm256_gf2p8affine_epi64_epi8
 - [ ] _mm256_mask_gf2p8affine_epi64_epi8
 - [ ] _mm256_maskz_gf2p8affine_epi64_epi8
 - [ ] _mm_gf2p8affineinv_epi64_epi8
 - [ ] _mm_mask_gf2p8affineinv_epi64_epi8
 - [ ] _mm_maskz_gf2p8affineinv_epi64_epi8
 - [ ] _mm256_gf2p8affineinv_epi64_epi8
 - [ ] _mm256_mask_gf2p8affineinv_epi64_epi8
[...]

PR #265, opened 2020-05-05
Add accelerated fallbacks and use them

rosbif:
Add functions contain fallbacks to accelerate epi8 mask and maskz intrinsics without native support.
The 128-bit and 256-bit versions are in avx512vl.h but they are useful for me in avx512bw.h.
So make them simde_x_ in avx512bw.h and just call them in avx512vl.h.

Add the following intrinsics
_mm512_mask_add_epi8
_mm512_maskz_add_epi8
_mm512_mask_adds_epi8
_mm512_maskz_adds_epi8
_mm512_mask_adds_epu8
_mm512_maskz_adds_epu8
_mm512_mask_shuffle_epi8
_mm512_maskz_shuffle_epi8
_mm512_cmpge_epi8_mask
_mm512_cmpge_epu8_mask
_mm512_cmpgt_epi8_mask
_mm512_cmpgt_epu8_mask
_mm512_cmple_epi8_mask
_mm512_cmple_epu8_mask
_mm512_cmplt_epi8_mask
_mm512_cmplt_epu8_mask


rosbif:
I executed the Travis-CI tests successfully in my fork as Evan suggested.
I have concentrated on the mask conversions for epi8 because they are the most time-consuming (and this is also the case in which I am most interested).
If these are accepted, I shall do similar conversions for epi16, epi32, epi64, ps and pd cases.

nemequ:
Definite yes to the basic idea.  I'd think it can be a bit simpler, though..

I recently changed the preprocessor macros for feature detection for exactly this reason.  All that stuff is now in `simde-features.h`, so no matter where you are in the codebase you should have access to the right macro to detect which features you're allowed to use.  So, you don't need to put them in `simde_x_*…` just so you don't conflict with a version from a later ISA extension.

I don't think there is a need to be overly strict about functions being in the same file as the ISA extension they're part of.  As long as, for example, #including "avx512vl.h" pulls in all the AVX-512VL functions we're good, and avx512vl.h already includes avx512bw.h.

So, for example, something like this should go in avx512bw.h (or even in avx512f.h if there is something there which could make use of them):

```c
SIMDE__FUNCTION_ATTRIBUTES
simde__m128i
simde_mm_movm_epi8 (simde__mmask16 k) {
  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512BW_NATIVE)
    return _mm_movm_epi8(k);
  #elif defined(SIMDE_X86_SSSE3_NATIVE)
    const simde__m128i zero = simde_mm_setzero_si128();
    const simde__m128i bits = simde_mm_set_epi16(0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80);
    const simde__m128i shuffle = simde_mm_set_epi8(15, 13, 11, 9, 7, 5, 3, 1, 14, 12, 10, 8, 6, 4, 2, 0);
    simde__m128i r;

    r = simde_mm_set1_epi16(HEDLEY_STATIC_CAST(short, k));
    r = simde_mm_mullo_epi16(r, bits);
    r = simde_mm_shuffle_epi8(r, shuffle);
    r = simde_mm_cmpgt_epi8(zero, r);

    return r;
  #else
    simde__m128i_private r_;

    SIMDE__VECTORIZE
    for (size_t i = 0 ; i < (sizeof(r_.i8) / sizeof(r_.i8[0])) ; i++) {
      r_.i8[i] = ((k >> i) & 1) ? INT8_C(0xFF) : INT8_C(0);
    }

    return simde__m128i_from_private(r_);
  #endif
}
```

(Note that I also added the native implementation, and changed the macros to `SIMDE_X86_*_NATIVE` to match the new feature detection code.)

That should let us use the fastest available implementation everywhere.

As a side note, thanks for pointing out the *_movm_* functions.  I didn't know those existed, I'm going to want to refactor some stuff in SIMDe to make use of them.

We're going to need test cases for all these functions before they are merged; if you'd like I can take care of that, or maybe @himanshi18037 would be interested (she has been adding a ton of tests latelly that Michael noticed were missing).

Other than that, I only see a couple very minor nit-picks in the code.  There is a note in `simde_mm512_cmpgt_epi8_mask` about using Hedley for the cast; please do.  I should really add `-Wold-style-casts` to the CXXFLAGS for clang to prevent these from creeping in, as I too have a tendency to use C-style casts.  Also, you use 1ULL in a couple places; that should probably be `UINT64_C(1)` if it's really a 64-bit unsigned integer (which is what __mmask64 is) that you want.

rosbif:
> Definite yes to the basic idea. I'd think it can be a bit simpler, though..
> 
> I recently changed the preprocessor macros for feature detection for exactly this reason. All that stuff is now in `simde-features.h`, so no matter where you are in the codebase you should have access to the right macro to detect which features you're allowed to use. So, you don't need to put them in `simde_x_*…` just so you don't conflict with a version from a later ISA extension.
> 
> I don't think there is a need to be overly strict about functions being in the same file as the ISA extension they're part of. As long as, for example, #including "avx512vl.h" pulls in all the AVX-512VL functions we're good, and avx512vl.h already includes avx512bw.h.
> 

OK. Unfortunately I made the extra effort to avoid putting the real version in the wrong header file :-(

> So, for example, something like this should go in avx512bw.h (or even in avx512f.h if there is something there which could make use of them):
> 
> ```c
> SIMDE__FUNCTION_ATTRIBUTES
> simde__m128i
> simde_mm_movm_epi8 (simde__mmask16 k) {
>   #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512BW_NATIVE)
>     return _mm_movm_epi8(k);
>   #elif defined(SIMDE_X86_SSSE3_NATIVE)
>     const simde__m128i zero = simde_mm_setzero_si128();
>     const simde__m128i bits = simde_mm_set_epi16(0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80);
>     const simde__m128i shuffle = simde_mm_set_epi8(15, 13, 11, 9, 7, 5, 3, 1, 14, 12, 10, 8, 6, 4, 2, 0);
>     simde__m128i r;
> 
>     r = simde_mm_set1_epi16(HEDLEY_STATIC_CAST(short, k));
>     r = simde_mm_mullo_epi16(r, bits);
>     r = simde_mm_shuffle_epi8(r, shuffle);
>     r = simde_mm_cmpgt_epi8(zero, r);
> 
>     return r;
>   #else
>     simde__m128i_private r_;
> 
>     SIMDE__VECTORIZE
>     for (size_t i = 0 ; i < (sizeof(r_.i8) / sizeof(r_.i8[0])) ; i++) {
>       r_.i8[i] = ((k >> i) & 1) ? INT8_C(0xFF) : INT8_C(0);
>     }
> 
>     return simde__m128i_from_private(r_);
>   #endif
> }
> ```
> 
> (Note that I also added the native implementation, and changed the macros to `SIMDE_X86_*_NATIVE` to match the new feature detection code.)
> 
> That should let us use the fastest available implementation everywhere.
> 

Oops, sorry I left out the _X86 in the macros.
Incidentally, for this reason personally I hate using `#if defined(x)`.
I obviously followed your practice but I prefer `#if x` (with x defined to 1 or 0), which, with the `-Wundef` compilation flag, throws a warning if you misspell the name of the macro.

As for the test whether the real intrinsic can simply be used, I used to do it like this inside the function as seemed to me to be common practice, but you told me to put it outside and after the function.
(Sorry, I can no longer find where you wrote this.)

> We're going to need test cases for all these functions before they are merged; if you'd like I can take care of that, or maybe @himanshi18037 would be interested (she has been adding a ton of tests latelly that Michael noticed were missing).
> 

Which test cases were missing? I thought I had covered all the intrinsics I added.

> Other than that, I only see a couple very minor nit-picks in the code. There is a note in `simde_mm512_cmpgt_epi8_mask` about using Hedley for the cast; please do. I should really add `-Wold-style-casts` to the CXXFLAGS for clang to prevent these from creeping in, as I too have a tendency to use C-style casts. Also, you use 1ULL in a couple places; that should probably be `UINT64_C(1)` if it's really a 64-bit unsigned integer (which is what __mmask64 is) that you want.
>

Sorry, I think I just copied the code with 1ULL from an existing example.

nemequ:
> Incidentally, for this reason personally I hate using `#if defined(x)`.
> I obviously followed your practice but I prefer `#if x` (with x defined to 1 or 0), which, with the `-Wundef` compilation flag, throws a warning if you misspell the name of the macro.

Yeah, I do that sometimes but I don't like the inconsistency.  You generally need to use defined() when interfacing with other code, so you end up with a mix of both ways, then you have to remember which code works which way.

> As for the test whether the real intrinsic can simply be used, I used to do it like this inside the function as seemed to me to be common practice, but you told me to put it outside and after the function.
> (Sorry, I can no longer find where you wrote this.)

I don't have the link handy, but it was probably when were were discussing something like:

```c
#if SIMDE_X86_FOO_NATIVE
  #define simde_mm_foo(x) _mm_foo(x, imm8)
#else
  void simde_mm_foo(simde__m128 x, int imm8) {
    ...
  }
#endif
```



Inside the function is preferable, but outside the function is the only option if there is an immediate-mode argument.  For a lot of compilers you can't even use compile time constants (in the `__builtin_constant_p()` sense), they actually have to be integer constant expressions.  In that case, for SIMDe the convention a function-like macro *after* the function is declared.

If you want to just leave it as-is I don't mind fixing it up when I merge.

> > We're going to need test cases for all these functions before they are merged; if you'd like I can take care of that, or maybe @himanshi18037 would be interested (she has been adding a ton of tests latelly that Michael noticed were missing).
> 
> Which test cases were missing? I thought I had covered all the intrinsics I added.

Maybe you forgot to add avx512vl.c to the PR?

> > Other than that, I only see a couple very minor nit-picks in the code. There is a note in `simde_mm512_cmpgt_epi8_mask` about using Hedley for the cast; please do. I should really add `-Wold-style-casts` to the CXXFLAGS for clang to prevent these from creeping in, as I too have a tendency to use C-style casts. Also, you use 1ULL in a couple places; that should probably be `UINT64_C(1)` if it's really a 64-bit unsigned integer (which is what __mmask64 is) that you want.
> 
> Sorry, I think I just copied the code with 1ULL from an existing example.

> Do you want me to perform these modifications or will you do it prior to merging?

I can take care of it, unless they're in your way in which case feel free.

rosbif:
> > As for the test whether the real intrinsic can simply be used, I used to do it like this inside the function as seemed to me to be common practice, but you told me to put it outside and after the function.
> > (Sorry, I can no longer find where you wrote this.)
> 
> I don't have the link handy, but it was probably when were were discussing something like:
> 
> ```c
> #if SIMDE_X86_FOO_NATIVE
>   #define simde_mm_foo(x) _mm_foo(x, imm8)
> #else
>   void simde_mm_foo(simde__m128 x, int imm8) {
>     ...
>   }
> #endif
> ```
> 
> Inside the function is preferable, but outside the function is the only option if there is an immediate-mode argument. For a lot of compilers you can't even use compile time constants (in the `__builtin_constant_p()` sense), they actually have to be integer constant expressions. In that case, for SIMDe the convention a function-like macro _after_ the function is declared.
> 

Sorry I completely misunderstood you on this.
I thought you wanted me to code like this systematically.
Which is what I did :-(
I imagined that your reason was that, like this, a bit more of the function was compile-tested.
I think that I pointed out at the time that it was different from common practice in SIMDe.
I even asked here https://github.com/nemequ/simde/pull/253#issuecomment-621735126 whether this coding style was what you expected.

> If you want to just leave it as-is I don't mind fixing it up when I merge.
> 
> > > We're going to need test cases for all these functions before they are merged; if you'd like I can take care of that, or maybe @himanshi18037 would be interested (she has been adding a ton of tests latelly that Michael noticed were missing).
> > 
> > 
> > Which test cases were missing? I thought I had covered all the intrinsics I added.
> 
> Maybe you forgot to add avx512vl.c to the PR?
> 

Probably I'll check.

> > Do you want me to perform these modifications or will you do it prior to merging?
> 
> I can take care of it, unless they're in your way in which case feel free.
> 

I am not sure what time zone you are in. Are you in the USA? I am in Europe as is Michael, I think.
It is the morning for me so I could do it.
I am also slightly worried that with the _X86 modifications my fallbacks weren't tested: I was expecting endian problems but there weren't any, perhaps becaus my fallbacks weren't used.

nemequ:
> Sorry I completely misunderstood you on this.
> I thought you wanted me to code like this systematically.
> Which is what I did :-(

Sorry I should have been clearer.  Again, if you want to just leave these as-is I don't mind making the changes so you can keep working on the more interesting bits.

> I imagined that your reason was that, like this, a bit more of the function was compile-tested.
> I think that I pointed out at the time that it was different from common practice in SIMDe.
> I even asked here [#253 (comment)](https://github.com/nemequ/simde/pull/253#issuecomment-621735126) whether this coding style was what you expected.

In this case it is what I was expecting; the `imm8` argument to `_mm512_slli_epi32` is supposed to be an immediate mode argument (hence "imm8").  OTOH, for `_mm512_sll_epi32` it should go inside the function.  I don't know if compilers actually require it to be an immediate made argument (i.e., issue 267), but according to the IIG it should be.

> Probably I'll check.
> Edit: As I thought when I pushed, all my changes in avx512vl.h seem to be covered by existing tests in avx512vl.c. Can you see any that aren't covered?
> /Edit

I don't see any of the `movm` functions.  `simde_mm512_movm_epi8`, for example.

> I am not sure what time zone you are in. Are you in the USA? I am in Europe as is Michael, I think.
> It is the morning for me so I could do it.

PDT (-0700).  I'm in San Diego.  *Now* it's morning ;)

I'll hold off until this is in so I don't create conficts.  If you want to add those changes as part of this PR go for it, otherwise I'll take care of it after I merge.

> I am also slightly worried that with the _X86 modifications my fallbacks weren't tested: I was expecting endian problems but there weren't any, perhaps because my fallbacks weren't used.

You're right that the macros are wrong in gfni.h (I'll fix that it a couple minutes), but that doesn't really matter here since the only version that will ever run right now on BE are the portable fallbacks.  It will never fall back on, for example, SSE2, since SSE2 is only implemented on x86 which is only ever LE.  Additionally, the `SIMDE_ARCH_*` macros are still there and do the same thing as before.

nemequ:
I'm going to go ahead and merge everything and do all the cleanup I can now.  I may not merge it into master quite yet, but you should have a fresh, conflict-free branch in the morning to base future work on.

rosbif:
It looks like we are both working in my fork.
I must have forgotten to remove some of the stuff in avx512vl.h resulting in duplicates.
I don't know why this didn't show up in my local tests.
So I'll hold off for now.
I have already pushed several additions (mullo in AVX2 which I needed and missing movm tests) and corrections.

mr-c:
> I must have forgotten to remove some of the stuff in avx512vl.h resulting in duplicates. 

I think that was caused by a bad merge from me ; sorry!

rosbif:
The rosbif / simde merge branch just passed all the Travis-CI tests :-)

nemequ:
I merged most of this in one form or another.  The only exception I'm aware of several of the `_mm512_cmp*` implementations, which are wrong.

I also switched the headers so avx512bw.h includes avx512vl.h instead of the other way around.  See b84c9e4cdc0238d884f706ca23d491f3781aa876.

I'll leave this open in case you want to use it for the remaining comparison functions, but you may want to just close it and start a new one.

rosbif:
> I merged most of this in one form or another. The only exception I'm aware of several of the `_mm512_cmp*` implementations, which are wrong.
> 

Sorry.
Strange: I am sure that I cleaned these up. Maybe just in a local version (which I can no longer find).
I'll push corrections.

> I also switched the headers so avx512bw.h includes avx512vl.h instead of the other way around. See [b84c9e4](https://github.com/nemequ/simde/commit/b84c9e4cdc0238d884f706ca23d491f3781aa876).
> 

NAK: unless you have reorganized the movm, mask_mov, and maskz_mov stuff, avx512vl.h needs them from avx512bw.h. So avx512vl.h must include avx512bw.h.
Or am I being stupid again?

> I'll leave this open in case you want to use it for the remaining comparison functions, but you may want to just close it and start a new one.

I think I prefer to leave it open.

nemequ:
> NAK: unless you have reorganized the movm, mask_mov, and maskz_mov stuff, avx512vl.h needs them from avx512bw.h. So avx512vl.h must include avx512bw.h.
> Or am I being stupid again?

Not at all. The movm functions are in avx512vl.h.  See 0d3609ed99109ee6fd020d35120b9059d17031f6.  I also converted a lot of function-like macros to be calls inside of of their respective functions.  For example, where you [had something](https://github.com/nemequ/simde/pull/265/files#diff-4999781dbac10153754ea0bdcb4126d4R76)

```c
#if defined(SIMDE_X86_AVX512BW_NATIVE) && defined(SIMDE_X86_AVX512VL_NATIVE)
  #define simde_mm_mask_mov_epi8(src, k, a) _mm_mask_mov_epi8(src, k, a)
#else
  #define simde_mm_mask_mov_epi8(src, k, a) simde_mm_blendv_epi8(src, a, simde_mm_movm_epi8(k))
#endif
#if defined(SIMDE_X86_AVX512VL_ENABLE_NATIVE_ALIASES)
  #define _mm_mask_mov_epi8(src, k, a) simde_mm_mask_mov_epi8(src, k, a)
#endif
```

I just did

```diff
diff --git a/simde/x86/avx512vl.h b/simde/x86/avx512vl.h
index d35d20f..4fce874 100644
--- a/simde/x86/avx512vl.h
+++ b/simde/x86/avx512vl.h
@@ -122,8 +122,10 @@ simde_mm512_movm_epi8 (simde__mmask64 k) {
 SIMDE__FUNCTION_ATTRIBUTES
 simde__m128i
 simde_mm_mask_mov_epi8(simde__m128i src, simde__mmask16 k, simde__m128i a) {
-  #if defined(SIMDE_X86_AVX512VL_NATIVE)
+  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512BW_NATIVE)
     return _mm_mask_mov_epi8(src, k, a);
+  #elif defined(SIMDE_X86_SSE4_1_NATIVE)
+    return simde_mm_blendv_epi8(src, a, simde_mm_movm_epi8(k));
   #else
     simde__m128i_private
       src_ = simde__m128i_to_private(src),
@@ -138,7 +140,7 @@ simde_mm_mask_mov_epi8(simde__m128i src, simde__mmask16 k, simde__m128i a) {
     return simde__m128i_from_private(r_);
   #endif
 }
```

Here are the relevant commits:

 * 0d3609ed99109ee6fd020d35120b9059d17031f6
 * 5702c5a6a0a8d1380c95ded6f1886f304688beda
 * 2b9751f70f1cac2dbf1e70e142b607c7b056e172
 * 7ae207836b90f6be69ce0ef888cc39e4cdeb2711
 * 27917ad33bd855834a118dee31d60f5b7a533fc2
 * b689653a0359c00e2bd068b7b41a62cfb9d77338
 * b3fa2c21ccdad9d6cae012914f27e56bdd1d16bf
 * c4f87426c50b92bc33b5dcef09abbaaaf5fbc609

If there are still changes you want to make (such as adding simde_mm512_cmple_epi8_mask, simde_mm512_cmple_epu8_mask, simde_mm512_cmplt_epi8_mask, and simde_mm512_cmplt_epu8_mask) you should rebase them on top of master (currently 1582d2f8c83c64cef0f0d37edd185b2d9f66a85b).

As a cosequence of reorganizing everything several minor issues, like incorrect macro names or the typo you fixed with dad33587f291bdd4b6011ff90fe057f4399bebb1, should disappear.

> I think I prefer to leave it open.

Okay, in that case it would probably be best to force push to your "merge" branch after rebasing your changes on top of current master (or at least master post-c4f87426c50b92bc33b5dcef09abbaaaf5fbc609).  That way we'll be working from the same place.  If you want some help (IIRC you said you weren't particularly familiar with git) let me know and I can help you through the process on some kind of chat platform (Gitter or Signal maybe?), I don't think the issue tracker would be a good place only because it's not designed for a real-time back-and-forth.

rosbif:
OK, I'll trust you on the avx512bw.h vs avx512vl.h question.

> If there are still changes you want to make (such as adding simde_mm512_cmple_epi8_mask, simde_mm512_cmple_epu8_mask, simde_mm512_cmplt_epi8_mask, and simde_mm512_cmplt_epu8_mask) you should rebase them on top of master (currently [1582d2f](https://github.com/nemequ/simde/commit/1582d2f8c83c64cef0f0d37edd185b2d9f66a85b)).
> 

I am not sure what I should do. here
I pushed these updated cmp intrinsics to my fork about 9 hours ago.
From what you say I suppose you removed them before merging to master.
In that case I suppose you must also have removed their tests.

> > I think I prefer to leave it open.
> >

(I only said that because I wanted to push the above cmp intrinsics before you closed everything.)

> Okay, in that case it would probably be best to force push to your "merge" branch after rebasing your changes on top of current master (or at least master post-c4f87426c50b92bc33b5dcef09abbaaaf5fbc609). That way we'll be working from the same place. If you want some help (IIRC you said you weren't particularly familiar with git) let me know and I can help you through the process on some kind of chat platform (Gitter or Signal maybe?), I don't think the issue tracker would be a good place only because it's not designed for a real-time back-and-forth.
>

OK, I'll try to do this.

rosbif:
I am lost :-(
In master I find lots of e.g. SIMDE_X86_SSE2_NATIVE in avx512bw.h
I thought you had said to use SIMDE_ARCH_X86_SSE2.
I changed all mine to this style.
Have you reverted it?

I also see a SIMDE_X166_AVX512BW_NATIVE which is obviously a result of s/8/16/ :-(
Again if you used #if X instead of #if defined(X) this could be picked up on compilation.

nemequ:
> In master I find lots of e.g. SIMDE_X86_SSE2_NATIVE in avx512bw.h
> I thought you had said to use SIMDE_ARCH_X86_SSE2.
> I changed all mine to this style.

There are three macros here:

 * `SIMDE_ARCH_X86_SSE2`
 * `SIMDE_SSE2_NATIVE`
 * `SIMDE_X86_SSE2_NATIVE`

`SIMDE_ARCH_X86_SSE2` does the same thing it used to do, which is detecting what the compiler thinks is available (i.e., it's basically like `__SSE2__` except it works on MSVC too).

To figure out which functionality we actually want to use in SIMDe you should use `SIMDE_X86_SSE2_NATIVE`, which uses `SIMDE_ARCH_X86_SSE2` internally but can also be disabled by defining `SIMDE_X86_SSE2_NO_NATIVE` or `SIMDE_NO_NATIVE` or forcibly enabled by defininig `SIMDE_X86_SSE2`.  That's particulary important for MSVC, since for a lot of ISA extensions you can't actually tell the compiler to enable them, they're *only* available as explicit calls to the intrinsics in question.  SSE2 isn't a good example here (it's one MSVC does actually advertise in macros, by setting `_M_IX86_FP` >= 2), but for example there is no `__SSSE3__` in MSVC.

Compounding the confusion is that it's a recent change.  Until May 2 `SIMDE_ARCH_X86_SSE2` was usually the right macro to use *outside* of sse2.h (unless you were in sse3.h, for example, in which case there *may* have been a `SIMDE_SSE3_SSE2`), but *inside* of sse2.h it would have been `SIMDE_X86_SSE2_NATIVE`.

Take a quick look at simde-arch.h and simde-features.h, that should clear things up.  I don't know about you, but it's generally easier for me to just read the code than a description.  The new system is actually a lot simpler, which is why I made the changes.

> Have you reverted it?

Kind of, it depends what you mean.  I didn't exactly revert anything.  I went through your patch and picked out the substantive changes which were correct, then integrated them into master.  That was something I didn't merge.

> I also see a SIMDE_X166_AVX512BW_NATIVE which is obviously a result of s/8/16/ :-(

Yes, I'll fix that.

> Again if you used #if X instead of #if defined(X) this could be picked up on compilation.

Yes, but it would also create a lot of inconsistencies, and if you want to write code where the definition may or may not exist (like because you may be using a library where you can't control which specific version of the library is available at compile time) you have to do something like

```c
#if defined(FOO)
  #if FOO > X
    // use foo
  #endif
#endif
```

Since the preprocessor doesn't short-circuit `defined(FOO) && (FOO > X)` will cause a problem with -Wundef if FOO is not defined.  That makes chaining these conditions together difficult.  You trade one problem which is easy to fix for one that requires ugly worked arounds and you don't even fully solve the original problem since a lot of people don't bother with non-default warnings like -Wundef.

I get the argument for `#if FOO > X`, and if it weren't for the drawbacks I'd definitely switch.  However, when you factor in all the pros and cons of each approach I feel like defined/!defined is the ~better~ less bad option.

rosbif:
You don't seem to like me implementing intrinsics as macros, even when they are trivial.
(You changed several in avx512bw.h.)
Should I always use functions?

nemequ:
> You don't seem to like me implementing intrinsics as macros, even when they are trivial.
> (You changed several in avx512bw.h.)
> Should I always use functions?

Yes.  The only time to use a macro are the native implementations of intrinsics with immediate mode arguments which compilers actually require to be compile time constants (IIG isn't trustworthy here).

Even if the x86 implementations are trivial, others may not be.  Putting them in functions means future patches are less intrusive (and less destructive of the git blame), and the consistency is nice.  Macros aren't good for type safety, are generally more error prone, harder for tooling, etc.  Really the only down side I can think of is that they're a bit more verbose, but I'd much rather the code be easy to read than quick to write.

nemequ:
> I am not sure what I should do here
> I pushed these updated cmp intrinsics to my fork about 9 hours ago.
> From what you say I suppose you removed them before merging to master.
> In that case I suppose you must also have removed their tests.

Sorry, I missed this paragraph earlier.

Yes, I dropped the tests since they only passed because they had incorrect values.  For example, you have `#define simde_mm512_cmple_epu8_mask(a, b) simde_mm512_cmpge_epu8_mask(b, a)` for the non-native version.  The tests match what you expect because they're based on the `SIMDE_AVX512BW_NATIVE`, which will never be defined, so it's always hitting the fallback path and calling cmpge instead of cmple.

I'll go ahead and add the missing functions and tests so we can close this PR, and you'll have a clean place to work from.  That way you don't have to worry about rebasing or anything.  I should have it done within the next couple hours.

nemequ:
The missing cmp functions are in master now, I think that's all the changes from this PR.

Thanks for the patch, sorry the merge process was so painful.

rosbif:
> > I also see a SIMDE_X166_AVX512BW_NATIVE which is obviously a result of s/8/16/ :-(
> 
> Yes, I'll fix that.
> 
> _Edit_: fixed with [4f2582c](https://github.com/nemequ/simde/commit/4f2582ce5ab10a797cd5b8d57d8d9a1469b488b9). I also added a check to CI to make sure it doesn't happen again.
> 
> > Again if you used #if X instead of #if defined(X) this could be picked up on compilation.
> 
> Yes, but it would also create a lot of inconsistencies, and if you want to write code where the definition may or may not exist (like because you may be using a library where you can't control which specific version of the library is available at compile time) you have to do something like
> 
> ```c
> #if defined(FOO)
>   #if FOO > X
>     // use foo
>   #endif
> #endif
> ```
> 
> Since the preprocessor doesn't short-circuit `defined(FOO) && (FOO > X)` will cause a problem with -Wundef if FOO is not defined. That makes chaining these conditions together difficult. You trade one problem which is easy to fix for one that requires ugly worked arounds and you don't even fully solve the original problem since a lot of people don't bother with non-default warnings like -Wundef.
> 
> I get the argument for `#if FOO > X`, and if it weren't for the drawbacks I'd definitely switch. However, when you factor in all the pros and cons of each approach I feel like defined/!defined is the ~better~ less bad option.

I didn't realize that `defined(FOO) && (FOO > X)` would throw a warning with `-Wundef ` if FOO is undefined. You very frequently see this construction in header files.
However I was considering  "binary" macros which would be either defined to 0 or to 1, like those we were talking about here.
These are purely for internal SIMDe  use and converting to the other style would allow developers who care to use `-Wundef`.
I realize that the two are incompatible and that, during a transition period, we would have to introduce a second macro with a slightly modified name.
Anyway I think we must agree to disagree on this one.
(I am old with failing memory and eyesight so I would appreciate anything the compiler can do to pick up my silly mistakes. This is also why I also suggested previously that run-tests should **compile-test** other code paths.)

rosbif:
> Yes, I dropped the tests since they only passed because they had incorrect values. For example, you have `#define simde_mm512_cmple_epu8_mask(a, b) simde_mm512_cmpge_epu8_mask(b, a)` for the non-native version. The tests match what you expect because they're based on the `SIMDE_AVX512BW_NATIVE`, which will never be defined, so it's always hitting the fallback path and calling cmpge instead of cmple.
> 

I don't understand you at all here :-(
What values were incorrect?
E.g. cmple calling cmpge with reversed arguments was deliberate.
(a <= b is equivalent to b >= a.)
It seemed more economical to introduce a macro to just call an existing function with reversed arguments than to write a new function.
But as you don't like me implementing intrinsics in macros the point is moot.

nemequ:
> I didn't realize that `defined(FOO) && (FOO > X)` would throw a warning with `-Wundef ` if FOO is undefined. You very frequently see this construction in header files.

It might not be -Wundef, but I know there is an off-by-default warning that triggers it.  It's kind of annoying because IIRC the standard is pretty clear about undefined macros evaluating (I don't think that's right word, but you get the idea) to 0.  Still, I really like for the public headers to compile cleanly even for ridiculous warnings since I don't want to teach people to disable or ignore compiler warnings.

> However I was considering "binary" macros which would be either defined to 0 or to 1, like those we were talking about here.

Yes, but you still have to compare them to something.  Instead of `> X` maybe it's `== 1` or just an implicit comparison to `!= 0` (i.e., just `#if FOO`), and you still have the same problem.  If you can't be sure the macro is defined you have to first test that it is defined, then check the value.

> These are purely for internal SIMDe use and converting to the other style would allow developers who care to use -Wundef.

Yeah, in the case of macros defined in the same project as we want to use them it's really about consistency.  I don't like the idea of having to do things differently outside of SIMDe.

That said, what I would be willing to do is define all these macros to 1 instead of nothing.  That would let people who disable -Wundef just do `#if FOO`.  But I do still want to allow `defined(FOO)` to work, so I don't want to `#define FOO 0` if for false.

To be honest, it's possible these are just all rationalizations and all I really care about is consistency.  `defined(x)` has been what I've always used, and it's been what I encounter most in the wild.

> Anyway I think we must agree to disagree on this one.

Yeah, I think so.  Still, it's nice to discuss this stuff occasionally.  Complacency is bad, and having my preferred conventions challenged forces me to actually think about them :)

> (I am old with failing memory and eyesight so I would appreciate anything the compiler can do to pick up my silly mistakes.

Maybe I'm getting old, too.  I'm usually a big fan of getting the compiler to pick up more bugs for me (hence wanting to compile cleanly even with dumb compiler warnings).  C is a hard language to get right, we need all the help we can get from compilers and static analyzers…

> This is also why I also suggested previously that run-tests should compile-test other code paths.)
>  Edit: Even better would be for run-tests to take a flag giving the choice between:

I don't remember what you're talking about for run-tests, do you have a link for that?  I like to think our CI is pretty aggressive, but getting more out of local tests wouldn't be bad.  I'm not really sure how you esvision this working, but maybe you could open up a new issue to discuss it?  It's definitely the sort of thing I'd be interested in.

> It seemed more economical to introduce a macro to just call an existing function with reversed arguments than to write a new function.

I don't think I noticed that the arguments were reversed.  I remember I looked for negation, but completely forgot about reversing the arguments, sorry :(

> It seemed more economical to introduce a macro to just call an existing function with reversed arguments than to write a new function.
> But as you don't like me implementing intrinsics in macros the point is moot.

Not entirely.  We could still do the same thing inside functions, which would make it so ports to other architectures would really only *need* to implement a single function.  If negation is required there would be a small performance hit to return `~r` instead of `r`, so maybe a couple functions if you want to avoid that.  I don't think it would have a big input since these are the type of things compilers should really be able to autovectorize, but I don't really see a down-side.

rosbif:
> > I didn't realize that `defined(FOO) && (FOO > X)` would throw a warning with `-Wundef ` if FOO is undefined. You very frequently see this construction in header files.
> 
> It might not be -Wundef, but I know there is an off-by-default warning that triggers it. It's kind of annoying because IIRC the standard is pretty clear about undefined macros evaluating (I don't think that's right word, but you get the idea) to 0. Still, I really like for the public headers to compile cleanly even for ridiculous warnings since I don't want to teach people to disable or ignore compiler warnings.
> 

Warnings are just for that: signalling things that are permitted but which could indicate a mistake or an oversight. I had a colleague who always wrote tests against a constant backwards. I suppose it was to throw an error if she wrote e.g. `if (9 = foo)` instead of  `if (9 == foo)`. For equality it is not too bad but I found readability suffered for other tests. (But maybe it is just me.)

> > This is also why I also suggested previously that run-tests should compile-test other code paths.)
> > Edit: Even better would be for run-tests to take a flag giving the choice between:
> 
> I don't remember what you're talking about for run-tests, do you have a link for that? I like to think our CI is pretty aggressive, but getting more out of local tests wouldn't be bad. I'm not really sure how you esvision this working, but maybe you could open up a new issue to discuss it? It's definitely the sort of thing I'd be interested in.
> 

I suggested it here: https://github.com/nemequ/simde/issues/169#issuecomment-620582863.
Yes the CI is excellent, congratulations.
However you have to add, commit, push and PR to pick up every silly error in a code path that you didn't test locally. Plus it generates a lot of noise (and a lot of shame) when everyone sees your stupid mistakes :-( However this last point is somewhat mitigated by being able to use Travis-CI without a PR.

rosbif:
I put an RFC version of avx512bw.h with my 0 or 1 macro ideas in the define branch of my fork.
It passes all Travis-CI tests.

Obviously the stuff at the beginning which just converts into 0 or 1 macros will disappear if this solution is adopted.
My 0 or 1 macros just have a trailing underscore added.

The advantages of my approach:
- with -Wundef almost all misspelt macros will be detected,
- almost all code paths are compile-tested with any compilation.

The proof: two latent pre-existant MSVC warning/errors (bad casts) were picked up.

rosbif:
I think I have finished this round if the tests pan out OK.
(xlc failed but I don't think it is my fault. I restarted it and all Travis-CI is green now.)
I shall no longer touch this branch.
But it is getting late so if there is a problem it will have to wait till tomorrow.
Sorry it has taken so long.

For the future I have improvements to gfni.h which is nearly finished and NEON stuff which is embryonic.

@nemequ: Would you please give feedback (even negative ;-)) on what I did here https://github.com/nemequ/simde/pull/265#issuecomment-626328121 ?
See also the previous messages.
TIA.

nemequ:
Sorry, I thought this issue was finished.  Are there any commits in here that haven't been merged but you feel are ready to be?  It gets a bit confusing for issues like this where the PR continues to evolve after it has been merged, especially when it hasn't been rebased, but if there are still changes in here I'd definitely like to get them merged ASAP so they don't bit rot any more than they already have.

Reading comments since my last comment now…

> However you have to add, commit, push and PR to pick up every silly error in a code path that you didn't test locally. Plus it generates a lot of noise (and a lot of shame) when everyone sees your stupid mistakes :-( However this last point is somewhat mitigated by being able to use Travis-CI without a PR.

I'd like to have the ability to test more stuff locally too.  For me the biggest issue with CI is the waiting.  IIRC it usually takes more than 30 minutes for CI to test a commit because we have so many things to check, and unfortunately that's only going to get worse as we add more instructions.  Even spread across 5 different providers (Travis CI, GitHub Actions, Drone.io, Azure Pipelines, and Appveyor) it's excruciating.

The noise I can definitely relate to; that's why I try to use a different repo which nobody should be watching for commits where I'm really just trying something to see how it goes.

As for the shame, that goes away pretty quickly once you realize *everyone* is going through the same stuff, especially for highly portable projects with aggressive diagnostics treated as errors like SIMDe.

I just didn't know what your were proposing.  Switching over to replying to the linked comment:

> While I appreciate that to actually run all tests on all code paths would be very time consuming, couldn't run-tests at least compile-test all code paths (different pure C/MMX/SSE/AVX/AVX512/GFNI levels)?

I wouldn't want to do that by default, but it would be possible to add an option to Meson for something like that.  You could detect the CPU features in the test harness to figure out which tests to run, so you don't need to limit yourself to compile-only.  It would be a pretty big undertaking, though.  

An easier way would be to just add a shell script to the test directory to automate the process of building each version.  At its core, just something like (untested)

```bash
COMPILER_FLAGS="-Wall -Wextra -Werror"
BUILDDIR="$(mktemp -d)"

for ISAX in \
    "-mno-mmx -mno-sse -mno-sse2" \
    "-mmmx -mno-sse -mno-sse2" \
    "-mmmx -msse -mno-sse2" \
    "-mmmx -msse -msse2" \
    "-mno-mmx -msse -msse2" \
    "-msse3" "-mssse3" "-msse4.1" "-msse4.2" "-mavx" "-mavx2" "-mfma" \
    "-mavx512f" \
    "-mavx512f -mavx512vl" \
    "-mavx512f -mavx512bw" \
    "-mavx512f -mavx512vl -mavx512bw" \
    "-mavx512f -mavx512dq" \
    "-mavx512f -mavx512vl -mavx512bw -mavx512dq"; do
  if [ ! -e build-lots ]; then
    mkdir build-lots
  else
   rm -rf build-lots/*
  fi
  CFLAGS='"${COMPILER_FLAGS} ${ISAX}" CXXFLAGS="${COMPILER_FLAGS} ${ISAX}" meson "${BUILDDIR}" || break
  ninja -C "${BUILDDIR}"
done
```

The AVX-512F stuff would need some tuning to figure out which combinations of flags to test, but that should give you the general idea.  I'd be happy to add something like that.  We could even add code to try different compilers.

Of course, all this assumes you're on x86.  On other architectures there is a different (though less extensive) set of flags to try for each.

Obviously this would only supplement CI, not replace it, but I think that's all you're talking about anyways.

> Also as you are really serious about big-endian support it would be useful to have portable macros or functions for 16-bit, 32-bit and 64-bit byte swaps.

https://github.com/nemequ/simde/blob/f9e183b37de00965e007e6773488a4eea4678a8e/simde/simde-common.h#L342-L417

I think I added this after I saw the __builtin_bswap64 in your code along with a note to fix it (so I fixed it).  I didn't bother with other lengths, I figured we could add them as necessary.  I just haven't needed endian-swapping code for SIMDe, but if you do I can add it.  Maybe even update it a bit to merge it into a single header.

> Edit: I haven't thought through this but maybe with real tests against a const (or a macro evaluating to a constant) rather than preprocessor tests, the compiler would check all code paths but optimize away unreachable ones.

Even if you gave all the macros values (instead of just checking if they're defined), it would fail if, for example, you call an invalid function, like a NEON function on x86, or a __builtin_* that the compiler doesn't support.  Or if there was a syntax error because the compile doesn't support an extension, like statement exprs or vector extensions (or even parts of vector extensions which it doesn't support).  Or cases where we're working around compiler bugs.

There is a fairly broad set of circumstances where this would work, like falling back on previous ISA extensions for the same architecture, but there would also be tons of exceptions and no consistency.  We would still need to use preprocessor logic in a lot of places, and having to constantly switch back and forth between the different styles would make working on the code extremely unpleasant.

> I put an RFC version of avx512bw.h with my 0 or 1 macro ideas in the define branch of my fork.
It passes all Travis-CI tests.

I get the idea, and it does have advantages.  However, seeing the code convinces me that I *really* don't like it, largely for the reasons I mentioned above.

> The proof: two latent pre-existant MSVC warning/errors (bad casts) were picked up.

Nice, do you know where those were?  We should fix them.

rosbif:
> Sorry, I thought this issue was finished. Are there any commits in here that haven't been merged but you feel are ready to be? It gets a bit confusing for issues like this where the PR continues to evolve after it has been merged, especially when it hasn't been rebased, but if there are still changes in here I'd definitely like to get them merged ASAP so they don't bit rot any more than they already have.
> 

In https://github.com/nemequ/simde/issues/261#issuecomment-623590615 you said that dialogue could continue after things are closed. I continued here because it is vaguely related: with the current scheme additional fallbacks create code paths that are removed by the pre-processor and that the compiler cannot even compile-test. That is why I _really_  don't like your `#if defined(FOO`) approach ;-)
But apparently we violently disagree on this.
Apart from this the only other commits would be for the two MSVC cast errors but I'll generate a PR for these based on the current master if they are still present.

> The noise I can definitely relate to; that's why I try to use a different repo which nobody should be watching for commits where I'm really just trying something to see how it goes.
> 

Does a fork count as a different repo?
It is not clear to me who sees what messages. For instance if I add a specific @nemequ does the message just go to you?

> As for the shame, that goes away pretty quickly once you realize _everyone_ is going through the same stuff, especially for highly portable projects with aggressive diagnostics treated as errors like SIMDe.
> 

Yes but unfortunately there are also (in my case) many typos in completely untested code paths removed by the pre-processor with your #if approach.

> I just didn't know what your were proposing. Switching over to replying to the linked comment:
> 
> > While I appreciate that to actually run all tests on all code paths would be very time consuming, couldn't run-tests at least compile-test all code paths (different pure C/MMX/SSE/AVX/AVX512/GFNI levels)?
> 
> I wouldn't want to do that by default, but it would be possible to add an option to Meson for something like that. You could detect the CPU features in the test harness to figure out which tests to run, so you don't need to limit yourself to compile-only. It would be a pretty big undertaking, though.
> 
> An easier way would be to just add a shell script to the test directory to automate the process of building each version. At its core, just something like (untested)
> 
> ```shell
> COMPILER_FLAGS="-Wall -Wextra -Werror"
> BUILDDIR="$(mktemp -d)"
> 
> for ISAX in \
>     "-mno-mmx -mno-sse -mno-sse2" \
>     "-mmmx -mno-sse -mno-sse2" \
>     "-mmmx -msse -mno-sse2" \
>     "-mmmx -msse -msse2" \
>     "-mno-mmx -msse -msse2" \
>     "-msse3" "-mssse3" "-msse4.1" "-msse4.2" "-mavx" "-mavx2" "-mfma" \
>     "-mavx512f" \
>     "-mavx512f -mavx512vl" \
>     "-mavx512f -mavx512bw" \
>     "-mavx512f -mavx512vl -mavx512bw" \
>     "-mavx512f -mavx512dq" \
>     "-mavx512f -mavx512vl -mavx512bw -mavx512dq"; do
>   if [ ! -e build-lots ]; then
>     mkdir build-lots
>   else
>    rm -rf build-lots/*
>   fi
>   CFLAGS='"${COMPILER_FLAGS} ${ISAX}" CXXFLAGS="${COMPILER_FLAGS} ${ISAX}" meson "${BUILDDIR}" || break
>   ninja -C "${BUILDDIR}"
> done
> ```
> 
> The AVX-512F stuff would need some tuning to figure out which combinations of flags to test, but that should give you the general idea. I'd be happy to add something like that. We could even add code to try different compilers.
> 
> Of course, all this assumes you're on x86. On other architectures there is a different (though less extensive) set of flags to try for each.
> 
> Obviously this would only supplement CI, not replace it, but I think that's all you're talking about anyways.
> 

But with my approach all these complications are removed as (almost) all code paths are compile-tested because the compiler sees and reports errors in code in an `if (0) `but not in a `#if 0`.

> > Also as you are really serious about big-endian support it would be useful to have portable macros or functions for 16-bit, 32-bit and 64-bit byte swaps.
> 
> https://github.com/nemequ/simde/blob/f9e183b37de00965e007e6773488a4eea4678a8e/simde/simde-common.h#L342-L417
> 
> I think I added this after I saw the __builtin_bswap64 in your code along with a note to fix it (so I fixed it). I didn't bother with other lengths, I figured we could add them as necessary. I just haven't needed endian-swapping code for SIMDe, but if you do I can add it. Maybe even update it a bit to merge it into a single header.
> 

Thank you. The code was a temporary quick and dirty (but successful) attempt to fix the endian problem. I cannot do big-endian tests locally because Ubuntu screws up the four or five big-endian cross-compilations and cross-executions that I have tried.
(I confirmed that this works in a Debian Sid chroot but it is inconvenient.)

> > Edit: I haven't thought through this but maybe with real tests against a const (or a macro evaluating to a constant) rather than preprocessor tests, the compiler would check all code paths but optimize away unreachable ones.
> 
> Even if you gave all the macros values (instead of just checking if they're defined), it would fail if, for example, you call an invalid function, like a NEON function on x86, or a __builtin_* that the compiler doesn't support. Or if there was a syntax error because the compile doesn't support an extension, like statement exprs or vector extensions (or even parts of vector extensions which it doesn't support). Or cases where we're working around compiler bugs.
> 

Yes you will see in my code that I had to leave a _few_  #if for (mainly vector) constructs that aren't  standard C. I also moved the native replacement outside the functions because the native intrinsic would be undefined in compilations for "lower levels". I am not sure that it would fail on a NEON function if it were simde_ prefixed. As for builtins it is best to have portable versions elsewhere like you did for the byte swap.

> There is a fairly broad set of circumstances where this would work, like falling back on previous ISA extensions for the same architecture, but there would also be tons of exceptions and no consistency. We would still need to use preprocessor logic in a lot of places, and having to constantly switch back and forth between the different styles would make working on the code extremely unpleasant.
> 

Yes, as I said above, there are _some_ exceptions but IMHO _not  tons_.

> > I put an RFC version of avx512bw.h with my 0 or 1 macro ideas in the define branch of my fork.
> > It passes all Travis-CI tests.
> 
> I get the idea, and it does have advantages. However, seeing the code convinces me that I _really_ don't like it, largely for the reasons I mentioned above.
> 

And I _really_ do like it, but it's your call :-(
I didn't mention, but it is obvious, that the two approaches can co-exist.

> > The proof: two latent pre-existant MSVC warning/errors (bad casts) were picked up.
> 
> Nice, do you know where those were? We should fix them.
>

Yes, as I said above I'll generate a PR for current master to fix them if they are still present.

nemequ:
> > Sorry, I thought this issue was finished. Are there any commits in here that haven't been merged but you feel are ready to be? It gets a bit confusing for issues like this where the PR continues to evolve after it has been merged, especially when it hasn't been rebased, but if there are still changes in here I'd definitely like to get them merged ASAP so they don't bit rot any more than they already have.
> 
> In [#261 (comment)](https://github.com/nemequ/simde/issues/261#issuecomment-623590615) you said that dialogue could continue after things are closed. I continued here because it is vaguely related:

Yes, I don't have a problem continuing here, it just kind of got buried in my notifications.

> with the current scheme additional fallbacks create code paths that are removed by the pre-processor and that the compiler cannot even compile-test. That is why I _really_ don't like your `#if defined(FOO`) approach ;-)

Yes, that's definitely a big advantage of your suggestion.  I just don't think it's worth all the drawbacks, especially when we can compile-test much more thoroughly using other means.

> But apparently we violently disagree on this.

Yeah, I think so.  Unfortunatley I don't think there is a right answer; it's just two conflicting opinions :(

> Apart from this the only other commits would be for the two MSVC cast errors but I'll generate a PR for these based on the current master if they are still present.

Excellent, thanks!

> > The noise I can definitely relate to; that's why I try to use a different repo which nobody should be watching for commits where I'm really just trying something to see how it goes.
> 
> Does a fork count as a different repo?

Yes.  I don't receive any notifications for your rosbif/simde repo unless I "watch" the repo, which I don't.

> It is not clear to me who sees what messages. For instance if I add a specific @nemequ does the message just go to you?

Yes, if you @ me in an issue (or a commit, etc.), I'll get a message, but AFAIK nobody else will.

There *might* be an exception for people who watch me as a user instead of watching SIMDe.  I'm not sure since I've never actually watched a user, but think it just means you see their activity not their mentions.  That said, if someone watches you they're (literally) asking for it, so *shrug*.

> > As for the shame, that goes away pretty quickly once you realize _everyone_ is going through the same stuff, especially for highly portable projects with aggressive diagnostics treated as errors like SIMDe.
> 
> Yes but unfortunately there are also (in my case) many typos in completely untested code paths removed by the pre-processor with your #if approach.

That's okay, it happens to everyone.  Unfortunately that's never going to go away, unless you want to install all the compilers on local hardware.  Even with your approach where will still be a lot of untested code; not all compilers can handle the same supersets of C.  I guess you could limit yourself to a strict lowest-common-denominator, but then you're leaving a lot of really nice features on the table.

> > I just didn't know what your were proposing. Switching over to replying to the linked comment:
> > > While I appreciate that to actually run all tests on all code paths would be very time consuming, couldn't run-tests at least compile-test all code paths (different pure C/MMX/SSE/AVX/AVX512/GFNI levels)?
> > 
> > 
> > I wouldn't want to do that by default, but it would be possible to add an option to Meson for something like that. You could detect the CPU features in the test harness to figure out which tests to run, so you don't need to limit yourself to compile-only. It would be a pretty big undertaking, though.
> > An easier way would be to just add a shell script to the test directory to automate the process of building each version. At its core, just something like (untested)
> > ```shell
> > COMPILER_FLAGS="-Wall -Wextra -Werror"
> > BUILDDIR="$(mktemp -d)"
> > 
> > for ISAX in \
> >     "-mno-mmx -mno-sse -mno-sse2" \
> >     "-mmmx -mno-sse -mno-sse2" \
> >     "-mmmx -msse -mno-sse2" \
> >     "-mmmx -msse -msse2" \
> >     "-mno-mmx -msse -msse2" \
> >     "-msse3" "-mssse3" "-msse4.1" "-msse4.2" "-mavx" "-mavx2" "-mfma" \
> >     "-mavx512f" \
> >     "-mavx512f -mavx512vl" \
> >     "-mavx512f -mavx512bw" \
> >     "-mavx512f -mavx512vl -mavx512bw" \
> >     "-mavx512f -mavx512dq" \
> >     "-mavx512f -mavx512vl -mavx512bw -mavx512dq"; do
> >   if [ ! -e build-lots ]; then
> >     mkdir build-lots
> >   else
> >    rm -rf build-lots/*
> >   fi
> >   CFLAGS='"${COMPILER_FLAGS} ${ISAX}" CXXFLAGS="${COMPILER_FLAGS} ${ISAX}" meson "${BUILDDIR}" || break
> >   ninja -C "${BUILDDIR}"
> > done
> > ```
> > 
> > 
> > The AVX-512F stuff would need some tuning to figure out which combinations of flags to test, but that should give you the general idea. I'd be happy to add something like that. We could even add code to try different compilers.
> > Of course, all this assumes you're on x86. On other architectures there is a different (though less extensive) set of flags to try for each.
> > Obviously this would only supplement CI, not replace it, but I think that's all you're talking about anyways.
> 
> But with my approach all these complications are removed as (almost) all code paths are compile-tested because the compiler sees and reports errors in code in an `if (0) `but not in a `#if 0`.

I think you overestimate the number of paths that you could test like this that aren't also covered by the existing tests (not talking about CI, just the native and emul tests for each function).  It's basically just fallbacks using previous ISA extensions.  No GCC-style vector extensions, no SIMDE_SHUFFLE/CONVERT_VECTOR (__builtin_shuffle/__builtin_shufflevector/__builtin_convertvector). No other architectures.

> > > Also as you are really serious about big-endian support it would be useful to have portable macros or functions for 16-bit, 32-bit and 64-bit byte swaps.
> > 
> > 
> > https://github.com/nemequ/simde/blob/f9e183b37de00965e007e6773488a4eea4678a8e/simde/simde-common.h#L342-L417
> > 
> > I think I added this after I saw the __builtin_bswap64 in your code along with a note to fix it (so I fixed it). I didn't bother with other lengths, I figured we could add them as necessary. I just haven't needed endian-swapping code for SIMDe, but if you do I can add it. Maybe even update it a bit to merge it into a single header.
> 
> Thank you. The code was a temporary quick and dirty (but successful) attempt to fix the endian problem. I cannot do big-endian tests locally because Ubuntu screws up the four or five big-endian cross-compilations and cross-executions that I have tried.
> (I confirmed that this works in a Debian Sid chroot but it is inconvenient.)
> 
> > > Edit: I haven't thought through this but maybe with real tests against a const (or a macro evaluating to a constant) rather than preprocessor tests, the compiler would check all code paths but optimize away unreachable ones.

You can do it with a union:

https://github.com/nemequ/portable-snippets/blob/master/endian/endian.h#L152-L181

But you're not supposed to use #if/#else in your code.  That's why simde_endian_bswap64_be/le are defined the way they are:

https://github.com/nemequ/simde/blob/master/simde/simde-common.h#L422-L428

You just call `simde_endian_bswap64_le(value)` and it will preprocess to a swap if you're on little-endian and just `(value)` if you're on big-endian.  `simde_endian_bswap64_be(value)` does the reverse; it's a swap if you're on big-endian and the value if you're on little endian.

> > Even if you gave all the macros values (instead of just checking if they're defined), it would fail if, for example, you call an invalid function, like a NEON function on x86, or a __builtin_* that the compiler doesn't support. Or if there was a syntax error because the compile doesn't support an extension, like statement exprs or vector extensions (or even parts of vector extensions which it doesn't support). Or cases where we're working around compiler bugs.
> 
> Yes you will see in my code that I had to leave a _few_ #if for (mainly vector) constructs that aren't standard C.

The number of those are going to increase soon.  You've been working on adding x86 fallbacks, and I intend to go through and start adding vector extensions, other architectures, and OpenMP SIMD versions (which would work with your proposal).  What you're seeing is a heavily biased view; take a look at sse.h if you want to get a better idea of how the code should look in the future.  I intend to start working on that soon; that's what #275 is about.

> I also moved the native replacement outside the functions because the native intrinsic would be undefined in compilations for "lower levels". I am not sure that it would fail on a NEON function if it were simde_ prefixed.

That would work, but it would require us to implement every ISA extension simultaneously.  That's asking a lot.

> As for builtins it is best to have portable versions elsewhere like you did for the byte swap.

You could use something like [the builtin module in portable-snippets](https://github.com/nemequ/portable-snippets/tree/master/builtin) (or possibly a subset) for some functions, but there are a lot of builtins which can't be emulated, and frankly those tend to be the ones which are useful in SIMDe.  I haven't had to use `__builtin_ffs` or `__builtin_ctz` yet, but `__builtin_shuffle`, `__builtin_shufflevector`, and `__builtin_convertvector` are all over the place.

> > There is a fairly broad set of circumstances where this would work, like falling back on previous ISA extensions for the same architecture, but there would also be tons of exceptions and no consistency. We would still need to use preprocessor logic in a lot of places, and having to constantly switch back and forth between the different styles would make working on the code extremely unpleasant.
> 
> Yes, as I said above, there are _some_ exceptions but IMHO _not tons_.

The classes of exceptions aren't that great, but each will be fairly common.

Anyways, I think you're right that we're going to have to agree to disagree here :(

PR #266, opened 2020-05-05
drone: remove failure ignore on Fedora clang.

junaruga:
I noticed Drone CI "fedora clang arm64 flags" build succeeds now by the fix https://github.com/nemequ/simde/issues/167 .
So, let's remove `"failure": "ignore"` on the build.


nemequ:
Nice, thanks :)

PR #274, opened 2020-05-08
avx512vl.h: Add SSE2 fallback to simde_mm_mask_mov_epi8.

rosbif:
I think this is likely to be heavily used (indirectly).
So add an SSE2 fallback for systems without SSE4.1.


rosbif:
I tried to rebase onto the current master.
But this seems to have doubled the commit :-(

nemequ:
Looks good, thanks!

> I tried to rebase onto the current master.
> But this seems to have doubled the commit :-(

It looks like you did a merge not a rebase.  No worries, a quick rebase is still al it needed.  It's going through CI, I'll merge it into master as soon as it's done.

FWIW, all I had to do was `git rebase origin/master`.  I actually edited the commit message a bit too to add the issue number, but that's not necessary; I only bother with that if I'm already rebasing or editing the commit… it's just nice to be able to quickly find the relevant PR in case there is additional discussion, but for this PR it's probably a waste since no discussion is necessary.

PR #275, opened 2020-05-08
Improve how we decide on fallbacks

nemequ:
I've been thinking about how to improve how we decide whether or not to fall back on another function for a while, and this is what I've settled on.

@rosbif, what do you think of this?  You've been doing a lot of work on the fallbacks, so this probably effects you the most.

It's just a WIP, I haven't done the fallbacks to 256-bit functions yet.  Just wanted to get the idea out there and get some feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,12 +41,17 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install APT Dependencies
-      run: sudo add-apt-repository 'ppa:ubuntu-toolchain-r/test' && sudo apt-get update && sudo apt-get install -y ninja-build python3-lxml
+      run: sudo add-apt-repository 'ppa:ubuntu-toolchain-r/test' && sudo apt-get update && sudo apt-get install -y ninja-build
     - name: Download SDE
-      run: cd test && python3 download-sde.py sde.tar.bz2 && tar jxvf sde.tar.bz2
+      run: |
+        mkdir sde && \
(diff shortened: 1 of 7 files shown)


rosbif:
I think that, in general, each "level" should only fall back to the next lower "level" which can provide an efficient implementation.
However sometimes (I think fairly rarely) it is more efficient to provide a "level n-2" implementation in addition to a "level n-1" implementation.
Off the top of me head I can only think of this recent example:
For SSE4.1 simde_mm_mask_mov_epi8 falls back to movm + blendv
For SSE2 it falls back to movm + and + andnot + or
In SSE4.1 I think that blendv should fall back to SSE2 setzero + cmpgt + and + andnot + or (but it doesn't).
Falling back directly to SSE2 would avoid an unnecessary setzero + cmpgt.
Not a big deal you'll say but I couldn't think of a better example.

rosbif:
Sorry I hadn't realized that you wanted a review of a specific proposal rather than just a discussion.
I don't think that your suggestion will permit deciding between two fallbacks with the same vector size, like my example above.
There is also maybe the case where a bitwise identical operation is defined at two different levels for different data types.
For instance I have never understood why
_mm256_xor_ps and _mm256_xor_pd are in AVX
but
_mm256_xor_si256 requires AVX2
Aren't they bitwise the same operation?
(Worse than this I can't see a use-case for xor on floats or doubles.
Possibly and/or if you are manipulating mantissas and exponents. But xor ???)

PR #284, opened 2020-05-12
sse4.2: added implementation for mm_cmpestrs

masterchef2209:
sse4.2: added implementation for mm_cmpestrs

I noticed that this depends on only imm8[0], so this must be simpler to do first.


test/x86/sse4.2.c
```diff
@@ -24,9 +24,231 @@
 #define SIMDE_TESTS_CURRENT_ISAX sse4_2
 #include <test/x86/test-x86-internal.h>
 #include <simde/x86/sse4.2.h>
+#include <assert.h>
```
nemequ, line 27: Please don't use `assert.h`.

test/x86/sse4.2.c
```diff
@@ -24,9 +24,231 @@
...
+#define SIMDE_SIDD_CMP_EQUAL_ORDERED 0x0c
+#define SIMDE_SIDD_POSITIVE_POLARITY 0x00
+#define SIMDE_SIDD_NEGATIVE_POLARITY 0x10
+#define SIMDE_SIDD_MASKED_POSITIVE_POLARITY 0x20
+#define SIMDE_SIDD_MASKED_NEGATIVE_POLARITY 0x30
+#define SIMDE_SIDD_LEAST_SIGNIFICANT 0x00
+#define SIMDE_SIDD_MOST_SIGNIFICANT 0x40
+#define SIMDE_SIDD_BIT_MASK 0x00
+#define SIMDE_SIDD_UNIT_MASK 0x40
+#endif
```
nemequ, line 66: These generally look good, but  they should be in sse4.2.h not the tests.  They're part of the public SSE4.2 API.

They also still need native aliases, otherwise it will fail if the target doesn't support SSE4.2 but `SIMDE_ENABLE_NATIVE_ALIASES` is defined.  You just need to add another set like

```c
#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
  #define _SIDD_UBYTE_OPS SIMDE_SIDD_UBYTE_OPS
  // ...
#endif
```

test/x86/sse4.2.c
```diff
@@ -24,9 +24,231 @@
...
 #include <simde/x86/sse4.2.h>
+#include <assert.h>
 
 #if defined(SIMDE_X86_SSE4_2_NATIVE) || defined(SIMDE_NO_NATIVE) || defined(SIMDE_ALWAYS_BUILD_NATIVE_TESTS)
 
+#if defined(SIMDE_X86_SSE4_2_NATIVE)
+#define SIMDE_SIDD_UBYTE_OPS _SIDD_UBYTE_OPS
+#define SIMDE_SIDD_UWORD_OPS _SIDD_UWORD_OPS
+#define SIMDE_SIDD_SBYTE_OPS _SIDD_SBYTE_OPS
+#define SIMDE__SIDD_SWORD_OPS _SIDD_SWORD_OPS
```
nemequ, line 35: Stray underscore; this should be `SIMDE_SIDD_SWORD_OPS` not `SIMDE__SIDD_SWORD_OPS`

test/x86/sse4.2.c
```diff
@@ -24,9 +24,231 @@
 #define SIMDE_TESTS_CURRENT_ISAX sse4_2
 #include <test/x86/test-x86-internal.h>
 #include <simde/x86/sse4.2.h>
+#include <assert.h>
 
 #if defined(SIMDE_X86_SSE4_2_NATIVE) || defined(SIMDE_NO_NATIVE) || defined(SIMDE_ALWAYS_BUILD_NATIVE_TESTS)
 
+#if defined(SIMDE_X86_SSE4_2_NATIVE)
+#define SIMDE_SIDD_UBYTE_OPS _SIDD_UBYTE_OPS
```
nemequ, line 32: Small style issue.  The inside of the #if block should be indented; i.e., 

```c
#if defined(SIMDE_X86_SSE4_2_NATIVE)
  #define SIMDE_SIDD_UBYTE_OPS _SIDD_UBYTE_OPS
```

This style is a little weird, most projects would prefer something like

```c
#if defined(SIMDE_X86_SSE4_2_NATIVE)
#  define SIMDE_SIDD_UBYTE_OPS _SIDD_UBYTE_OPS
```

But for SIMDe I'm trying to use the former since it's a bit easier to handle consistently when automatically generating code.  Basically, preprocessor conditionals are indented just like regular conditionals; they start at the same indentation as normal code, and all code inside of them (regardless of whether we're talking about normal code or preprocessor code) gets indented.

simde/x86/sse4.2.h
```diff
@@ -37,6 +37,32 @@ SIMDE__BEGIN_DECLS
...
+
+#if defined(SIMDE_X86_SSE4_2_NATIVE)
+#  define simde_mm_cmpestrs(a, la, b, lb, imm8) _mm_cmpestrs(a, la, b, lb, imm8)
+#else
+#  define simde_mm_cmpestrs(a, la, b, lb, imm8) \
+     (((imm8) & SIMDE_SIDD_UWORD_OPS) \
+       ? simde_mm_cmpestrs_16_(la) \
+       : simde_mm_cmpestrs_8_(la))
+#endif
+#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
```
nemequ, line 62: You're right, this is simpler.  So much so, in fact, that I don't think it makes sense to break this into separate functions.  You should only break up code like this into separate functions when it makes things easier to understand, but this just makes things more complicated.  This could be implemented simply as `return la < ((128 / ((imm8 & SIMDE_SIDD_UWORD_OPS) ? 16 : 8)) - 1);`, plus of course the native implementation.  And we can put a `SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 0x7f)` on it to validate the input.

test/x86/sse4.2.c
```diff
@@ -24,9 +24,231 @@
...
+                        INT8_C(  30), INT8_C( 110), INT8_C( -36), INT8_C(  70),
+                        INT8_C(-126), INT8_C( -29), INT8_C( -65), INT8_C( -41),
+                        INT8_C( -71), INT8_C(   1), INT8_C( 121), INT8_C(-119)),
+      10 ,
+      1 }
+  };
+
+  for (size_t i = 0 ; i < (sizeof(test_vec) / sizeof(test_vec[0])); i++) {
+    int r = simde_mm_cmpestrs(test_vec[i].a, test_vec[i].la, test_vec[i].b, test_vec[i].lb, 0);
+    assert(r == test_vec[i].r);
```
nemequ, line 171: `munit_assert_int(r, ==, test_vec[i].r);`

Same below.

masterchef2209:
I have pushed requested changes, but I don't understand why it fails on some systems, please look into it.
Thank You

simde/x86/sse4.2.h
```diff
@@ -37,6 +37,77 @@ SIMDE_BEGIN_DECLS_
...
+    _mm_cmpestrs(a, la, b, lb, imm8)
+#else
+  #define simde_mm_cmpestrs(a, la, b, lb, imm8) \
+    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 0x7f) \
+    la <= ((128 / ((imm8 & SIMDE_SIDD_UWORD_OPS) ? 16 : 8)) - 1);
+#endif
+#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
+  #define _mm_cmpestrs(a, la, b, lb, imm8) \
+    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 0x7f) \
+    simde_mm_cmpestrs(a, la, b, lb, imm8)
```
nemequ, line 108: SIMDE_REQUIRE_CONSTANT_RANGE generates a diagnose_if function attribute; you can't use it to decorate a macro.  

The `simde_mm_cmpestrs` implementation should be a function (with a `SIMDE_REQUIRE_CONSTANT_RANGE`).  Something like

```c
SIMDE_FUNCTION_ATTRIBUTES
int simde_mm_cmpestrs (__m128i a, int la, __m128i b, int lb, const int imm8)
    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 127) {
  return la <= ((128 / ((imm8 & SIMDE_SIDD_UWORD_OPS) ? 16 : 8)) - 1);
)
#if defined(SIMDE_X86_SSE4_2_NATIVE)
  #define simde_mm_cmpestrs(a, la, b, lb, imm8) _mm_cmpestrs(a, la, b, lb, imm8)
#endif
#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
  #define _mm_cmpestrs(a, la, b, lb, imm8) simde_mm_cmpestrs(a, la, b, lb, imm8)
#endif
```

If the native implementation is used instead of the SIMDe implementation the compiler will still emit an error if the imm8 value is outside the expected range, so we don't really need the SIMDE_REQUIRE_CONSTANT_RANGE on that case.  SIMDE_REQUIRE_CONSTANT_RANGE is to prevent the portable implementation from accepting a value which wouldn't work in the native API (thereby creating a poratability problem).

simde/x86/sse4.2.h
```diff
@@ -37,6 +37,77 @@ SIMDE_BEGIN_DECLS_
...
+  #define SIMDE_SIDD_CMP_EQUAL_ORDERED _SIDD_CMP_EQUAL_ORDERED
+  #define SIMDE_SIDD_POSITIVE_POLARITY _SIDD_POSITIVE_POLARITY
+  #define SIMDE_SIDD_NEGATIVE_POLARITY _SIDD_NEGATIVE_POLARITY
+  #define SIMDE_SIDD_MASKED_POSITIVE_POLARITY _SIDD_MASKED_POSITIVE_POLARITY
+  #define SIMDE_SIDD_MASKED_NEGATIVE_POLARITY _SIDD_MASKED_NEGATIVE_POLARITY
+  #define SIMDE_SIDD_LEAST_SIGNIFICANT _SIDD_LEAST_SIGNIFICANT
+  #define SIMDE_SIDD_MOST_SIGNIFICANT _SIDD_MOST_SIGNIFICANT
+  #define SIMDE_SIDD_BIT_MASK _SIDD_BIT_MASK
+  #define SIMDE_SIDD_UNIT_MASK _SIDD_UNIT_MASK
+
```
nemequ, line 57: This blank line should be removed.

simde/x86/sse4.2.h
masterchef2209: oh, i see

masterchef2209: what should be the right way of dealing with unused parameter problem which is resulting in some checks to fail?

nemequ: Cast them to void:

```c
SIMDE_FUNCTION_ATTRIBUTES
int simde_mm_cmpestrs (__m128i a, int la, __m128i b, int lb, const int imm8)
    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 127) {
  (void) a;
  (void) la;
  (void) b;
  (void) lb;
  return la <= ((128 / ((imm8 & SIMDE_SIDD_UWORD_OPS) ? 16 : 8)) - 1);
)
```

masterchef2209: I have corrected this.

PR #289, opened 2020-05-13
svml: adding macros for {mm,mm256}_{idiv,udiv}_{epi32,epu32}

himanshi18037:


simde/x86/svml.h
```diff
@@ -127,11 +127,13 @@ simde_mm_div_epi32 (simde__m128i a, simde__m128i b) {
     return simde__m128i_from_private(r_);
   #endif
 }
+#define simde_mm_idiv_epi32(a, b) simde_mm_div_epi32(a, b)
 #if defined(SIMDE_X86_SVML_ENABLE_NATIVE_ALIASES)
-  #undef _mm_div_epi32
```
nemequ, line 131: You do actually need the #undef.  In fact, you need to to keep the one for _mm_div_epi32 and add one for _mm_idiv_epi32.

These are necessary since SVML is defined by including `immintrin.h`.  If you use `immintrin.h`, everything is defined unconditionally; for example, you can call an AVX2 function even if you don't have `-mavx2` in your flags.  MSVC will allow this, assuming you've tested at runtime for AVX2 support, other compilers will usually emit an error (unless you're in a function with a [target](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-target-function-attribute) attribute.

Normally this isn't a problem; our function is `simde_mm_div_epi32`, which doesn't conflict with anything from SVML.  We then define a function-like macro, `_mm_div_epi32`, which rewrites all calls to the "real" `_mm_div_epi32` to go to our `simde_mm_div_epi32`.

The problem is that sometimes functions from the official APIs are defined as macros instead of regular functions.  In this case, there already *is* a macro called `_mm_div_epi32`, and the standard says you can only redefine an existing macro if the replacement is exactly the same (which it isn't).  So, to get around that we have to first undefine the existing `_mm_div_epi32` macro before we can create a new one.

This isn't a problem for SSE, since it has its own dedicated header, so we don't really need an #undef there.  But since functions in SVML use `immintrin.h` (along with everything from AVX on) we do need them there.

himanshi18037: I have updated the requested changes, Hopefully it should be fine this time

PR #293, opened 2020-05-14
svml: implement mm_rem_{epi8,epi16,epi32,epi64}

himanshi18037:


nemequ:
Really good work on this, BTW.  Any more problems with getting ICC to cooperate?

himanshi18037:
> Really good work on this, BTW. Any more problems with getting ICC to cooperate?

Not anything until now, I have got a hang of it in past few days, it's only a bit slow but we certainly can't do anything about it :)

PR #306, opened 2020-05-19
Clang 3.8

mr-c:


test/native-aliases.sh
```diff
@@ -19,8 +19,6 @@ if [ ! -e iig.xml ]; then
 fi
 
 PATTERN="$(xmllint --xpath '//intrinsic/@name' iig.xml | grep -Po '(?<=")[^"]+' | xargs printf '%s|' | rev | cut -c 2- | rev)"
+echo "s/([^_])simde(${PATTERN})/\1\2/g" > pattern
 
-for file in x86/*.c; do
-  echo "${file}..."
-  sed -i -E -e "s/([^_])simde(${PATTERN})/\1\2/g" "$file";
-done
+ls x86/*.c | xargs -n1 -P$(nproc) sed -i -E -f pattern
```
nemequ, line 24: Interesting, I didn't know about `xargs -P`.  i always just use GNU Parallel.

simde/hedley.h
```diff
@@ -74,6 +74,22 @@
...
+
+#if defined(HEDLEY_CLANG_VERSION_CHECK)
+#  undef HEDLEY_CLANG_VERSION_CHECK
+#endif
+#if defined(HEDLEY_CLANG_VERSION)
+#  define HEDLEY_CLANG_VERSION_CHECK(major,minor,patch) (HEDLEY_CLANG_VERSION >= HEDLEY_VERSION_ENCODE(major, minor, patch))
+#else
+#  define HEDLEY_CLANG_VERSION_CHECK(major,minor,patch) (0)
+#endif
+
```
nemequ, line 92: Hedley not having a macro like this isn't an oversight; you can't actually do this.

The problem is that various vendors (such as Apple) have their own compilers which are based on clang, and they define `__clang__` and `__clang_patchlevel__` to their version numbers.  https://sourceforge.net/p/predef/wiki/Compilers/#clang has some information on it.

The result is that __clang__ and __clang_patchlevel__ are pretty much useless.  You have to use the feature detection macros (`__has_attribute()`, `__has_builtin()`, `__has_feature()`, `__has_extension()`, `__has_warning()`), etc.  Of course, the problem here is there is no `__has_bug()` macro.  The best solution I've found is to look at the release notes for the next version of clang and defect a feature that just so happened to be added in that release.

For example, if you look a just a bit later in that file, you'll see

```c
/* diagnose_if + __builtin_constant_p was broken until clang 9,
 * which is when __FILE_NAME__ was added. */
#if defined(SIMDE_CHECK_CONSTANT_) && defined(__FILE_NAME__)
  #define SIMDE_REQUIRE_CONSTANT(arg) HEDLEY_REQUIRE_MSG(SIMDE_CHECK_CONSTANT_(arg), "`" #arg "' must be constant")
#else
  #define SIMDE_REQUIRE_CONSTANT(arg)
#endif
```

It's kind of a mess.  I love the idea of `__has_*` macros for feature detection, but not having a usable version number too is infuriating.

nemequ:
Merged, thanks.  Working its way through CI now.

BTW: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95227

mr-c:
> BTW: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95227

Thanks for the bug report!

PR #309, opened 2020-05-19
added native code for various x86 SSE instructions on Wasm

seanptmaher:
Hello! 

I wrote some native implementations of various SSE1 intrinsics for Wasm. 
Please let me know if any of the code is badly formatted/unidiomatic/etc. and I'll fix it up. 

You'll be seeing a lot more of me in the coming months, I'm an intern at Google working on Wasm SIMD. 

CC @ngzhian @tlively


mr-c:
Hello @seanptmaher and welcome!

nemequ:
Welcome @seanptmaher!  We're very glad to have the help!

If you have any questions please don't hesitate to reach out here on the issue tracker, on [the mailing list](https://groups.google.com/forum/#!forum/simde), in [the chat room](https://gitter.im/simd-everywhere/community).

This looks good; you've obviously got the idea well in hand.

In general, this is exactly the right way to add WASM implementations, but the `simde_mm_*_ss` functions are a bit of a special case..  If you just implement `simde_mm_move_ss` one time, all the `simde_mm_*_ss` functions will use that function without having to add a special case in each `*_ss` function.

This patch won't hurt anything, but since it will compile to the same thing as just adding a WASM implementation of `simde_mm_move_ss` I think it would be better to keep the code clean and just do that instead.

You can do the same thing with `simde_mm_move_pd` for the `simde_mm_*_pd` functions, but this is very much not the norm.

nemequ:
This looks great, thanks!  I'll merge it soon.

PR #311, opened 2020-05-20
Jsbache/fix mm cvttss si32

jsbache:
Fix for incorrect native alias for _mm_cvttss_si32. It should be simde_mm_cvtt_ss2si


nemequ:
It looks like the simde version is missing.  How about we also `#define simde_mm_cvttss_si32(a) simde_mm_cvtt_ss2si((a))` right before the if directive?

I'd still like to keep this patch, though; that way we limit the stack of macros that has to be printed to diagnose errors.  So the additional define would be in addition to this patch.

If you want to add it to the PR that would be great, otherwise I'll take care of it in a few hours.

jsbache:
Adding simde_mm_cvttss_si32 seems fine.
I simply noted that the Intel documentation for _mm_cvtt_ss2si and _mm_cvttss_si32 was identical, and I therefore figured that they were pseudonyms. If they are not, then we need to add an additional simde function & you are probably the best to do so. Thanks.

nemequ:
They're pseudonyms, but we still want to allow people to use `simde_mm_cvttss_si32`, if only for completeness.  No need to add a whole separate implementation, though, we can just cheat and `#define simde_mm_cvttss_si32(a) simde_mm_cvtt_ss2si((a))`.

Thanks for the patch, and good catch on finding the problem :)

On a somewhat different topic, we have tests for the native aliases now so there should be a *lot* fewer issues with them going forward.  Really the only ones we don't test are aliases like this one, and some macros like the rounding modes.  Definitely a better situation than when you submitted your previous patch set.

PR #312, opened 2020-05-20
Add the missing MM_GET_ROUNDING_MODE for sse native aliases.

GorePradnyesh:
Add the missing MM_GET_ROUNDING_MODE define for the cases where SIMDE_X86_SSE_ENABLE_NATIVE_ALIASES has been enabled.


nemequ:
Nice catch, thank you!

This does the right thing (which is the hard part) but it's not in the right place; native aliases should immediately follow the SIMDe versions (in this case, `SIMDE_MM_GET_ROUNDING_MODE`).  It looks like we copied the one for `_mm_extract_pi16`, but didn't update it to point to use the right function, so the correct location would be [sse.h:1817](https://github.com/nemequ/simde/blob/6ee040c68d6e0e96fdb9b760ec67a90c769c7391/simde/x86/sse.h#L1816).

I'll go ahead and update the patch; I'm going to need to rebase it anyways (since I need to merge several PRs tonight).

GorePradnyesh:
Awesome. Do you still want me to merge or should I close the PR then ?

nemequ:
You don't need to do anything; I'll close the PR when the updated commit hits master (actually, I'll add a line to the commit message that mentions this issue, which will make GitHub automatically close the PR and provide a way to link the commit to the PR request).

PR #321, opened 2020-05-24
sse4.2: first attempt at implementing mm_crc32_u8

masterchef2209:
sse4.2: first attempt at implementing mm_crc32_u8

I tried implementing the function from IIG. I know it has too many problems, but I thought it would be better to get some hint regarding implementation of the function and its tests.


simde/x86/sse4.2.h
```diff
@@ -241,6 +241,57 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
...
+int64_t bitreflection(int64_t n) {
+  int64_t rev = 0;
+  while(n > 0){
+    rev <<= 1;
+    if(n&1)
+      rev^=1;
+    n >>= 1;
+  }
+  return rev;
+}
```
nemequ, line 254: This is going to be quite slow.  It would be much better to use 256-byte lookup table ([example](https://graphics.stanford.edu/~seander/bithacks.html#BitReverseTable)).

clang has a [`__builtin_bitreverse` family of functions](https://releases.llvm.org/10.0.0/tools/clang/docs/LanguageExtensions.html#builtin-bitreverse) which should be used if it's available; check with `__has_builtin(__builtin_bitreverse64)`, but last time I checked emscripten and XL C/C++ didn't support it so you may need exceptions for them.

simde/x86/sse4.2.h
```diff
@@ -241,6 +241,57 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
...
+
+SIMDE_FUNCTION_ATTRIBUTES
+int64_t mod2div(int64_t dividend, int64_t divisor){
+  int64_t remainder = dividend;
+  int64_t sdata = floor(log2(dividend))+1;
```
nemequ, line 259: Instead of using libm directly, you should use the math functions in simde/simde-math.h.  IIRC you'll need to add a `simde_math_log2`, but `simde_math_floor` already exists.

simde/x86/sse4.2.h
```diff
@@ -241,6 +241,57 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
...
+      remainder^=divisor;
+    }
+    remainder=(remainder<<1);
+  }
+  return (remainder >> (sdata-sdiv+1));
+}
+
+SIMDE_FUNCTION_ATTRIBUTES
+unsigned int
+simde_mm_crc32_u8(unsigned int crc, unsigned char v) {
```
nemequ, line 275: You should use fixed-width types (`uint32_t` and `uint8_t`) here.

simde/x86/sse4.2.h
```diff
@@ -241,6 +241,57 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
   #define _mm_cmpistrz(a, b, imm8) simde_mm_cmpistrz(a, b, imm8)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+int64_t bitreflection(int64_t n) {
```
nemequ, line 245: These should all be in the simde namespace.  They're intended to be private APIs, so use a trailing underscore (e.g., `simde_mm_crc32_bitreverse_`, `simde_mm_crc32_mod2div_`).

Usually you could just make these static functions in a C file or a private header, or even just make them normal symbols but set the [visibility](https://gcc.gnu.org/wiki/Visibility) to hidden.  However, since SIMDe is a header-only library, we don't really have a way to hide implementations.

simde/x86/sse4.2.h
```diff
@@ -241,6 +241,57 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
...
+  return (remainder >> (sdata-sdiv+1));
+}
+
+SIMDE_FUNCTION_ATTRIBUTES
+unsigned int
+simde_mm_crc32_u8(unsigned int crc, unsigned char v) {
+#if defined(SIMDE_X86_SSE4_2_NATIVE)
+  return _mm_crc32_u8(crc, v);
+#else
+int64_t tmp1, tmp2, tmp3, tmp4, tmp5, tmp6, dst;
```
nemequ, line 279: You really don't need all these temporary variables.  Some operations can be combined (e.g., `tmp1 = bitreflection(v) << 32;`), but even if you don't do that you can reuse variables (`tmp1 = bitreflection(v); tmp1 <<= 32;`).

I'd probably do something like:

```c
int64_t tmp = (bitreflection(v) << 32)  ^ (bitreflection(crc) << 8)
tmp = mod2div(tmp, INT64_C(0x11EDC6F41));
return HEDLEY_STATIC_CAST(int32_t, bitreflection(tmp));
```

masterchef2209:
Thanks for review, I will go through all the comments and make changes accordingly.

PR #323, opened 2020-05-25
avx2: added mm{,256}_broadcastx_epi{16,32,64},ps,pd

masterchef2209:
avx2: added mm{,256}_broadcastx_epi{16,32,64},ps,pd

### DIFF
--- a/simde/x86/avx2.h
+++ b/simde/x86/avx2.h
@@ -663,6 +663,214 @@ simde_mm256_broadcastb_epi8 (simde__m128i a) {
 #endif
 
 SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_mm_broadcastw_epi16 (simde__m128i a) {
+#if defined(SIMDE_X86_AVX2_NATIVE)
+  return _mm_broadcastw_epi16(a);
+#else
+  simde__m128i_private r_;
+  simde__m128i_private a_= simde__m128i_to_private(a);
+
+  SIMDE_VECTORIZE
+  for (size_t i = 0 ; i < (sizeof(r_.i16) / sizeof(r_.i16[0])) ; i++) {
+    r_.i16[i] = a_.i16[0];
+  }
+
+  return simde__m128i_from_private(r_);
+#endif
+}
+#if defined(SIMDE_X86_AVX2_ENABLE_NATIVE_ALIASES)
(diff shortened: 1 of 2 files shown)


masterchef2209:
hi @nemequ , I am facing a very strange error, native version of mm_broadcastsd_pd is not available. What can be the reason for this?

nemequ:
That's not as uncommon as you might think.  Sometimes one compiler will add a function and the other compilers don't know anything about it.

In this case, the issue has already been reported an [GCC bug #92013](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92013).  As is pointed out in that bug, you can use SSE3's `_mm_movedup_pd` to implement it.  You can just make the implementation

```c
SIMDE_FUNCTION_ATTRIBUTES
simde__m128d
simde_mm_broadcastsd_pd (simde__m128d a) {
  return simde_mm_movedup_pd(a);
}
#if defined(SIMDE_X86_AVX2_ENABLE_NATIVE_ALIASES)
  #undef _mm_broadcastsd_pd
  #define _mm_broadcastsd_pd(a) simde_mm_broadcastsd_pd(a)
#endif
```

And it should work.

Another option would be to just create a special case for GCC:

```c
SIMDE_FUNCTION_ATTRIBUTES
simde__m128d
simde_mm_broadcastsd_pd (simde__m128d a) {
#if defined(SIMDE_X86_AVX2_NATIVE) && !defined(HEDLEY_GCC_VERSION)
  return _mm_broadcastsd_pd(a);
#else
  simde__m128d_private r_;
  simde__m128d_private a_= simde__m128d_to_private(a);

  SIMDE_VECTORIZE
  for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
    r_.f64[i] = a_.f64[0];
  }

  return simde__m128d_from_private(r_);
#endif
}
#if defined(SIMDE_X86_AVX2_ENABLE_NATIVE_ALIASES)
  #undef _mm_broadcastsd_pd
  #define _mm_broadcastsd_pd(a) simde_mm_broadcastsd_pd(a)
#endif
```

But I think the first option is a much better solution; that way if we have an optimized version of `_mm_movedup_pd` it will be called instead of having to reimplement it in `_mm_broadcastsd_pd`.

PR #239, opened 2020-04-26
Add gfni support in gfni.[hc]

rosbif:
Fixes #169

PR #328, opened 2020-05-27
avx512f: Add many broadcast, extract, insert and permute intrinsics

rosbif:
Add _mm512_broadcast_f32x4, mm512_broadcast_[fi]64x4
Add _mm512_extractf32x4_ps, _mm512_extractf64x4_pd
Add _mm512_extracti32x4_epi32, _mm512_extracti64x4_epi64
Add _mm512_insert[fi]{32,64}x4
Add _mm512_permutex{,2}var_{epi32,epi64,pd,ps]

Yet another case of bad vectorization by icc adding further support to my long standing opinion
that icc is crap.
And this time the code to be vectorized is extremely simple.

Emscripten failed but I probably don't have Evan's latest correction.

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -4447,7 +4447,9 @@ simde_mm512_permutexvar_epi32 (simde__m512i idx, simde__m512i a) {
       a_ = simde__m512i_to_private(a),
       r_;
 
-    SIMDE_VECTORIZE
+    #if !defined(__INTEL_COMPILER)
+      SIMDE_VECTORIZE
+    #endif
     for (size_t i = 0 ; i < (sizeof(r_.i32) / sizeof(r_.i32[0])) ; i++) {
       r_.i32[i] = a_.i32[idx_.i32[i] & 0x0F];
     }
@@ -4471,7 +4473,9 @@ simde_mm512_permutexvar_epi64 (simde__m512i idx, simde__m512i a) {
       a_ = simde__m512i_to_private(a),
       r_;
 
-    SIMDE_VECTORIZE
+    #if !defined(__INTEL_COMPILER)
+      SIMDE_VECTORIZE
+    #endif
     for (size_t i = 0 ; i < (sizeof(r_.i64) / sizeof(r_.i64[0])) ; i++) {
       r_.i64[i] = a_.i64[idx_.i64[i] & 7];
     }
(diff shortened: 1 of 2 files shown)


nemequ:
> Yet another case of bad vectorization by icc adding further support to my long standing opinion
> that icc is crap.

LOL :)

Unfortunately ICC tends to be the best in terms of optimization so a lot of projects that care about speed want to use it, which supporting at least the latest version important.

That said, in my experience all proprietary compilers are buggy and have an inferior feature set.  ICC actually seems to be the least buggy, followed by MSVC, then XL.  ICC also has the best feature set of any proprietary compiler, MSVC the worst.

Anyways, I'm going to try to generate a reduced test case for this to report the problem to Intel, but I'm not going to hold by breath for a fix (another thing about proprietary compilers is that they're pretty terrible about fixing bugs, at least unless you're a paying customer).

rosbif:
> > Yet another case of bad vectorization by icc adding further support to my long standing opinion
> > that icc is crap.
> 
> LOL :)
> 
> Unfortunately ICC tends to be the best in terms of optimization so a lot of projects that care about speed want to use it, which supporting at least the latest version important.
> 
> That said, in my experience all proprietary compilers are buggy and have an inferior feature set. ICC actually seems to be the least buggy, followed by MSVC, then XL. ICC also has the best feature set of any proprietary compiler, MSVC the worst.
> 
> Anyways, I'm going to try to generate a reduced test case for this to report the problem to Intel, but I'm not going to hold by breath for a fix (another thing about proprietary compilers is that they're pretty terrible about fixing bugs, at least unless you're a paying customer).

Yes, I fully concur that hardware manufacturers write bad compilers: c.f. what I wrote here https://github.com/nemequ/simde/pull/239#issuecomment-620508123

For me it is of absolute importance to have complete confidence in the compiler you use, otherwise you suspect the compiler for all your bugs. For this reason I refused to use Intel's compiler in the 1990s.

rosbif:
While checking these off in the list I noticed that all these intrinsics have mask and maskz versions.
Some even have a mask2 version.
I shall add these while I still have the test vector generation code.

I haven't looked yet but for this I think it will help me to move most of the movm and mask{,z}_mov stuff from VL to F. I think we agreed that this is OK.
However I think there could be problems, but maybe only with KNL, if we move the test code so I prefer to leave it in VL. I am not sure that we agreed on this :-(
I shall probably just do the moving first as the whole caboodle will take me some time and the merge could become complicated if the files change in the meantime.

Do we agree on all this?

nemequ:
> but maybe only with KNL

KNL/KNC is pretty much FUBAR anyways.  All those functions in the IIG guide that say something like "AVX512F for AVX-512, KNCNI for KNC"… all we test for is AVX-512F.  I don't even know what the preprocessor macros are for Knight's Landing.

IIRC Intel removed support for it from ICC, or maybe they just deprecated it.  Either way it's pretty much a dead platform.  I don't have any animus for it, but unless someone puts some serious effort into properly detecting it SIMDe is just going to be using the portable versions (or AVX/AVX2, SSE*, etc.).  Unless that happens you needn't concern yourself with KNL/KNC.

BTW, I left you a private message on Gitter.

PR #329, opened 2020-05-28
avx512f: added impl for mm512_{,mask,maskz}_and_epi{32,64}

masterchef2209:
avx512f: added impl for mm512_{,mask,maskz}_and_epi{32,64}


simde/x86/avx512f.h
```diff
@@ -2347,6 +2346,118 @@ simde_mm512_maskz_add_pd(simde__mmask8 k, simde__m512d a, simde__m512d b) {
...
+simde_mm512_mask_and_epi32(simde__m512i src, simde__mmask16 k, simde__m512i v2, simde__m512i v3) {
+  #if defined(SIMDE_X86_AVX512F_NATIVE)
+    return _mm512_mask_and_epi32(src, k, v2, v3);
+  #else
+    return simde_mm512_mask_mov_epi32(src, k, simde_mm512_and_epi32(v2, v3));
+  #endif
+}
+#if defined(SIMDE_X86_AVX512F_ENABLE_NATIVE_ALIASES)
+  #undef _mm512_mask_and_epi32
+  #define _mm512_mask_and_epi32(src, k, v2, v3) simde_mm512_mask_add_epi32(src, k, v2, v3)
```
nemequ, line 2388: It looks like this is what is causing the native aliases test to fail; you have the native alias defined to `add` instead of `and`.

PR #330, opened 2020-05-28
avx512{f,bw,vl}: Move all movm and mask{,z}_mov intrinsics to avx512f

rosbif:
These intrinsics are useful everywhere so put them in avx512f.
I have reordered them slightly but there are no modifications.

N.B. I haven't moved their tests. I prefer not to but should I?


nemequ:
> N.B. I haven't moved their tests. I prefer not to but should I?

Yeah, I think so.  I think we're going to have to split up avx512f.c anyways, but in the meantime it would be confusing having the tests in one file and the implementations in another since for all other functions (even the other ones we've moved to avx512f already) the test file matches up with the header file.

For NEON, I've been doing one file per function family (i.e., there is an add.c, ld1.c, st1.c, etc.).  What do you think of doing something similar for AVX512, at least for the tests?  I'd like to allow people to only include the functions they actually want for the implementations (i.e., avx512f.h, avx512bw.h, etc.), but for the tests I don't think we need to bother…

rosbif:
> > N.B. I haven't moved their tests. I prefer not to but should I?
> 
> Yeah, I think so. I think we're going to have to split up avx512f.c anyways, but in the meantime it would be confusing having the tests in one file and the implementations in another since for all other functions (even the other ones we've moved to avx512f already) the test file matches up with the header file.
> 

OK, if you accept these modifications I shall try to do that tomorrow.

> For NEON, I've been doing one file per function family (i.e., there is an add.c, ld1.c, st1.c, etc.). What do you think of doing something similar for AVX512, at least for the tests? I'd like to allow people to only include the functions they actually want for the implementations (i.e., avx512f.h, avx512bw.h, etc.), but for the tests I don't think we need to bother…
>

OK, but even for the main file there is nothing to stop avx512.h from including sub-files if necessary. For clarity their contents could even correspond to the test sub-files with the same base name.

As far as my emacs problem is concerned I am sure that it corresponds to accidentally selecting a large portion of the file. Often in order to find where to insert my code I go to the bottom of the file with the list of functions, copy the name of the function before the insertion point then search backwards. If I am not careful this can cause all the intervening code to be selected and emacs stalls. Otherwise emacs works acceptably with large files, although there is sometimes a slight but perceptible lag.

nemequ:
> OK, if you accept these modifications I shall try to do that tomorrow.

Sure, I'll pull them in soon.

> > For NEON, I've been doing one file per function family (i.e., there is an add.c, ld1.c, st1.c, etc.). What do you think of doing something similar for AVX512, at least for the tests? I'd like to allow people to only include the functions they actually want for the implementations (i.e., avx512f.h, avx512bw.h, etc.), but for the tests I don't think we need to bother…
> 
> OK, but even for the main file there is nothing to stop avx512.h from including sub-files if necessary. For clarity their contents could even correspond to the test sub-files with the same base name.

I've thought about that too, but every way I've thought of to split it seems pretty arbitrary.  Splitting it by function family doesn't really make sense because the functions tend to be spread out across multiple ISA extensions anyways.  It would work for the tests because we don't care about including all of the AVX-512* extensions, but I really don't want to force people to do that for the implementations.

I think the only option is going to be splitting by alphabetical order, plus one file to come first for the functions that other functions depend on.

PR #331, opened 2020-05-28
added mm512_{mask,maskz}_broadcast_{i,f}{32,64}x4

masterchef2209:
added mm512_{mask,maskz}_broadcast_{i,f}{32,64}x4

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -3548,6 +3548,32 @@ simde_mm512_broadcast_f32x4 (simde__m128 a) {
 #endif
 
 SIMDE_FUNCTION_ATTRIBUTES
+simde__m512
+simde_mm512_mask_broadcast_f32x4(simde__m512 src, simde__mmask16 k, simde__m128 a) {
+#if defined(SIMDE_X86_AVX512F_NATIVE)
+  return _mm512_mask_broadcast_f32x4(src, k, a);
+#else
+  return simde_mm512_mask_mov_ps(src, k, simde_mm512_broadcast_f32x4(a));
+#endif
+}
+#if defined(SIMDE_X86_AVX512F_ENABLE_NATIVE_ALIASES)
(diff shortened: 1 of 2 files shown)


simde/x86/avx512f.h
```diff
@@ -2682,6 +2760,32 @@ simde_mm512_broadcast_i64x4 (simde__m256i a) {
...
+#define _mm512_mask_broadcast_i64x4(src, k, a) simde_mm512_mask_broadcast_i64x4(src, k, a)
+#endif
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m512i
+simde_mm512_maskz_broadcast_i64x4(simde__mmask8 k, simde__m256i a) {
+#if defined(SIMDE_X86_AVX512F_NATIVE)
+  return _mm512_maskz_broadcast_i64x4(k, a);
+#else
+  return simde_mm512_maskz_mov_epi32(k, simde_mm512_broadcast_i64x4(a));
```
nemequ, line 2782: I believe this should be using `simde_mm512_maskz_mov_epi64` instead of `..._epi32`?  The tests will need to be regenerated.

When working on AVX-512 on a machine without AVX-512 support, you should use the [Intel Software Development Emulator](https://software.intel.com/en-us/articles/intel-software-development-emulator/) to generate the native tests.  Note that you don't need to run *all* the tests with it; you can just do something like `sde64 -- ./test/run-tests /x86/avx512f/mm512_mask_broadcast_i64x4` to run just this function, or `sde64 -- ./test/run-tests /x86/avx512f` (to run all the AVX-512F tests).

I find it can be handy to keep around two build directories when working on this; one with `-march=native` and one with `-march=icelake-server` (which will need to be run using the SDE).  Basically:

```sh
CFLAGS='-march=native' CXXFLAGS='-march=native' meson setup build && ninja -C build
CFLAGS='-march=icelake-server' CXXFLAGS='-march=icelake-server' meson setup build-sde && ninja -C build-sde
```

Then do most of the work with `build/`, but when you need to test the native version just do it from `build-sde/` instead.

PR #336, opened 2020-05-30
avx512f: Add missing and retouch existing movm and mask{,z}_mov functions

rosbif:
Add _mm{,256}_movm_epi64
Add _mm512_movm_epi{16,32,64]
Add _mm_mask{,z}_mov_p[ds]

Some casts are needed so move casts to the beginning.
Add some fallbacks and clean up existing movm and mask{,z}_mov functions.

### DIFF
--- a/simde/x86/avx512f.h
+++ b/simde/x86/avx512f.h
@@ -926,8 +926,8 @@ simde_mm512_movm_epi32 (simde__mmask16 k) {
     simde__m512i_private r_;
 
     #if defined(SIMDE_X86_SSE2_NATIVE)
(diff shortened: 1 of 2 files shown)


rosbif:
@nemequ: Thank you for merging this so quickly. Sorry for my impatience last time.
I don't know what time it was for you when you merged it. Don't you sleep at all? ;-)

nemequ:
Not at all.  With a big commit like that conflicts are a strong possibility, it makes sense to want to merge quickly.

I think it was about 3.  Don't get used to it, the insomnia has just been a bit worse than normal lately :)

Issue #310, opened 2020-05-19
Performance Warning Support


seanptmaher:
Hello,

First off, I've been looking over the code that's been written so far, and it's very solid. Thank you all for your work on the library. 

I've had some concerns expressed to me over performance-related issues in SIMDe, where due to the way that SIMDe is built, it will fallback to scalar operations on operations not supported on the target platform, but this isn't the behavior that some people would be looking for. 

How feasible would it be to implement warnings/errors when trying to compile instructions that would fallback to scalar operations on certain platforms? I assume this would be opt-in via #define. 

Having code silently run a lot slower perhaps than expected is not ideal for some applications. 

I'm happy to work on this myself, given a slight nudge in the right direction in terms of long-term vision for the project, it's possible this is out of scope of what you'd like to implement. 

Somewhere to look for a user with a similar problem is [here](https://github.com/emscripten-core/emscripten/pull/11193), where application performance is more important than code coverage. Because a lot of the work done here is duplicated in SIMDe, I feel that adding this support might be better than the alternative.

Thanks

(CC @ngzhian @tlively)

nemequ:
The best thing I can think of would be to use [`HEDLEY_WARNING`](https://github.com/nemequ/hedley/blob/d9261eeb6907b502dd06581932a6738ce4f65012/hedley.h#L1788) in simde-common.h to do something like:

```c
#if defined(SIMDE_WARN_ON_PORTABLE_FALLBACK)
  #define SIMDE_PORTABLE_FALLBACK_REACHED
#else
  #define SIMDE_PORTABLE_FALLBACK_REACHED HEDLEY_WARNING("Portable fallback reached, performance may not be optimal")
#endif
```

(I would be open to a different name).  Then, every time you hit a portable fallback in the code you would have to add `SIMDE_PORTABLE_FALLBACK_REACHED`.

While I'm sympathetic to the idea, the reason I haven't done this already is that a *lot* of the portable fallbacks can be optimized by the compiler every bit as well as by a human, as you can see by [defining `SIMDE_NO_NATIVE`](https://godbolt.org/z/fzXLZR).

[...]

Issue #278, opened 2020-05-09
_mm512_mask_cmpgt_epi32_mask: internal compiler error: Floating point exception


mr-c:
``` make
[ 37%] Building C object CMakeFiles/simde-test-emul.dir/x86/avx512f.c.o
/usr/bin/gcc -DSIMDE_BUILD_CPP_TESTS -DSIMDE_NO_NATIVE -I/simde/test/..  -Wall -Wextra -Werror -O3 -DSIMDE_ENABLE_OPENMP -fopenmp-simd -g   -std=gnu99 -Wno-psabi -Wno-psabi -o CMakeFiles/simde-test-emul.dir/x86/avx512f.c.o   -c /simde/test/x86/avx512f.c
simde/test/x86/avx512f.c: In function 'test_simde_mm512_mask_cmpgt_epi32_mask':
/simde/test/x86/avx512f.c:6527:1: internal compiler error: Floating point exception
 test_simde_mm512_mask_cmpgt_epi32_mask(const MunitParameter params[], void* data) {
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-7/README.Bugs> for instructions.
CMakeFiles/simde-test-emul.dir/build.make:281: recipe for target 'CMakeFiles/simde-test-emul.dir/x86/avx512f.c.o' failed
```

Dockerfile to reproduce this:
``` dockerfile
# bionic with gcc 7.4
FROM ubuntu:18.04
RUN apt-get update && apt-get install --no-install-recommends -y gcc g++ cmake make
COPY . /simde
WORKDIR /simde/test
RUN mkdir -p build_ubuntu18.04
WORKDIR /simde/test/build_ubuntu18.04
RUN CC=/usr/bin/gcc CXX=/usr/bin/g++ cmake -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
 -DCMAKE_C_FLAGS="-Wall -Wextra -Werror -O3" \
 -DCMAKE_CXX_FLAGS="-Wall -Wextra -Werror -O3"  .. \
 && make -j $(nproc) && ./run-tests
```

mr-c:
Also reproducible cross-building for arm64 using gcc 9.3

``` make
[ 58%] Building C object CMakeFiles/simde-test-emul.dir/x86/avx512f.c.o
[...]

PR #341, opened 2020-05-31
avx512f: Add many mask, mask2 and maskz intrinsics

rosbif:
Add _mm512_mask{,z}_extract{f32x4_ps,f64x4_pd,i32x4_epi32,i64x4_epi64}
Add _mm512_mask{,z}_insert[fi]{32,64}x4
Add _mm512_mask{,z}_permutexvar_{epi32,epi64,pd,ps}
Add _mm512_mask{,2,z}_permutex2var_{epi32,epi64,pd,ps}

Redo tests for _mm512_inserti32x4 which had bad coverage (b always zero).


rosbif:
Strange: Travis CI succeeded for me, otherwise I wouldn't have issued the PR.
However it failed later.
I think the problem is probably in the fact that idx is used both as an integral type and a as floating point type in _mm512_mask2_permutex2var_p[ds].
Initially I generated an integral idx but it led to a silly floating point value.
Then I generated a floating point idx but I figure when this is read on different machines it gives a different bit pattern, hence (I guess) the failures.
Now I store the floating point bit pattern as an integral type.

I now see more failures which don't seem correlated with my modifications.

I don't know what is going on but it will have to wait till the morning.

nemequ:
NP.  Just let me know when you want me to take a look at it, or if you need a hand.

I haven't looked at the code, but…

The -DSIMDE_NO_SHUFFLE_VECTOR build uses the default version of GCC on Bionic, which is 7.  So both failures right now are GCC 7.  Sounds like a potential compiler bug to me.  I'd probably try to reproduce it in an Ubuntu 18.04 (bionic) VM, then use creduce to generate a reduced test case, and hopefully that would tell you what you need to ifdef around.

GCC 7 is know to have problems with optimizing 512-bit vectors; see #278.  It wouldn't exactly be shocking to see another bug here.

rosbif:
> NP. Just let me know when you want me to take a look at it, or if you need a hand.
> 
> I haven't looked at the code, but…
> 
> The -DSIMDE_NO_SHUFFLE_VECTOR build uses the default version of GCC on Bionic, which is 7. So both failures right now are GCC 7. Sounds like a potential compiler bug to me. I'd probably try to reproduce it in an Ubuntu 18.04 (bionic) VM, then use creduce to generate a reduced test case, and hopefully that would tell you what you need to ifdef around.
> 
> GCC 7 is know to have problems with optimizing 512-bit vectors; see #278. It wouldn't exactly be shocking to see another bug here.
>

I have looked at this and this is what I found:
- the failures are only on /x86/avx512f/mm512_mask{,z}_insert[fi]32x4/native
- both the failures are identical: same functions, same results
- the failures are always on data32[8] corresponding to a switch from the low to high 256 bits of 512

If the /native really means that it is running the native instruction, I fail to see why all platforms succeed except GCC 7. I think this indeed suggests a compiler bug.
AFAICS an "ifdef around" would mean completely suppressing these tests on GCC 7.

However I fail to understand why GCC 7 succeeded on my Travis CI build and failed on the one started by the PR. This could give a clue. What is different?

rosbif:
@nemequ or @mr-c Do you have any idea why the -DSIMDE_NO_SHUFFLE_VECTOR test on Travis CI which succeeded when I committed to my fork, failed when I subsequently issued a PR on exactly the same data set?
I haven't modified the configuration of Travis CI but I imagine that there must be a difference.

nemequ:
> @nemequ or @mr-c Do you have any idea why the -DSIMDE_NO_SHUFFLE_VECTOR test on Travis CI which succeeded when I committed to my fork, failed when I subsequently issued a PR on exactly the same data set?
> I haven't modified the configuration of Travis CI but I imagine that there must be a difference.

We have a `cat /dev/cpuinfo` step, try checking the output of that to see if there are differences.  Maybe one machine has AVX-512 support and the other doesn't?  I haven't seen a Travis machine w/o AVX-512 in a while, though…

It's possible that the issue only appears if the build is targeting AVX-512, and since we use -march=native on CI maybe one worked and one didn't.  It kind of sucks that we can't always use the same flags, but we don't get to choose the hardware :(

rosbif:
Both the one that failed and the one that succeeded have an identical processor with AVX512 (F, DQ, CD, BW, VL, and VNNI).

rosbif:
It is not clear to me which code paths are executed during the tests.
Wouldn't it be possible to introduce an (always successful) pseudo-test which indicates clearly the state of the different #defines determining the code paths for the current platform?
(Maybe this information is already present but it isn't explicit enough for me. I only see native and emu but obviously they do not always indicate the same code path.)
If the same code path is executed by several platforms, maybe some failures should be acceptable, especially if the platform is known to be suspect. (Apart from the current GCC 7 case I am thinking mainly of ICC.)
The only exceptions I can see are:
- s390x must pass as AFAIK it is the only big-endian platform.
- compilation failures: often warning-errors which seem to be mainly generated by MSVC.
However I see that recently you have encountered some incorrect warnings from ICC.

nemequ:
> Wouldn't it be possible to introduce an (always successful) pseudo-test which indicates clearly the state of the different #defines determining the code paths for the current platform?

Yeah, we could do that.  Something like this?

```c
void print_status(const char* property, int enabled) {
  printf("%20s: %s\n", property, enabled ? "enabled" : "disabled");
}

#if defined(SIMDE_X86_AVX_NATIVE)
print_status("AVX", 1);
#else
print_status("AVX", 0);
#endif
```

I never really thought about it, but it's probably a good idea.  Of course you'd need to run it 4 times (emul/native, c/c++).  The trouble is we have a *lot* of these macros, but we could probably just start with the ones we need to know most often, and just build up the list over time.  Obviously the NATIVE ones, and a few from simde-common.h (like OpenMP, SHUFFLE, CONVERT, etc.).

> (Maybe this information is already present but it isn't explicit enough for me. I only see native and emu but obviously they do not always indicate the same code path.)

Not really, no.  Obviously it's encoded into the source code, but that's it, and a lot of the logic tends to get hidden away in macros.  That's basically what Hedley *is*, then there are a bunch more in simde-common.h, simde-features.h, etc.  It's definitely a bit of a mess.

Native basically means that we try to use SIMD functions when available, emul means we rely on the auto-vectorizer.  Beyond that there are still a large number of different configurations possible, and special cases for various compilers.

> If the same code path is executed by several platforms, maybe some failures should be acceptable, especially if the platform is known to be suspect. (Apart from the current GCC 7 case I am thinking mainly of ICC.)

I'm not sure I'm understanding you correctly here.  It sounds like you're basically suggesting that sometimes we should just punt on issues since it's really a compiler bug not a SIMDe bug, and tell people that if they want that functionality to work they should use a different/newer compiler.  If that's not what you're suggesting, maybe you could be a little more explicit?  If it is…

As a rule I'd rather not do this.  If possible we should work around the issue by creating a special case in the code.  We already have tons of these scattered throughout the tree.  I try to use `SIMDE_BUG_*` macros to help document them, but TBH sometimes I'm lazy and don't bother, especially for older compiler releases where the issue has already been fixed in newer versions.

That said, there will be some exceptions for when it is impossible to work around a particular issue, but so far [that is pretty rare](https://github.com/nemequ/simde/wiki/Compiler-Bugs).  We can deal with that by either not testing that particular configuration in CI, or if it's just a couple of functions we can disable those tests.

Unfortunately SIMDe seems to be *really* good at finding compiler bugs.  I think the only compiler we don't have any unresolved issues with right now is MSVC, and that's only because I don't bother reporting bugs to them… Emscripten is normally the exception because they have been *extremely* responsive to bug reports, but right now there is actually one (the patches are awaiting review before moving into clang).  It's not just ICC and an old version of GCC that's a problem, it's basically every compiler.

> compilation failures: often warning-errors which seem to be mainly generated by MSVC.

To be fair, that is because we use `/analyze /W4 /WX` to get MSVC to emit all warnings possible, and treat them as errors.  We do `-Weverything -Werror` on clang, and that has been every bit as much of a headache for me.  MSVC's warnings tend to be pedantic but valid, but a lot of warnings from clang are just stupid.

If you haven't already, you should take a look at simde/simde-diagnostics.h.  I've started to add explanations to the warnings we disable, but there is still work to do there.

> However I see that recently you have encountered some incorrect warnings from ICC.

And [GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95227), and [clang](https://bugs.llvm.org/show_bug.cgi?id=45959).

rosbif:
I could be wrong but it seems to me that in the current case SIMDe is just redirecting to the native intrinsic. So if the compiler can't handle the native intrinsic correctly it seems to me that you aren't doing the user any favours by hiding the fact that his compiler is broken. If you workaround the breakage you should at least issue a remark (I would have said a warning but I don't want a warning-error) that his compiler may not handle the instruction correctly. Otherwise he will be very surprised when his code fails if he runs it without SIMDe.

I suppose all this depends on your use case for SIMDe. In my case it was initially to port my code to wider vectors and new instructions (e.g. GFNI) for which I don't yet have the hardware. SIMDe with its fallbacks enables me to test my software with reasonable performance which is not the case with e.g. the Intel SDE. Subsequently I have just been trying to help the project in my spare time.

nemequ:
> I could be wrong but it seems to me that in the current case SIMDe is just redirecting to the native intrinsic. So if the compiler can't handle the native intrinsic correctly it seems to me that you aren't doing the user any favours by hiding the fact that his compiler is broken.

I don't think that's true.  I think the most common cases are running on non-x86 CPUs, and running on x86 CPUs that don't natively support the function in question.  In fact, these issues mostly only appear if we can't redirect to a native intrinsic and instead have to rely on autovectorization.

It seems to me that most people using SIMDe aren't using it to write new code (though as we add more AVX-512 and NEON functionality I hope to see more of that), but rather to port old code.  In that case, they already have a working (for example) AVX2 implementation that they want to run on a new platform which doesn't natively support AVX2.

> I suppose all this depends on your use case for SIMDe. In my case it was initially to port my code to wider vectors and new instructions (e.g. GFNI) for which I don't yet have the hardware. SIMDe with its fallbacks enables me to test my software with reasonable performance which is not the case with e.g. the Intel SDE. Subsequently I have just been trying to help the project in my spare time.

Ah, okay this definitely makes a lot more sense to me now.  I'm actually really happy to hear that that is what you're interested in because it's one of the use cases I had in mind, but I don't think it's the norm.  It's a shame you had to actually implement GFNI in order to play with it, but I do really appreciate it (and all the other patches)!

This particular problem is also a bit of a special case as far as these types of issues go.  In the vast majority of cases (probably in the 90-95% range) the problems don't occur when we're using the native intrisics, they occur in the portable fallbacks.  Even for the remaining 5-10%, the problem is almost always just a [missing intrinsic](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95483); if the compiler actually supports the intrinsic in question these types of errors are actually really uncommon, at least in GCC and clang.  And, among the "missing intrinsic" cases, the vast majority of those are actually for functions which are really just alternative APIs for another intrinsic which were added on after an initial release of the API, and we can just implement the relevant function using the original (likely ugly, but functional) API.

I think there have only been a handful of bugs like this so far, they're really not the norm.

> If you workaround the breakage you should at least issue a remark (I would have said a warning but I don't want a warning-error) that his compiler may not handle the instruction correctly.

If there is a performance issue I'd be sympathetic to adding a remark.  Unfortunately there isn't really a good way of doing this portably (you might want to take a look at #310, I mention a few implementation details there, and some of the problems; [GCC #95239](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95239) is particularly annoying).  I'm definitely open to suggestions about how to implement something like this, because TBH I can't really think of any *good* options.

If it's just a case of implementing one function using a slightly different API I don't think there is a need.

> Otherwise he will be very surprised when his code fails if he runs it without SIMDe.

I see your point, but I don't think the current situation is unreasonable.  If you run the code without SIMDe you're running it without the portability layer, so it makes sense to me that there might be portability issues.

The only case where I think optionally using SIMDe makes sense is if you don't want to create a hard dependency on SIMDe.  In this case they should be maintaining the non-SIMDe path (in all likelihood the non-SIMDe path is probably the default); the important thing for SIMDe is that we don't want to *decrease* portability.

Warning about broken compilers seems like a rabbit hole I'd rather avoid.  If we warn about the native GCC 7 implementation on GCC 7 with SIMDe, why would we not warn also warn about it on GCC 10?  Or clang?  What about on ARM?  I think the most sensible line to draw is that is to only warn if we're *currently* falling back on a slow path because we can't access the faster path.

nemequ:
Okay, I've got creduce working on generating a reduced test case.  It looks like the issue was fixed in 8.0, and only occurs at -O0.

We can work around this by just using the `_mm512_mask_mov_ps`-based fallback.  This still works at -O0, and at -O1 it generates [the exact same code](https://godbolt.org/z/Rh48G6), so performance shouldn't be effected if you enable optimization, and if you're at -O0 you don't care about performance anyways.

I *think* this was fixed with GCC r262566, but I'm not certain.

I'll push a commit soon to verify the fix.  Assuming it works I'll go ahead and merge this.

nemequ:
Looks like that did the trick.  Merged as 89ef7394d06e5aed541ff2a0789a932a85d4e35a, thanks!

I finally got the NEON stuff working, I'm merging it now, so if you're still interested in implementing NEON functions it's wide open :)

rosbif:
> > I could be wrong but it seems to me that in the current case SIMDe is just redirecting to the native intrinsic. So if the compiler can't handle the native intrinsic correctly it seems to me that you aren't doing the user any favours by hiding the fact that his compiler is broken.
> 
> I don't think that's true. I think the most common cases are running on non-x86 CPUs, and running on x86 CPUs that don't natively support the function in question. In fact, these issues mostly only appear if we can't redirect to a native intrinsic and instead have to rely on autovectorization.
> 

I think that we are talking at cross-purposes here. By "in the current case" I meant the GCC 7 failures that are preventing my additions from being merged. It seems to me clear, although I could be wrong, that these failures occurred while running native code on an x86 CPU with adequate AVX512 support.

> It seems to me that most people using SIMDe aren't using it to write new code (though as we add more AVX-512 and NEON functionality I hope to see more of that), but rather to port old code. In that case, they already have a working (for example) AVX2 implementation that they want to run on a new platform which doesn't natively support AVX2.
> 

OK, I don't dispute that.

> > I suppose all this depends on your use case for SIMDe. In my case it was initially to port my code to wider vectors and new instructions (e.g. GFNI) for which I don't yet have the hardware. SIMDe with its fallbacks enables me to test my software with reasonable performance which is not the case with e.g. the Intel SDE. Subsequently I have just been trying to help the project in my spare time.
> 
> Ah, okay this definitely makes a lot more sense to me now. I'm actually really happy to hear that that is what you're interested in because it's one of the use cases I had in mind, but I don't think it's the norm. It's a shame you had to actually implement GFNI in order to play with it, but I do really appreciate it (and all the other patches)!
> 
> This particular problem is also a bit of a special case as far as these types of issues go. In the vast majority of cases (probably in the 90-95% range) the problems don't occur when we're using the native intrisics, they occur in the portable fallbacks. Even for the remaining 5-10%, the problem is almost always just a [missing intrinsic](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95483); if the compiler actually supports the intrinsic in question these types of errors are actually really uncommon, at least in GCC and clang. And, among the "missing intrinsic" cases, the vast majority of those are actually for functions which are really just alternative APIs for another intrinsic which were added on after an initial release of the API, and we can just implement the relevant function using the original (likely ugly, but functional) API.
> 
> I think there have only been a handful of bugs like this so far, they're really not the norm.
> 

OK, I Think we understand each other better here.

> > If you workaround the breakage you should at least issue a remark (I would have said a warning but I don't want a warning-error) that his compiler may not handle the instruction correctly.
> 
> If there is a performance issue I'd be sympathetic to adding a remark. Unfortunately there isn't really a good way of doing this portably (you might want to take a look at #310, I mention a few implementation details there, and some of the problems; [GCC #95239](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95239) is particularly annoying). I'm definitely open to suggestions about how to implement something like this, because TBH I can't really think of any _good_ options.
> 
> If it's just a case of implementing one function using a slightly different API I don't think there is a need.
> 
> > Otherwise he will be very surprised when his code fails if he runs it without SIMDe.
> 
> I see your point, but I don't think the current situation is unreasonable. If you run the code without SIMDe you're running it without the portability layer, so it makes sense to me that there might be portability issues.
> 
> The only case where I think optionally using SIMDe makes sense is if you don't want to create a hard dependency on SIMDe. In this case they should be maintaining the non-SIMDe path (in all likelihood the non-SIMDe path is probably the default); the important thing for SIMDe is that we don't want to _decrease_ portability.
> 
> Warning about broken compilers seems like a rabbit hole I'd rather avoid. If we warn about the native GCC 7 implementation on GCC 7 with SIMDe, why would we not warn also warn about it on GCC 10? Or clang? What about on ARM? I think the most sensible line to draw is that is to only warn if we're _currently_ falling back on a slow path because we can't access the faster path.
>

OK, it's your call.

But, concretely, what do you want me to do in the case of these GCC 7 failures?
- I don't want to drop the 36 intrinsics I am adding because of what seems like 4 with very similar failures on a very old compiler.
- I suppose that I could pull the 4 failing intrinsics and reissue the PR with just the 32 remaining.
- I feel that it is a waste of time chasing what seems to me to be an old compiler incorrectly implementing these closely related intrinsics.

To sum up, at the risk of repeating myself, for the case of these GCC 7 failures:
- The test code in avx512f.c could be wrong but this is highly unlikely as it succeeds on all the other platforms. The compiler could be incorrectly implementing avx512f.c. This also seems highly unlikely as the same simple intrinsics are used in the other tests which succeed.
- The code in avx512f.h could be wrong. This is not impossible because some of my movm intrinsics have very counter-intuitive fallbacks due to missing 8-bit and 64-bit multiply and shift intrinsics in SSE and AVX. However AFAICS these failures occur when implementing the intrinsic natively, not in fallback or portable code. I don't see how the problem could be in the simple redirection in avx512f.h, especially as all other platforms are OK.

Hence my conclusion that the compiler is incorrectly implementing these closely related intrinsics.
It seems to me that all that can be done is to disable the native intrinsic for this compiler and IMHO (but we seem to differ on this) inform the user that this has been done.

If you agree with my diagnosis, could you please do this or should I just drop the 4 failing intrinsics?

rosbif:
> Looks like that did the trick. Merged as [89ef739](https://github.com/nemequ/simde/commit/89ef7394d06e5aed541ff2a0789a932a85d4e35a), thanks!
> 

Sorry, I hadn't seen your last two comments when I wrote the above. I thought that I had done a Ctrl-F5 but they only just now refreshed in my window :-( Thank you.
> I finally got the NEON stuff working, I'm merging it now, so if you're still interested in implementing

OK, cool. Thank you.

nemequ:
No worries.  Sorry I didn't get to this earlier; the NEON stuff was significantly more problematic than I'd hoped.  Largely because improving the code for the test cases revealed some rather nasty bugs; getting the NEON part to work was comparatively straightforward.

Even it's not really relevant for *this* issue anymore, this could come up again in the future.  So, for a theoretical function in the future where the compiler implementation of an intrinsic is buggy:

 * If it's possible to fall back on other intrinsics, that's preferable.  We got kind of lucky with this one because the work-around actually generates exactly the same code when optimization is enabled, but even if that weren't the case it would still probably be the fastest (and therefore preferred) option.  Once we have the performance warning code implemented I'd be willing to add a performance warning which is emitted when we fall back on the slow path.
 * If it's not possible to fall back on other intrinsics, fall back on the portable version.  Again, a performance warning here would be good.
 * If the portable version is still buggy, or if the problem was bad codegen from the portable version, try to find an alternate (possibly slower) implementation which does actually work as intended.
 * If we can't find any implementation that compiles to accurate code, disable that test case for that particular compiler, and we should probably seriously consider an error (not just a warning) for that compiler when calling that function.  Maybe that error would just be a missing function because it is ifdef'd out on that platform.

Does that sound reasonable to you?

PR #345, opened 2020-06-02
Make check.h self-contained, and add define directive

ngzhian:
In the spirit of
https://google.github.io/styleguide/cppguide.html#Self_contained_Headers,
we add a missing include to check.h to make it self contained.

Also, simde-diagnostic.h was missing the usual define directive in the
define guard.


nemequ:
Nice, thanks!

PR #353, opened 2020-06-07
sse4.2: implementation for mm_crc32_u8

masterchef2209:
I went through https://create.stephan-brumme.com/crc32/#git1 and https://github.com/stbrumme/crc32/blob/master/Crc32Best.ino implementation again, and I noticed that it is taking a previouscrc and a bunch of values and returning new crc32, and we need the same except the polynomial is different and we have one value v instead of array of values(data) so I tried using that implementation in this PR, but the difference is that this implementation is different from IIG. Please review this and also suggest how should I write tests for this function as this is different from other functions and skel.c is also different now so I am a little confused.

### DIFF
--- a/simde/x86/sse4.2.h
+++ b/simde/x86/sse4.2.h
@@ -269,6 +269,25 @@ simde_mm_cmpistrz_16_(simde__m128i b) {
   #define _mm_cmpistrz(a, b, imm8) simde_mm_cmpistrz(a, b, imm8)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+uint32_t
+simde_mm_crc32_u8(uint32_t prevcrc, uint8_t v) {
+#if defined(SIMDE_X86_SSE4_2_NATIVE)
+  return _mm_crc32_u8(prevcrc, v);
+#else
+uint32_t crc = prevcrc;
+crc ^= v;
+for(int bit = 0; bit < 8 ; bit++){
+  if(crc & 1) crc = (crc >> 1)^(0x82f63b78);
+  else        crc = (crc >> 1);
+}
+return crc;
+#endif
+}
+#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
+#  define _mm_crc32_u8(prevcrc, v) simde_mm_crc32_u8(prevcrc, v)
+#endif
+
 SIMDE_END_DECLS_
 
 HEDLEY_DIAGNOSTIC_POP
--- a/test/x86/sse4.2.c
+++ b/test/x86/sse4.2.c
@@ -891,6 +891,49 @@ test_simde_mm_cmpistrz_16(void) {
   return 0;
 }
 
+static int
+test_simde_mm_crc32_u8 (void) {
+  static const struct {
+    uint32_t crc;
+    uint8_t v;
+    unsigned int r;
+  } test_vec[] = {
+    { UINT32_C(3488119326),
+      UINT8_C(233),
+      UINT32_C( 661382116) },
+    { UINT32_C(4181338815),
+      UINT8_C(106),
+      UINT32_C(3873165213) },
+    { UINT32_C(3611029619),
+      UINT8_C(190),
+      UINT32_C(2087866855) },
+    { UINT32_C(3633137044),
+      UINT8_C(206),
+      UINT32_C( 975142830) },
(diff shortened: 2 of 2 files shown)


mr-c:
@masterchef2209 Yeah, I am also a bit confused by the new `skel.c` ; the older versions should still be useful. I think generating random numbers for both `crc` and `v` is a valid testing approach

nemequ:
I added`simde_test_codegen_random_{i,u}{8,16,32,64}` functions to make this a bit easier, and switched the `simde_mm_set1_epi8` test over to the new style as an example.  It's in staging now, I'll merge it as soon as I get it through CI.

In the case of `simde_mm_crc32_u8`, you would just use `simde_test_codegen_random_u32()` and `simde_test_codegen_random_u8()` to generate the parameters, and `simde_test_codegen_write_u32` and `simde_test_codegen_write_u8` to write them.

> Yeah, I am also a bit confused by the new `skel.c` ; the older versions should still be useful. I think generating random numbers for both `crc` and `v` is a valid testing approach

The old integer test generation functions should still work, but for the float ones some supporting functions (the ones for generating random values) have been removed.

The new versions have quite a few benefits, including:

 * Better design — the old API was a horrible mess.  Functions were spread out all over the place with logic duplicated.  The new version is much cleaner and more maintainable, and implementation details are restricted to internal APIs.
 * Type safety — There were several functions in the old version that would basically accept anything and cast the input to the proper type, which meant it was really easy to use the wrong functions and the compiler would happily accept it.
 * Support for NaNs and infinities — I know the emscripten people were interested in this, but really it's something we should have had a long time ago.  Since the logic is shared it only needs to be handled once.
 * Correctness — Instead of hardcoding printfs the logic is centralized, so we it's feasible to actually output correctly formatted code.  For example, instead of just doing `INT8_C(%hhd)`, we can move the sign *outside* of the `INT8_C` macro, where it is supposed to be.  We can also correctly handle INT*_MIN without running into errors, which I think Himanshi, Hidayat, and Christopher have all run into at different times.
 * Easier to use — The tests are actually a lot simpler to write since more of the logic is centralized instead of being distributed in each test.
 * Doesn't use private APIs — the old tests used `simde_*_private` types.  We want to test the public API, not the hacks we have to use to make it work behind the scenes.
 * Easier to use without µnit — this should make it easier to use creduce with the tests.  I'd like for it to be as simple as `-DSIMDE_TEST_BARE`; we're not there yet, but I'll spend some time working on that next time I need to use creduce.
 * Less boilerplate — for example, some of the cruft has been moved into common footers.
 * The test runner code is *much* cleaner now — adding new suites is a lot more straightforward, which is nice because the ARM tests are one suite per family.
 * Better separation between x86 and common code — before, a lot of x86 stuff was mixed into the common code, and stuff that should have been common was x86-specific.

The new API might take a bit of getting used to, but I think everyone will be happier in the long run.  In the short term there may be some missing functionality (like a convenient API for generating random integers was), but it should be much easier to add that functionality as we discover it is missing.

masterchef2209:
@nemequ update on crc:
I tried writing tests for this function but unfortunately it failed. So I tried digging deeper into CRC stuff (went through README of [this repo](https://github.com/Michaelangel007/crc32). So, I got to know that there are 2 versions of CRC normal and reflected (depending on if the polynomial is reversed or not), and the CRC checksum can have 2 different values for same polynomials depending on if we are reflecting data or not. Also I saw (this)[https://stackoverflow.com/questions/17645167/implementing-sse-4-2s-crc32c-in-software] thread and a refined implementation [here](https://github.com/anandsuresh/sse4_crc32) which looks like something we need as it has software-based fallback. From that implementation and (wiki)[https://en.wikipedia.org/wiki/Cyclic_redundancy_check#table] I figured which version of CRC we need and that forward polynomial is 0x1EDC6F41 and the reflected polynomial is 0x82F63B78. So I corrected the polynomial and still the tests didn't pass. After that I copied the implementation from the repo linked above but it still produces the same result as my code(different from native), one interesting thing to notice that in IIG the polynomial is 0x11EDC6F41 instead of 0x1EDC6F41 (which is also asked in stackoverflow thread but no clear answer there) so I am confused if the CRC in SSE4.2 is actually CRC33 and not CRC32, but the github repo linked above also have intrinisic implementation so that shouldn't be the case. At present, I am trying to setup the project on my system to see if I made some mistake.

PR #357, opened 2020-06-08
NEON: Add and, eor and orr intrinsics

rosbif:
I am pushing these easy intrinsics as unfortunately I don't have much spare time at the moment.
However I shall try to push more stuff as soon as possible.


nemequ:
These look great, thanks!

rosbif:
I forgot to check that all these were in NEON A32V7 and AltiVec P5.
I guess they are as they are pretty basic.
Would some tests have failed if I got this wrong or should I really check for each intrinsic in future?

PR #358, opened 2020-06-08
Add implementation of dup_n.h

seanptmaher:
Hi,

Let me know if anything's off here. 

Thanks,
Sean
P.S. Thanks a lot for the NEON rewrite, it's a lot easier to deal with this way

CC @ngzhian @tlively


seanptmaher:
Hm, I'm not sure what's causing the undefined reference -- I've added it to CMakeLists, and it builds on my machine...

nemequ:
> Hm, I'm not sure what's causing the undefined reference -- I've added it to CMakeLists, and it builds on my machine...

It's not in meson.build.  We currently have both meson and CMake build systems.  All the CI other than Travis is using Meson; as soon as we get Travis switched I'm planning on dropping CMake.

simde/arm/neon/dup_n.h
```diff
@@ -0,0 +1,519 @@
...
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Copyright:
+ *   2020      Evan Nemerson <evan@nemerson.com>
```
mr-c, line 24: Should this be @seanptmaher instead of @nemequ?

nemequ:
> Let me know if anything's off here.

At first glance it looks really good.  I only see a couple **extremely** minor issues.  The larger one is that your name should definitely be in the copyright declaration.

The other is downright trivial; the formatting is a bit off for the loops in the portable version.  Instead of

```c
    simde_float32x2_private
      r_;

      SIMDE_VECTORIZE
        r_.values[i] = value;
      }

    return simde_float32x2_from_private(r_);
```

It should be

```c
    simde_float32x2_private r_;

    SIMDE_VECTORIZE
      r_.values[i] = value;
    }

    return simde_float32x2_from_private(r_);
```

The r_ being on the next line in the declaration is just for when there are multiple values we want to initialize; for example

```
simde_float32x2_private
  a = simde_float32x2_to_private(a),
  b = simde_float32x2_to_private(b),
  r_;
```

I get that you were trying to follow the existing style (which I *really* appreciate), but when the variable is all alone I don't see much need to put it on its own line.  `dup` is just a bit different than most of the others since it only takes the one parameter and we don't even need to convert it to an internal type.

Also, the loop should be indented to line up with the beginning of the other statements.  Indenting should *mostly* be what you would expect.  The only thing that is a little odd about is that preprocessor directives are treated the same as normal code.  For example:

```
if (foo) {
  #if defined(bar)
    if (baz) {
      #if defined(qux)
        quux();
      #endif
    }
  #endif
}
```

Not all of SIMDe adheres to that, new code (including all the NEON stuff) should.

You don't have to worry about "fixing" anything.  I'll take care of it when I merge (probably this evening); I just wanted to mention them for future reference.

simde/arm/neon/dup_n.h
nemequ: Yes please :)

It can be in addition to my name if you feel like the parts you copied were significant, I don't really care either way.  But your name should definitely be there.

I know I'm always a little uneasy about adding myself to the copyright in someone else's project, but please don't feel that way about SIMDe.  I'm quite happy to have everyone who has helped even with relatively minor things in the list, and this is definitely a non-trivial contribution (thanks, by the way!).

seanptmaher: Alright, I added myself in with Evan there :)

seanptmaher:
Alright, should be all good now, I rebased it as well to package it up.

seanptmaher:
Nevermind, I just saw your comment on the indentation. Fixed that. Don't worry about it, it's really easy to make these trivial changes, and my schedule is probably less hectic than yours :wink:

rosbif:
Wouldn't it be a good idea to try to keep (approximate) alphabetical order for the lists unless there is a good reason not to?
(However I generally put 8 before 16, 32 and 64 although alphabetically it should be last.)

PR #359, opened 2020-06-08
Fix array sizes and native aliases

seanptmaher:
Hey, I noticed the array sizes and argument lists for the vld1 and vst1 intrinsics weren't quite matching up, so I rewrote some of them. 

Let me know if something's missing here, but I'm _pretty sure_ this should be right.


seanptmaher:
Also, if my changes are in fact correct, this might show that the CI doesn't properly verify the native aliases, which could be a problem.

nemequ:
Your patch looks right to me.  You're correct that we don't have a native aliases test for ARM in CI yet, I'll look into adding it.

PR #360, opened 2020-06-08
Add type aliases when exposing native aliases

seanptmaher:
I noticed that when defining the native aliases, the types weren't aliased. 

I added that in here. Hope it's kosher.

### DIFF
--- a/simde/arm/neon/types.h
+++ b/simde/arm/neon/types.h
@@ -272,6 +272,29 @@ SIMDE_ARM_NEON_TYPE_FLOAT_DEFINE_(64, 2, 16)
   typedef simde_float64 simde_float64_t;
 #endif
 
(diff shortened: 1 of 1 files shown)


simde/arm/neon/types.h
```diff
@@ -272,6 +272,29 @@ SIMDE_ARM_NEON_TYPE_FLOAT_DEFINE_(64, 2, 16)
   typedef simde_float64 simde_float64_t;
 #endif
 
+#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES) || defined(SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES)
```
nemequ, line 275: This should just be `#if defined(SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES)`.  With the exception of `float64x1_t` and `float64x2_t` all of these are part of the NEON API from ARMv7.

Then there should be a second group, with an `#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)`, containing `float64x1_t` and `float64x2_t`.

PR #363, opened 2020-06-09
neon: Implementation of get_high and get_low

seanptmaher:
These are not fast, but I couldn't really figure out a way to do this
fast on SSE, and Wasm doesn't support 64-bit vectors, so I think we
might be stuck with slow scalar fallbacks (I'm not sure how slow
these are, though)

I also changed skel.c because I noticed that the default was to use 16-len vectors in non-q vec ops. 

(CC @tlively @ngzhian)


nemequ:
Awesome, thanks!

> These are not fast, but I couldn't really figure out a way to do this fast on SSE, and Wasm doesn't support 64-bit vectors, so I think we might be stuck with slow scalar fallbacks (I'm not sure how slow these are, though)

For get low there is [`_mm_movepi64_pi64`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=__m64%20_mm_&techs=SSE,SSE2&expand=3880).  I suspect that most compilers should be able to autovectorize to that anyways.

I know you're primarily interested in WASM, if you want to just do the portable (and WASM if possible) implementations that's fine with me; we can go in later and add x86, POWER, etc.

For these operations I suspect there won't be much of a difference since they're operating on 64-bit chunks of data, and x86_64 supports that as a non-vector.  For most of the 

> Edit: I always fail the formatting bot pensive

Yeah, our CI is a huge pain, sorry.  I didn't see the patch before you fixed this, but everything looks good now, I'll try to merge this right away.

PR #364, opened 2020-06-09
neon: Add implementation of mul

seanptmaher:
I had to modify the codegen print functions to only print `%f` instead of `%9.2f`, as the latter was producing incorrect values (off by about 0.5), across both arm and x86 when testing floats. I hope that's fine.

Also, there was an incorrect wasm add instruction that I changed from 32x4 to 64x2. 

(CC @ngzhian @tlively)


seanptmaher:
Looks like there are some merge conflicts, I'll try and fix those

seanptmaher:
Hm, apparently vmul_s64 doesn't exist, despite running on my raspi, I guess I have the wrong CPP directive. Might take me a day to get to this though.

nemequ:
> I had to modify the codegen print functions to only print `%f` instead of `%9.2f`, as the latter was producing incorrect values (off by about 0.5), across both arm and x86 when testing floats. I hope that's fine.

I think a better solution would be to add some rounding code to the random number generation code.  I don't think it's worth regenerating the test vectors, though, so I'll accept this patch (and the mla one) as-is and just add the rounding code for the future.

> Looks like there are some merge conflicts, I'll try and fix those

There should be fewer conflicts as the number of NEON functions grows; since the list of implemented functions is so short right now there is just a high probability of conflicts :(

> Hm, apparently vmul_s64 doesn't exist, despite running on my raspi, I guess I have the wrong CPP directive. Might take me a day to get to this though.

Are you sure you're passing the right flags to enable NEON on your compiler?  I don't see `vmul_s64` in [GCC's arm_neon.h](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/arm/arm_neon.h); maybe you're hitting the portable fallbacks?  Try adding an `#error whatever` preprocessor directive inside one of the NEON sections to check.

If you're not enabling NEON, the right flags rather annoyringly depend on the compiler and platform.  Something like `-march=armv8-a+simd -mfpu=neon` usually does the trick (though you may need to drop the "+simd") if you're on ARMv8+.  IIRC Raspbian is still using ARMv7, though, so if you're using that you'll probably want `-march=armv7-a -mfpu=neon`.

My RPi3 is pretty slow, but I haven't tried with an RPi4.  It's a lot faster for me to cross-compile and use qemu to run the binaries.  There is some info [in the wiki](https://github.com/nemequ/simde/wiki/Development-Environment#emulation) about it; unfortunately I think Ubuntu is a bit different, but even in a VM it's much faster.

PR #365, opened 2020-06-09
neon: Implementation of mla instruction

seanptmaher:
Here's the implementation of mla. 

However, I'm unsure if I'm doing the CPP stuff right for arm devices. I read through the code that you have in `simde-features.h` and there were some things that didn't make sense to me at first glance, so I'll talk to you about them later, probably on gitter. @nemequ 

(cc @tlively @ngzhian)


nemequ:
Sorry, this took a little longer than usual to merge.  It looks like it was causing some unrelated code to fail; see 053960c99a4120a67d39d179ee7be2b9f76ea33b for the temporary fix "fix", but I think I'm going to have to change the type of the mem_addr parameters to for the `simde_mm*_loadu_*` functions to `void*`.

I added some SSE/SSE2 and AltiVec implementations, plus several FMA implementations which I accidentally squashed into your path (sorry!).

PR #366, opened 2020-06-10
neon/test: reformat start of skel.c file

seanptmaher:
This is a very small change, but allows proper modification of the file via regexprs to more easily adapt tests without having to manually rewrite part of the file. Small nitpick/fix, but I'd appreciate
it being merged just for the convenience.


nemequ:
Seems reasonable enough, thanks!

PR #368, opened 2020-06-11
neon: Add shl_n and shlq_n intrinsics

rosbif:
Should I have called the files shl_n?
However the shl and shlq intrinsics can still be added later.


rosbif:
I am not sure why formatting couldn't find shl.h.
The other tests have obviously found it.

rosbif:
Could you hold off merging this now please?
I hope I have nearly finished the shr_n and shrq_n intrinsics and I shall add them here.

rosbif:
I have added the shr_n and shrq_n intrinsics but I named the files without the _n.
Others can be added later but I don't think that the strange vshl intrinsics (which also shift right) translate well into other SIMD instruction sets.

rosbif:
OK, I have finally understood what formatting is saying. I originally thought it was a spurious message saying that shl.h didn't exist. I only just realized that it was saying it wasn't included by neon.h.

Shouldn't the test files include neon.h instead of e.g. shl.h so that this error is detected early?

rosbif:
Just to say that the AppVeyor failure isn't my fault.
I can't see how to restart it without sending a dummy commit :-(

nemequ:
> Shouldn't the test files include neon.h instead of e.g. shl.h so that this error is detected early?

I want to make it possible to include only the functions you need for NEON if you want.  If we want that feature, this is really the only option… if we include neon.h we can't really test that any dependencies have been included (e.g., if you're implementing vpmin with vmin, you need to include min.h).

The CI check is there to make sure you don't forget to add it to arm/neon.h, so we cover both cases.

rosbif:
> The CI check is there to make sure you don't forget to add it to arm/neon.h, so we cover both cases.
>

Yes, but IMHO it arrives too late: it is not detected either by the local tests or by the Travis CI tests.
You only realize after issuing a PR leading to wasted time and wasted Travis (and other) CI resources.

nemequ:
I understand, but with what you propose we lose the ability to test only 
including what you need, and of the two I think that's the more error-prone.

Maybe the formatting checks should be in a script you can run locally? 
You'd still have to remember to run it, but you could add that as a 
pre-commit hook...

rosbif:
I was thinking about implementing the vshl and vshlq macros next (those without the `_n_`).
This will complete shl.h (shr.h is already complete).
(Remark: there are also other shift instructions but they have different mnemonics.)
I was waiting for you to merge these `_n_` ones first but should I continue on top of this PR as I did for shr.h?

nemequ:
Sorry, I got a bit behind on merging PRs this week.

I'm going to move this to `shl_n.h`.  My original intent was to actually use `shl.h` for all the different versions, but after @seanptmaher submitted the `dup_n` stuff in its own file I found I quite liked the idea.  I think it should make it a little easier to manage, especially when copying new a file over to start working on a new function.  We can always merge them in the future if the number of files gets out of hand, but I don't think it should be a problem.

BTW, I was really just thinking about making C-Reduce easier to use, but I thought of another major advantage of using insn.h instead of neon.h; incremental compilation should be a *lot* faster since the build system doesn't need to recompile everything; just the object files which use the headers that changed, then re-link eveything.

Anyways, in the future if I get behind on the PRs you can always just keep going and I'll figure out how to merge it.  I figure if the merge gets messy because I was slow that's on me, you shouldn't have to worry about it.

rosbif:
One minor nitpick: the macros to prevent multiple inclusion should also have _N_, particularly for shl_n as they will conflict with the logical name of the future file containing the missing shl intrinsics.
AFAICS they don't.

nemequ:
> One minor nitpick: the macros to prevent multiple inclusion should also have `_N`, particularly for `shl_n` as they will conflict with the logical name of the future file containing the missing shl intrinsics.
> AFAICS they don't.

Ah, you're absolutely right, thanks ☺

I noticed a few other things while I was reviewing this; I already changed them, but for future reference:

 * The parameter names should match the documentation.  There were a ton of `b` variables that are called `n` in the NEON docs.  Obviously it's `b` works, but it's nice to be consistent with the docs.
 * The shift amounts don't need to be `int32_t`; `int` is appropriate here.  AFAICT there is no need to worry about the width of these parameters, since the value is never more than 64 (63?).
 * I move a bunch of implementations from macros into the functions.  I know you thought that the x86 ones required an immediate-mode parameter, but even if that were true the WASM ones don't, there is no need to make them macros for defensive purposes.  This actually revealed a couple of minor bugs regarding types because now the input has to be compatible with int, which is then cast to unsigned char/short/int/long long in the function.

I feel good about this CI run on on my other repo, I think I'll have this merged into master in an hour or so.

> OK, I agree that the number of files should remain reasonable but, if it does get out of hand, I am not volunteering to merge a significant number of them ;-)

No problem, I'll take care of it.  It should actually be really easy; splitting them up would be a bit of a pain, but merging seems pretty straightforward.  The only problem I see is that it would break backwards compatibility for the whole only include what you need idea.  Maybe I should add a guard against including the _n versions directly and just tell people to include the  "main" header for each instruction…?

rosbif:
> I noticed a few other things while I was reviewing this; I already changed them, but for future reference:
> 
>     * The parameter names should match the documentation.  There were a ton of `b` variables that are called `n` in the NEON docs.  Obviously it's `b` works, but it's nice to be consistent with the docs.
> 

Sorry, I think the b's were a leftover from the example I copied, probably add.h.

>     * The shift amounts don't need to be `int32_t`; `int` is appropriate here.  AFAICT there is no need to worry about the width of these parameters, since the value is never more than 64 (63?).
> 

Agreed.

>     * I move a bunch of implementations from macros into the functions.  I know you thought that the x86 ones required an immediate-mode parameter, but even if that were true the WASM ones don't, there is no need to make them macros for defensive purposes.  This actually revealed a couple of minor bugs regarding types because now the input has to be compatible with int, which is then cast to unsigned char/short/int/long long in the function.
> 

I still maintain that these require an immediate-mode parameter: the IIG even gives the assembler instruction which should be generated. While current compilers seem to work around this limitation, there is absolutely nothing preventing a future compiler from conforming strictly with the IIG. So I don't really agree, but it is your call.
Sorry for the bugs :-(

> I feel good about this CI run on on my other repo, I think I'll have this merged into master in an hour or so.
> 
> > OK, I agree that the number of files should remain reasonable but, if it does get out of hand, I am not volunteering to merge a significant number of them ;-)
> 
> No problem, I'll take care of it. It should actually be really easy; splitting them up would be a bit of a pain, but merging seems pretty straightforward. The only problem I see is that it would break backwards compatibility for the whole only include what you need idea. Maybe I should add a guard against including the _n versions directly and just tell people to include the "main" header for each instruction…?
>

I don't think that there will be a "main" header for shr. AFAICS shr is already complete as all shr instructions are _n.

nemequ:
> Sorry, I think the b's were a leftover from the example I copied, probably add.h.

Yeah, that was my guess.  Obviously not a big deal.

> > ```
> > * I move a bunch of implementations from macros into the functions.  I know you thought that the x86 ones required an immediate-mode parameter, but even if that were true the WASM ones don't, there is no need to make them macros for defensive purposes.  This actually revealed a couple of minor bugs regarding types because now the input has to be compatible with int, which is then cast to unsigned char/short/int/long long in the function.
> > ```
> 
> I still maintain that these require an immediate-mode parameter: the IIG even gives the assembler instruction which should be generated. While current compilers seem to work around this limitation, there is absolutely nothing preventing a future compiler from conforming strictly with the IIG. So I don't really agree, but it is your call.

I kind of agree with you in the abstract, but there are simply too many programs in the wild that depend on this behavior.  If some future compiler actually requires them to be an ICE we can work around it for them by adding a macro which is only exposed to that compiler, but I really don't think that's likely to happen.

Meanwhile, there are real benefits (mostly from type safety) to having the implementations in functions, and it's a *lot* more readable.  IMHO the trade-off is more than worth it.

> Sorry for the bugs :-(

Not at all.  None of the stuff I mentioned were really bugs anyways.

> I don't think that there will be a "main" header for shr. AFAICS shr is already complete as all shr instructions are _n.

Oh, right.  IIRC for shr (not shr_n) you're supposed to use shl with a negative value.  It's such an odd thing for NEON to be missing; most of their APIs are so complete (especially compared to Intel's).

Anyways, I realized after I wrote this that if peple thing there are too many headers they can always just use the amalgamated one instead, so I don't think this is an issue anyways.

rosbif:
> > > ```
> > > * I move a bunch of implementations from macros into the functions.  I know you thought that the x86 ones required an immediate-mode parameter, but even if that were true the WASM ones don't, there is no need to make them macros for defensive purposes.  This actually revealed a couple of minor bugs regarding types because now the input has to be compatible with int, which is then cast to unsigned char/short/int/long long in the function.
> > > ```
> > 
> > 
> > I still maintain that these require an immediate-mode parameter: the IIG even gives the assembler instruction which should be generated. While current compilers seem to work around this limitation, there is absolutely nothing preventing a future compiler from conforming strictly with the IIG. So I don't really agree, but it is your call.
> 
> I kind of agree with you in the abstract, but there are simply too many programs in the wild that depend on this behavior. If some future compiler actually requires them to be an ICE we can work around it for them by adding a macro which is only exposed to that compiler, but I really don't think that's likely to happen.
> 

I think we must agree to disagree again on this :-(
You are relying on undocumented behaviour and I don't like that.
(Sorry I can't write behavior: I'm English ;-) )

> Meanwhile, there are real benefits (mostly from type safety) to having the implementations in functions, and it's a _lot_ more readable. IMHO the trade-off is more than worth it.
> 

Maybe you are almost starting to come around to my view that, apart from extensions to the language, `if` should be preferred to `#ifdef` inside functions as it enables the compiler to compile-test more code paths ;-)

> > I don't think that there will be a "main" header for shr. AFAICS shr is already complete as all shr instructions are _n.
> 
> Oh, right. IIRC for shr (not shr_n) you're supposed to use shl with a negative value. It's such an odd thing for NEON to be missing; most of their APIs are so complete (especially compared to Intel's).
> 

Yes, exactly shr doesn't exist and you have to use shl with a negative value. But I don't like the l in the mnemonic. You can't even say it's l for logical, not left, because there are also arithmetic shifts.

PR #370, opened 2020-06-11
neon: Implementation of padd instruction

seanptmaher:
Here's another one, hopefully with less weird constraints like `vget_lane`.

(cc @ngzhian @tlively)


seanptmaher:
Oh, I see the mistake here, I'll fix it

nemequ:
There is a relatively efficient way to implement pairwise operations on most architectures.  If you shuffle the inputs so you have one vector composed of the even elements from both inputs, and a second one composed of the odd elements, all you have to do is perform the operation (add) on them.

The shuffles are usually pretty fast.  For 32 bits on x86 there is `_mm_shuffle_ps`/`_mm_shuffle_epi32`. For 64 bits, `_mm_unpacklo_pd` and `_mm_unpackhi_pd` / `_mm_unpacklo_epi64` and `_mm_unpackhi_epi64`.   `_mm_shuffle_ps`, for example maps to [`shufps`](https://uops.info/html-instr/SHUFPS_XMM_XMM_I8.html), which is about as fast as you can get (it's right there with operations like xor).

For relatively recent versions of clang and GCC, you can use `__builtin_shuffle`/`__builtin_shufflevector` to do it portably; SIMDe has a [`SIMDE_SHUFFLE_VECTOR_`](https://github.com/nemequ/simde/blob/master/simde/simde-common.h#L299-L323) macro to abstract away the differences.  pmin.h has some examples of using `SIMDE_SHUFFLE_VECTOR_` for exactly this purpose.

More interestingly, though, NEON also has functions to do exactly what we want: `vuzp1*` and `vuzp2*`.  If we implement those functions we can use them to implement all the pairwise functions by just doing something like

```c
return vaddq_s32(vuzp1q_s32(a, b), vuzp1q_s32(a, b));
```

I think the best thing to do would be to implement `uzp1` and `uzp2` (pmin.h already has the logic necessary, it just needs to be reworked a bit to only do one of the shuffles (per file) instead of both shuffles and a `min`.  Once that's done, we can use that to implement this and pmin.

Is that something you would be interested in doing, or would you prefer for me to do it?

seanptmaher:
I'll give it a shot, await a PR soon :+1:

PR #371, opened 2020-06-11
Neon: implementation of mull

seanptmaher:
I didn't implement vmull_high, or vmull_n (or the combinations of the two) but those are moderately trivial to add later. 

(CC @ngzhian @tlively)


seanptmaher:
The build failures seem to be for my use of `(x?intxx_t)` casts, when compiling as C++, it wants me to use a `reinterpret_cast<>` or something. Should I `#ifdef __cplusplus`?

nemequ:
Casting from one integer type to another requires a static_cast not reinterpret_cast, but there is no need for an ifdef, that's what the `HEDLEY_STATIC_CAST`](https://github.com/nemequ/hedley/blob/master/hedley.h#L736-L743) macro is for.

There are also similar macros for reinterpret_cast and const_cast, plus a [`SIMDE_ALIGN_CAST`](https://github.com/nemequ/simde/blob/master/simde/simde-common.h#L233-L245) if you need to cast to a type with larger alignment requirements.

seanptmaher:
This should do it, thanks for the help.

seanptmaher:
Looks like AppVeyor is dead, looks like it can't install Ninja

nemequ:
Yeah, it will probably fix itself soon, don't worry about it... Until it's working I'll just ignore it.

seanptmaher:
Hm, I built this wrong for emscripten, I'll fix that. Sorry for the delay/bad code.

seanptmaher:
Turns out there's a mistake in `wasm_simd128.h` so it was missing some instructions. I'll try and remember to add the wasm implementation back in once it gets fixed.

nemequ:
> Turns out there's a mistake in `wasm_simd128.h` so it was missing some instructions. I'll try and remember to add the wasm implementation back in once it gets fixed.

Sorry about that.  I assume that you've reported it to emscripten or LLVM?  We tend to find quite a few compiler bugs and I like to make sure we report them when possible, especially to the open-source compilers.

The patch looks good, merging soon.

PR #373, opened 2020-06-12
neon: implementation of uzp{1,2}

seanptmaher:
Here are uzp1 and 2.

I don't have an arm64 machine I can actually run the native tests on to make sure that they pass [usually, I generate the tests on the native machine, instead of my x86 one], so I left the test generation code in there. If it passes CI, I'll clean it up. 

(cc @ngzhian @tlively)


nemequ:
No problem, when it's ready just request a review.

You may want to take a look at https://github.com/nemequ/simde/wiki/Development-Environment#emulation which describes how you can build and test on a Debian x86 machine using qemu.

PR #374, opened 2020-06-12
neon:Refactor pmin, padd with uzp{1,2}

seanptmaher:
I refactored pmin and padd here. 

I'm not 100% sure on how to package this up, I cherry-picked commits from padd and uzp{1,2} to do this. It might be best to wait until uzp{1,2} gets merged, and then I can just commit on those branches, but if you want to merge this in, that works too. 

It might also be worth just merging this PR, and ignoring the other two, because this packages all the changes into one, and stands on its own. 

(cc @tlively @ngzhian)


nemequ:
I merged the uzp stuff, then just rebased this and squashed so this patch had two commits; one to refactor pmin, the other to add padd (with the refactor padd patch squashed into that commit).

Excellent work on this; it looks great.  Barring any complications I should be able to merge it into master soon, it's working its way through CI now (slowly; Travis seems to be even more problematic than usual).

PR #375, opened 2020-06-12
avx2: added implementation for mm256_mul_ep{i,u}32

masterchef2209:
avx2: added implementation for mm256_mul_ep{i,u}32


test/x86/avx2.c
```diff
@@ -6930,6 +6930,92 @@ test_simde_mm256_movemask_epi8(void) {
   return 0;
 }
 
+static int
+test_simde_mm256_mul_epi32 (void) {
+  static const struct {
+    const int32_t a[8];
+    const int32_t b[8];
+    const int32_t r[8];
```
nemequ, line 6938: This should be 4 64-bit integers, not 8 32-bit.  If they are 32-bit integers this test fails on big-endian architectures like s390x.

masterchef2209: fixed this

PR #377, opened 2020-06-13
svml: implement {mm,mm256,mm512}_logb_{ps,pd}

himanshi18037:


nemequ:
I'm cherry picking the logb patch, I think you're still working on exp10 in simde-math.h, right?

himanshi18037:
> I'm cherry picking the logb patch, I think you're still working on exp10 in simde-math.h, right?

Yes, I am trying to figure something for exp10 functions.

PR #378, opened 2020-06-14
sse4.2: added implementation for mm_crc32_u{16,32,64}

masterchef2209:
sse4.2: added implementation for mm_crc32_u{16,32,64}

### DIFF
--- a/simde/x86/sse4.2.h
+++ b/simde/x86/sse4.2.h
@@ -290,6 +290,54 @@ simde_mm_crc32_u8(uint32_t prevcrc, uint8_t v) {
   #define _mm_crc32_u8(prevcrc, v) simde_mm_crc32_u8(prevcrc, v)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+uint32_t
+simde_mm_crc32_u16(uint32_t prevcrc, uint16_t v) {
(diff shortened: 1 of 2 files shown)


masterchef2209:
I tried implementing mm_crc32_u64 like (as crc is 64 bits here)
```cpp
SIMDE_FUNCTION_ATTRIBUTES
uint64_t
simde_mm_crc32_u64(uint64_t prevcrc, uint64_t v) {
  #if defined(SIMDE_X86_SSE4_2_NATIVE)
    return _mm_crc32_u64(prevcrc, v);
  #else
    uint64_t crc = prevcrc;
    for(int i = 0; i < 8 ; i++) {
      uint8_t v1 = ( v >> (i*8) ) & 0xff;
      for(int bit = 0 ; bit < 8 ; bit++) {
      if (crc & 1)
        crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
      else
        crc = (crc >> 1);
      }
    }
    return crc;
  #endif
}
#if defined(SIMDE_X86_SSE4_2_ENABLE_NATIVE_ALIASES)
  #define _mm_crc32_u64(prevcrc, v) simde_mm_crc32_u64(prevcrc, v)
#endif
```
this fails but if I use call to mm_crc32_u32 it passes native test, so it is a little unexpected result.

PR #380, opened 2020-06-15
svml: implement {mm,mm256,mm512}_{exp,epx2,exp10}_{ps,pd}

himanshi18037:


test/x86/svml.c
```diff
@@ -10648,6 +10648,1440 @@ test_simde_mm512_div_epu64(SIMDE_MUNIT_TEST_ARGS) {
   return 0;
 }
 
+static int
+test_simde_mm_exp_ps (void) {
```
nemequ, line 10652: This is the cause of the emscripten failures.  The functions should take `SIMDE_MUNIT_TEST_ARGS`, not `void`.  It was fixed in skel.c on Thursday, you probably just haven't updated the checkout you're using to generate these.

himanshi18037: Thanks! I have updated it.

PR #383, opened 2020-06-16
Implementation of cagt

seanptmaher:
Hello,

This one is also relatively short. Nice when I only have four types to implement. 

(CC @ngzhian @tlively)


simde/arm/neon/cagt.h
```diff
@@ -0,0 +1,140 @@
...
+    return vcagt_f32(a, b);
+  #else
+    simde_uint32x2_private r_;
+    simde_float32x2_private
+      a_ = simde_float32x2_to_private(a),
+      b_ = simde_float32x2_to_private(b);
+
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
+      r_.values[i] = (fabs(a_.values[i]) > fabs(b_.values[i])) ? 0xffffffff : 0;
```
nemequ, line 50: You probably already saw the double-promotion warnings on clang with -Weverything; those can be fixed by using `fabsf` instead of `fabs` for 32-bit floating point.

However, this should use [`simde_math_fabsf`](https://github.com/nemequ/simde/blob/b806f1ec04d6f15cc233b291ccb555dba3cce916/simde/simde-math.h#L473-L491), not just `fabsf`.  It's a little silly, but it lets us work on platforms without a libm (for example, the WASM support built into clang).

seanptmaher: How's this?

nemequ: That's the right idea, but you only replaced one call per line; the other is still fabs/fabsf.  clang-10 will probably fail again due to double-promotion, and although we don't have a test for it in CI it still won't without math.h.

seanptmaher: *Oops*

seanptmaher: Thank you for the... Patience.

nemequ: Don't worry about it; this is a very easy mistake to make.  Happens to all of us :)

The patch looks good, I'll merge it soon, thanks!

PR #388, opened 2020-06-18
neon: implementation of reinterpret

seanptmaher:
This is implemented via memcpy, which, honestly, is a pretty god-awful way of doing this, but _should_ be optimized out by the compiler. I had tried to get it working with just casts, but static cast doesn't work, reinterpret cast causes warnings, this was the . However, the way I wrote the file means that if
there's a way to do this better, all that I have to do is swap out three lines and it generates the code properly.

Thanks. 

(CC @ngzhian  @tlively)


seanptmaher:
Hm, I just realized, I forgot to add in the native aliases here. That will actually take awhile. I'll republish tomorrow.

test/arm/neon/reinterpret.c
```diff
@@ -0,0 +1,6760 @@
...
+
+  };
+
+  for (size_t i = 0 ; i < (sizeof(test_vec) / sizeof(test_vec[0])) ; i++) {
+    simde_int8x8_t a = simde_vld1_s8(test_vec[i].a);
+    simde_int8x8_private a_ = simde_int8x8_to_private(a);
+    simde_uint8x8_t r = simde_vreinterpret_u8_s8(a);
+    simde_uint8x8_private r_ = simde_uint8x8_to_private(r);
+
+    simde_assert_equal_i(0, simde_memcmp(&r_, &a_, sizeof(r_)));
```
seanptmaher, line 216: Any idea why there are `unused_variable` warnings here? I'm not getting them on my end despite compiling with `-Wall`, so I'm sorta confused. Plus, I'm taking the address of it here, so it should be getting used...

seanptmaher: (The warnings are on `a_`)

seanptmaher: Wait, nevermind, I think I've solved it. I'll push it tomorrow with the native aliases.

nemequ:
Yeah, memcpy is a good portable fallback, but it's really not as awful as you might think.  Casting is dangerous; you can run into issues with aliasing very easily.  We actually use memcpy to implement all the `*_to_private` / `*_from_private` calls because it's the only safe and portable way to do it without creating aliasing violations.

Earlier versions of SIMDe actually used to basically pass around what are now the `*_private` types.  For GCC, clang, and a few other compilers, we had the `__may_alias__` attribute, but in other places we didn't and ended up with lots of failures which.

It's also the only safe way to convert to a type with greater alignment requirements, but that shouldn't really apply here.

If a compiler can't optimize away a memcpy SIMDe will be terribly slow.  Those to/from_private functions are everywhere, especially the x86 APIs (since they rely heavily on aliasing).

While I don't have any problem with memcpy, using macros to implement these is a problem.  It doesn't allow us to add implementations for other architectures.  For example, on x86 these should really use cast functions (e.g., `vreinterpretq_f32_f64` -> `_mm_castpd_ps`) or, for integer to integer conversions, simply returning the integer.  On WASM, you should be able to always just return the input for the 128-bit vectors.  For AltiVec, casts are the official mechanism.

I don't see anything wrong with generating these functions; in fact, that seems like a good way to go.  But the code in SIMDe should look like the other functions.

As you can see from the Drone.io failures, there is also the problem that this generates extra functions to convent a type to itself (for example, a `vreinterpretq_u8_u8`).

simde/arm/neon/reinterpret.h
```diff
@@ -0,0 +1,335 @@
...
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(s32, f64, int32x4, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(s64, f64, int64x2, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(u8, f64, uint8x16, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(u16, f64, uint16x8, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(u32, f64, uint32x4, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(u64, f64, uint64x2, Element_Target_Return_Type_Long) \
+    SIMDE_ARM_NEON_A64V8_REINTERPRETQ_GEN(f32, f64, float32x4, Element_Target_Return_Type_Long)
+
+// These are spaced apart so I can more easily look at them with
+//  `gcc -E`, feel free the change the style.
```
nemequ, line 311: FWIW, I usually find it easiest to just pipe the output of `gcc -E` straight to `clang-format`.  Macros like this tend to get squashed into a single line of output, `clang-format` fixes that issue.

seanptmaher:
Alright, I rewrote that. Do you want me to put the script I used to generate them somewhere? It's a 50 line python script.

nemequ:
> Alright, I rewrote that.

Thanks!  Sorry, I know it's a bit of a hassle :(

This version looks good to me.  Assuming CI doesn't complain I'll merge it soon, thanks!

> Do you want me to put the script I used to generate them somewhere? It's a 50 line python script.

Nah, I don't think it's necessary.  If you want to keep it you could always put it in a gist and link to it from this issue, but I don't think there is any need to include it in the repo.

seanptmaher:
Here's the gist. I could put it in a comment, in the code, I guess, or just leave it here, but either way, it's not super important. 
https://gist.github.com/seanptmaher/560a696b990f504f04e4deba8d831dc5
What's the code coverage bot I failed?

PR #391, opened 2020-06-18
neon: add vshl and vshlq implementations

rosbif:
MSVC x86-64 doesn't seem to pick up some of the intrinsics involving MMX (__m64) registers.
I am not sure of the AltiVec (and even NEON) native version requirements.
Are they checked by the tests?
I have written some explanatory blurb at the start of shl.h.
I think it is useful but feel free to delete or modify it if you don't like it.


nemequ:
> MSVC x86-64 doesn't seem to pick up some of the intrinsics involving MMX (__m64) registers.

Yes, MSVC on x86_64 doesn't support MMX, or any of the SSE functions which operate on 64-bit vectors.  Basically, support for SSE+ doesn't necessarily imply support for MMX; you have to check `defined(SIMDE_X86_MMX_NATIVE)` separately.  It's a bit annoying, but TBH given all the problems with MMX I don't fault Microsoft for not supporting it on x86_64.

> I am not sure of the AltiVec (and even Neon) native version requirements.
> Are they checked by the tests?

NEON should be; IIRC we have armv7, armv8, and aarch64 builds on CI.  Luckily the Neon ones are pretty easy to check, since the [Neon Intrinsics Reference](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics) has the information (see the "Supported architectures" section for each function).

POWER is a different story; the documentation is pretty weak, and IIRC we only have a test for POWER8 on CI.  I'd like to add a few more POWER builds, but CI is already pretty slow :(.  I do already have POWER8 and POWER9 in wip/docker, I'll try ta add 5-7.  It won't run every commit, but I can try to test them somewhat regularly.

PR #394, opened 2020-06-19
neon: use AltiVec vec_sel in shl.h

rosbif:
Replace vec_or(vec_andc(a, c), vec_and(b, c)) with vec_sel(a, b, c).


rosbif:
Sorry, the commit title should cite shl.h not shl.c :-(

PR #399, opened 2020-06-19
neon: implementation of max

seanptmaher:
Here's an implementation of vmax. 

The tests build fine on Wasm, but I'm getting a test fail on one of the x86 intrinsics. 

```c
/x86/avx512f/mm512_cvtepi8_epi64/native/c[ OK    ] [ 0.00000000 / 0.00000000 CPU ]
../x86/avx512f.c:8678: assertion failed: r[3] != test_vec[i].r[3] (121 != 0)
```
This might be due to an outdated emscripten or v8, but again, it might not, I'm not sure. I'll update my build. 

(CC @ngzhian @tlively)


nemequ:
What compiler, version, target, and flags are you seeing that error on?  It looks like CI is successful, so I'll merge it, but I'd like to track down that failure if possible…

nemequ:
Fixed by f9772fd05bd994a5f646bc81ded2443c51c09d11, thanks!

PR #402, opened 2020-06-20
neon: implementation of zip{1,2,}

seanptmaher:
Hello, I implemented a few of the variations of zip. 

The first commit is just the other PR, the types. The zip1 and zip2 and zip commits should be packaged up and not actually be intertwined, I hope I didn't mess up the gitting. 

(CC @ngzhian @tlively)


seanptmaher:
Hm, travis and drone are failing, I thought I had fixed this issue, I don't think I committed it properly. 
Give me a sec.

nemequ:
It looks like ICC is still failing :(

I suspect another compiler bug; I'll try to get a reduced test case, which should tell us exactly what the issue is so we know what to fix, and give us something to send to ICC to report the bug.

seanptmaher:
It's odd, because it's only happening on the 'emulated' run, I don't know enough about what this implies, but that sounds kinda like a compiler bug to me (or, a weirdly specified compiler, I guess), seeing as it works 'natively'.

Either way, thanks for the help.

nemequ:
"emul" basically means `SIMDE_NO_NATIVE` is defined, which means it's using the portable fallbacks not an accelerated implementation (all the `SIMDE_*_NATIVE` macros are undefined).  I'm pretty sure it's a compiler bug; I actually started getting an ICE instead of just an incorrect result while trying to narrow down the issue, which I [reported](https://software.intel.com/en-us/forums/intel-c-compiler/topic/857088).

I'm not sure if that is a different manifestation of the same issue or not.  I'll try to take another look soonish, but I doubt I'll be able to get to it until later this week.

Unfortunately figuring out a way to work around this for ICC is going to be a pain, but necessary since we can't count on them fixing the underlying issue; in the past Intel hasn't been very good about fixing bugs I've reported in ICC; they usually just confirm that it's an issue and I never hear anything back.

ngzhian:
(Pinging since I clicked and saw that there was a response on the ICC bug, https://software.intel.com/en-us/forums/intel-c-compiler/topic/857088.)

It was fixed in newer version, suggest to run using
```
icc -o out -qnextgen foo.c
```

nemequ:
I should complain about how they don't fix bugs more often.

I think we can probably just require `-qnextgen` until the next ICC release, at least for people who want to use functions which use `…x2_t`.

nemequ:
Adding `-qnextgen` basically switches to a clang-based compiler, which explains why the code actually works.  `__INTEL_COMPILER` disappears, but the logic for clang should take over so that should be okay.

Unfortunately, after playing with it a bit, the compiler seems to be pretty far from ready with lots of pretty obvious bugs, and CMake doesn't work anymore.

I think I'm just going to but a bunch of ifdefs in the code so that these types, and any functions that use them, simply aren't available in ICC (the current version, -qnextgen should still work since `__INTEL_COMPILER` isn't defined).

PR #404, opened 2020-06-20
sse4.1: add NEON and AltiVec translations of *blend*; correct PPC versions globally

rosbif:
With PPC64 on qemu there is a problem with _mm_movemask_epi8. I think it is a qemu bug. Have you seen it before?

### DIFF
--- a/simde/arm/neon/add.h
+++ b/simde/arm/neon/add.h
@@ -364,7 +364,7 @@ simde_vaddq_f64(simde_float64x2_t a, simde_float64x2_t b) {
     return vaddq_f64(a, b);
   #elif defined(SIMDE_X86_SSE2_NATIVE)
     return _mm_add_pd(a, b);
-  #elif defined(SIMDE_POWER_ALTIVEC_P5_NATIVE)
+  #elif defined(SIMDE_POWER_ALTIVEC_P7_NATIVE)
     return vec_add(a, b);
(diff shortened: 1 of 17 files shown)


rosbif:
I think I have now finished the NEON and AltiVec translations of the blend and blendv intrinsics.
That said, I just thought that maybe there are also some in AVX512*.
I shall check later.

mr-c:
@rosbif Can you fix the merge conflict? That is stopping the CI from running. Thanks!

rosbif:
@mr-c 
> @rosbif Can you fix the merge conflict? That is stopping the CI from running. Thanks!
>

I shall try but I am never sure of the best way to go about it, especially if you have changed my commits. Often I end up with the same modifications twice, once in my commit and once in yours.
Can you tell me the best way to go about it?

rosbif:
I have roughly merged them but I have a problem.
I think Evan removed the SIMDE_X86_GENERATE_CONVERSION_FUNCTION generation for float and double because that is what caused the gcc ICE.
I thought he had replaced them by hand-coded functions.
But e.g. simde__m128d_to_altivec_f64 is missing so the code will no longer compile.

rosbif:
I have pushed a merged version which is close to the one in master but with my PPC version corrections. It should be easier to merge.

N.B. I had to reintroduce the SIMDE_X86_GENERATE_CONVERSION_FUNCTION generation for float and double otherwise it won't compile even on clang.

PR #406, opened 2020-06-20
avx2: added mm256_maddubs_epi16, mm256_h{add,sub}s_epi16, mm256_slli_si256

masterchef2209:
avx2: added impl for mm256_maddubs_epi16
avx2: added the impl for mm256_h{add,sub}s_epi16
avx2: added the impl for mm256_slli_si256


nemequ:
For the pairwise functions, there is a much more efficient way to do this.

Basically, what you want to do is shuffle the inputs so you have one vector composed of the odd elements of both input vectors, and another vector composed of the even elements.  Then just add those together.

Many ISA extensions include a very efficient way to do the shuffle.  For NEON there are `vuzp1` and `vuzp2` functions, on a lot of architectures a generic shuffle like `_mm_shuffle_epi8` will yield good performance.

There are some examples in SSSE3, such as `simde_mm_hadd_epi16`.  There are also the uzp1/uzp2 function in NEON.  

Having those uzp1/uzp2 functions available in NEON is really nice for implementing all the pairwise functions; it is a central place to keep all the optimizations for different architectures, and it makes implementing the pariwise functions pretty trivial (see arm/neon/padd.h for an example).  How about doing something similar for x86?  Intel doesn't have dedicated functions for this, but we could do something like `simde_x_mm_deinterleaveeven_*` / `simde_x_mm_deinterleaveodd_*`.

Right now the logic is duplicated and scattered around lots of different functions, this would make the code a lot cleaner and make it easier to ensure we're generating the fastest code we can.  It looks like the earliest pairwise function is in SSE3, so that's probably a sensible place to keep them.

Is that something you'd be interested in putting together?  It seems more fun that just churning out function after function (which is important, but I know it gets a bit boring).

masterchef2209:
Sure, sounds interesting to me. I will look into it.

masterchef2209:
Closing this as I have created separate PR for mm256_maddubs_epi16 and other ones are already merged.

PR #409, opened 2020-06-21
neon: correct some #undef in native aliases

rosbif:
Also correct two POWER_ALTIVEC_P7 to P8 in sub.h


nemequ:
Good catch, thanks!

PR #410, opened 2020-06-22
ssse3: added impl for x_mm_deinterleave{even,odd}_{epi{16,32},ps,pd}

masterchef2209:
Please review this, have I missed something? And what to do for 256i variants , because it is not exactly deinterleaving operation for functions like `mm256_hadd_epi16` ?

### DIFF
--- a/simde/x86/ssse3.h
+++ b/simde/x86/ssse3.h
@@ -35,6 +35,174 @@ SIMDE_BEGIN_DECLS_
 
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128i
+simde_x_mm_deinterleaveeven_epi16 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
+    a_ = simde__m128i_to_private(a),
+    b_ = simde__m128i_to_private(b);
+
+  #if defined(SIMDE_SHUFFLE_VECTOR_)
+    r_.i16 = SIMDE_SHUFFLE_VECTOR_(16, 16, a_.i16, b_.i16, 0, 2, 4, 6, 8, 10, 12, 14);
+  #else
+  const size_t halfway_point = (sizeof(r_.i16) / sizeof(r_.i16[0])) / 2;
+  for(size_t i = 0 ; i < halfway_point ; i++) {
+    r_.i16[i] = a_.i16[2 * i];
+    r_.i16[i + halfway_point] = b_.i16[2 * i];
+  }
+  #endif
+
+  return simde__m128i_from_private(r_);
+}
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_x_mm_deinterleaveodd_epi16 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
+    a_ = simde__m128i_to_private(a),
+    b_ = simde__m128i_to_private(b);
+
+  #if defined(SIMDE_SHUFFLE_VECTOR_)
+    r_.i16 = SIMDE_SHUFFLE_VECTOR_(16, 16, a_.i16, b_.i16, 1, 3, 5, 7, 9, 11, 13, 15);
+  #else
+  const size_t halfway_point = (sizeof(r_.i16) / sizeof(r_.i16[0])) / 2;
+  for(size_t i = 0 ; i < halfway_point ; i++) {
+    r_.i16[i] = a_.i16[2 * i + 1];
+    r_.i16[i + halfway_point] = b_.i16[2 * i + 1];
+  }
+  #endif
+
+  return simde__m128i_from_private(r_);
+}
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_x_mm_deinterleaveeven_epi32 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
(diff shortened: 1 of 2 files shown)


simde/x86/ssse3.h
```diff
@@ -33,6 +33,158 @@ HEDLEY_DIAGNOSTIC_PUSH
...
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_x_mm_deinterleaveeven_epi16 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
+    a_ = simde__m128i_to_private(a),
+    b_ = simde__m128i_to_private(b);
+  #if defined(SIMDE_ASSUME_VECTORIZATION) && defined(SIMDE_SHUFFLE_VECTOR_)
+    r_.i16 = SIMDE_SHUFFLE_VECTOR_(16, 16, a_.i16, b_.i16, 0, 2, 4, 6, 8, 10, 12, 14);
+  #endif
```
nemequ, line 45: This should be an `#else` (and of course you'll need an `#endif` after the loop).  If we have `__builtin_shuffle` or `__builtin_shufflevector` (which is what `SIMDE_SHUFFLE_VECTOR_` maps to) then that one call does everything we need for this function.  In this case, the loop just does it again, but likely slower.

simde/x86/ssse3.h
```diff
@@ -33,6 +33,158 @@ HEDLEY_DIAGNOSTIC_PUSH
...
 SIMDE_BEGIN_DECLS_
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_x_mm_deinterleaveeven_epi16 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
+    a_ = simde__m128i_to_private(a),
+    b_ = simde__m128i_to_private(b);
+  #if defined(SIMDE_ASSUME_VECTORIZATION) && defined(SIMDE_SHUFFLE_VECTOR_)
```
nemequ, line 43: We don't need the `SIMDE_ASSUME_VECTORIZATION` condition here.  If you've called this function you've already made a determination about your implementation strategy, and shuffling isn't going to perform any worse than the loop here (it will likely perform much better).

simde/x86/ssse3.h
```diff
@@ -33,6 +33,158 @@ HEDLEY_DIAGNOSTIC_PUSH
 SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
 SIMDE_BEGIN_DECLS_
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_x_mm_deinterleaveeven_epi16 (simde__m128i a, simde__m128i b) {
+  simde__m128i_private
+    r_,
+    a_ = simde__m128i_to_private(a),
+    b_ = simde__m128i_to_private(b);
```
nemequ, line 42: Minor style issue: there should be a blank line after this to separate the declarations from the code.

simde/x86/ssse3.h
masterchef2209: oops, fixed this

masterchef2209:
> A few minor corrections are necessary, and after that I think it would be good to add at least a NEON implementation (using uzp1/uzp2, see `_mm_hsub_ps` for an example).

Do we need to have NEON impl (uzp1/uzp2 fallback) inside these functions or this is about using uzp1/uzp2 for other functions.

nemequ:
> And what to do for 256i variants , because it is not exactly deinterleaving operation for functions like `mm256_hadd_epi16` ?

`_mm256_hadd_epi16` is a perfect candidate for this; it's actually a bit simpler than `_mm_addubs_epi16`.  `_mm_maddubs_epi16` first performs a widening multiply on the input vectors (of 16 8-bit unsigned integers) and stores it in a temporary variable, then does the pairwise operation (saturating 16-bit addition) on that. `_mm256_hadd_epi16` skips that first step, then uses normal (not saturating) addition.

nemequ:
> Do we need to have NEON impl (uzp1/uzp2 fallback) inside these functions or this is about using uzp1/uzp2 for other functions?

The former.  Implementing this function using uzp1/uzp2 if they are available.

simde/x86/ssse3.h
```diff
@@ -33,6 +33,174 @@ HEDLEY_DIAGNOSTIC_PUSH
...
+
+  #if defined(SIMDE_SHUFFLE_VECTOR_)
+    r_.i16 = SIMDE_SHUFFLE_VECTOR_(16, 16, a_.i16, b_.i16, 0, 2, 4, 6, 8, 10, 12, 14);
+  #else
+  const size_t halfway_point = (sizeof(r_.i16) / sizeof(r_.i16[0])) / 2;
```
nemequ, line 47: This should be indented since it's inside of the else condition now.

nemequ:
Sorry, I just noticed; this should be in SSE3 not SSSE3.  That way functions like `simde_mm_hsub_ps` can take advantage.

masterchef2209:
> Sorry, I just noticed; this should be in SSE3 not SSSE3. That way functions like `simde_mm_hsub_ps` can take advantage.

It is definitely my mistake, I read it wrong, sorry.

masterchef2209:
Closing this as I have created another PR with all the suggested changes.

PR #411, opened 2020-06-22
sse2: correct simde_mm_bslli_si128 for big endian PPC64

rosbif:
Also simplify PPC64 little endian version and add missing simde_assert_m128i_i8 test.


rosbif:
The Travis CI failure isn't my fault.
It passed the first time after I pushed my modifications.

nemequ:
I have a feeling there are a lot of PP64BE issues… I need to figure out a decent way to test it ☹

rosbif:
> I have a feeling there are a lot of PP64BE issues… I need to figure out a decent way to test it frowning_face
>

Yes I ran into two more which (I hope) I just corrected. It took me a while to find them but the good news is that I don't think there are any more. Everything passes on -mcpu=power6, 7 and 8 here.
For me cross-compiling and executing with qemu work fine. I deliberately chose big endian PPC64 because I'm a masochist ;-)  No, seriously I needed to test Altivec and I also wanted a big endian target to catch other bugs. I couldn't get AltiVec to work on s390x so I installed PPC64.

After these adjustments from P5 to P7, AFAICS all the remaining P5 should be P6.
But this is a big change and I am wondering if it is worth the effort.

Is the POWER9 machine you spoke about BE or LE?

Oops the ppc64le machine is now failing on the out of range test r = simde_mm_bslli_si128(a, 19);
This is probably normal but I am wondering why it didn't fail before. I may have done something stupid. I'll look in the morning. I think I know : I removed an `& 15 `which seemed unnecessary because of the `>= 16` test but the compiler doesn't realize that it won't be executed. There will be the same problem in bsrlli. I think I shall change the `vec_sld` to `vec_slo`. It is getting late so I'll push them untested with crossed fingers.

rosbif:
I hope that I have now finished the PPC64BE corrections.

nemequ:
> Is the POWER9 machine you spoke about BE or LE?

The OS on there now is LE, but IIRC I could put a BE OS on there.

By understanding is that since POWER8 IBM has really been pushing people to use LE, and BE is mainly a compatibility thing.  At least that's what someone from IBM led me to believe back when P8 was new.

AFAIK s390x is the last real BE platform remaining.

nemequ:
Merged, thanks!

PR #413, opened 2020-06-23
sse3: added impl for x_mm_deinterleave{even,odd}_{epi{16,32},ps,pd}

masterchef2209:
sse3: added impl for x_mm_deinterleave{even,odd}_{epi{16,32},ps,pd}

### DIFF
--- a/simde/x86/avx2.h
+++ b/simde/x86/avx2.h
@@ -1568,26 +1568,17 @@ simde_mm256_hadd_epi16 (simde__m256i a, simde__m256i b) {
 #if defined(SIMDE_X86_AVX2_NATIVE)
   return _mm256_hadd_epi16(a, b);
 #else
-  simde__m256i_private r_;
-  simde__m256i_private a_ = simde__m256i_to_private(a);
-  simde__m256i_private b_ = simde__m256i_to_private(b);
+  simde__m256i_private
+  r_,
+  a_ = simde__m256i_to_private(a),
+  b_ = simde__m256i_to_private(b);
 
-  r_.i16[0] = a_.i16[1] + a_.i16[0];
-  r_.i16[1] = a_.i16[3] + a_.i16[2];
-  r_.i16[2] = a_.i16[5] + a_.i16[4];
-  r_.i16[3] = a_.i16[7] + a_.i16[6];
-  r_.i16[4] = b_.i16[1] + b_.i16[0];
-  r_.i16[5] = b_.i16[3] + b_.i16[2];
-  r_.i16[6] = b_.i16[5] + b_.i16[4];
-  r_.i16[7] = b_.i16[7] + b_.i16[6];
-  r_.i16[8] = a_.i16[9] + a_.i16[8];
-  r_.i16[9] = a_.i16[11] + a_.i16[10];
-  r_.i16[10] = a_.i16[13] + a_.i16[12];
-  r_.i16[11] = a_.i16[15] + a_.i16[14];
-  r_.i16[12] = b_.i16[9] + b_.i16[8];
-  r_.i16[13] = b_.i16[11] + b_.i16[10];
-  r_.i16[14] = b_.i16[13] + b_.i16[12];
-  r_.i16[15] = b_.i16[15] + b_.i16[14];
+  SIMDE_VECTORIZE
+  for (size_t  i = 0 ; i < (sizeof(r_.m128i) / sizeof(r_.m128i[0])) ; i++ ) {
+    const simde__m128i even = simde_x_mm_deinterleaveeven_epi16(a_.m128i[i], b_.m128i[i]);
+    const simde__m128i odd = simde_x_mm_deinterleaveodd_epi16(a_.m128i[i], b_.m128i[i]);
+    r_.m128i[i] = simde_mm_add_epi16(even, odd);
+  }
 
   return simde__m256i_from_private(r_);
 #endif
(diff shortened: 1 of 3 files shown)


masterchef2209:
I have also refactored code for mm256_hadd_epi16 to use x_mm_deinterleave* operations. Please review it, if it is fine, I will refactor other such functions as well.

nemequ:
> I have refactored code for mm256_hadd_epi16 to use x_mm_deinterleave* operations. Please review it, if it is fine, I will refactor other such functions as well.

The idea is sound, but instead of extracting the individual 128-bit vectors it would be much better to create `simde_mm256_deinterleave{even,odd}_*` functions which operate on the entire 256-bit vector.  That will be a lot faster on hardware that supports 256-bit vectors.

Then the code you have here could basically be the fallback for when 256-bit vectors aren't natively supported but 128-bit vectors are.

Let me clean up #275 a bit and merge it, that would be a good foundation for this.  Meanwhile why don't you start converting the 128-bit horizontal operations (`_mm_h{add,sub}_{ps,pd,epi16,epi32}` and `_mm_h{add,sub}s_epi16` to use this?

simde/x86/avx2.h
```diff
@@ -1569,31 +1569,15 @@ simde_mm256_hadd_epi16 (simde__m256i a, simde__m256i b) {
...
-  r_.i16[15] = b_.i16[15] + b_.i16[14];
+  r_.m128i[0] = simde_mm_add_epi16(simde_x_mm_deinterleaveeven_epi16(simde_mm256_extracti128_si256(a,0),simde_mm256_extracti128_si256(b,0)),simde_x_mm_deinterleaveodd_epi16(simde_mm256_extracti128_si256(a,0),simde_mm256_extracti128_si256(b,0)));
+  r_.m128i[1] = simde_mm_add_epi16(simde_x_mm_deinterleaveeven_epi16(simde_mm256_extracti128_si256(a,1),simde_mm256_extracti128_si256(b,1)),simde_x_mm_deinterleaveodd_epi16(simde_mm256_extracti128_si256(a,1),simde_mm256_extracti128_si256(b,1)));
 
   return simde__m256i_from_private(r_);
 #endif
 }
 #if defined(SIMDE_X86_AVX2_ENABLE_NATIVE_ALIASES)
-  #undef _mm256_hadd_epi16
+  #undef _mm256_hadd_epi16hadd
```
nemequ, line 1580: ?

simde/x86/avx2.h
```diff
@@ -1569,31 +1569,15 @@ simde_mm256_hadd_epi16 (simde__m256i a, simde__m256i b) {
...
-  r_.i16[8] = a_.i16[9] + a_.i16[8];
-  r_.i16[9] = a_.i16[11] + a_.i16[10];
-  r_.i16[10] = a_.i16[13] + a_.i16[12];
-  r_.i16[11] = a_.i16[15] + a_.i16[14];
-  r_.i16[12] = b_.i16[9] + b_.i16[8];
-  r_.i16[13] = b_.i16[11] + b_.i16[10];
-  r_.i16[14] = b_.i16[13] + b_.i16[12];
-  r_.i16[15] = b_.i16[15] + b_.i16[14];
+  r_.m128i[0] = simde_mm_add_epi16(simde_x_mm_deinterleaveeven_epi16(simde_mm256_extracti128_si256(a,0),simde_mm256_extracti128_si256(b,0)),simde_x_mm_deinterleaveodd_epi16(simde_mm256_extracti128_si256(a,0),simde_mm256_extracti128_si256(b,0)));
+  r_.m128i[1] = simde_mm_add_epi16(simde_x_mm_deinterleaveeven_epi16(simde_mm256_extracti128_si256(a,1),simde_mm256_extracti128_si256(b,1)),simde_x_mm_deinterleaveodd_epi16(simde_mm256_extracti128_si256(a,1),simde_mm256_extracti128_si256(b,1)));
```
nemequ, line 1574: These are really a lot to cram into a single line.  You should break it up.

You can use the private members of `a_` and `b_` (which you would need to add back in) instead of `simde_mm256_extracti128_si256`.  Something like (completely untested, for illustrative purposes only):

```c
simde__m256i_private
  r_,
  a_ = simde__m256i_to_private(a),
  b_ = simde__m256i_to_private(b);

SIMDE_VECTORIZE
for (size_t = i = 0 ; i < (sizeof(r_.m128i) / sizeof(r_.m128i[0])) ; i++ ) {
  const simde__m128i even = simde_x_mm_deinterleaveeven_epi16(a_.m128i[i], b_.m128i[i]);
  const simde__m128i odd = simde_x_mm_deinterleaveodd_epi16(a_.m128i[i], b_.m128i[i]);
  r_.m128i[i] = simde_mm_add_epi16(even, odd);
}
```

simde/x86/avx2.h
masterchef2209: I will fix this, this has happened because I was trying to search the function in the file and didn't press "Ctrl+F" properly, and I didn't notice this.

nemequ: Np.  I have similar issues because I start typing in the wrong window.  You can usually catch stuff like this with a quick `git diff` before the committing, but stuff sometimes slips through.  It's nice to have a code review process to catch stuff like this.

masterchef2209:
I will complete this and other similar functions, after I finish implementing, `simde_mm256_deinterleave{even,odd}_*`. I have made all the suggested changes so far.

PR #414, opened 2020-06-23
neon: Add implementation of addl{,_high}

seanptmaher:
One thing to note was that in the portable fallback loop, I had to cast each integer to its larger counterpart to stop truncation. I'm
unsure if there's a faster way to do this, so if there is that you can think of off the top of your head, let me know.

Otherwise, should be pretty good. 

(CC @tlively @ngzhian)


simde/arm/neon/addl.h
```diff
@@ -0,0 +1,335 @@
...
+    return vaddl_high_u32(a, b);
+  #else
+    simde_uint64x2_private r_;
+    simde_uint32x4_private
+      a_ = simde_uint32x4_to_private(a),
+      b_ = simde_uint32x4_to_private(b);
+
+    size_t halfway = (sizeof(a_.values) / sizeof(a_.values[0])) / 2;
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
```
ngzhian, line 319: Is there a `arraylen` macro or something for this?

seanptmaher: Hm, I don't think there is, but adding it would be trivial, and would actually simplify a lot of the code. @nemequ want me to add it in?

nemequ: I just opened up #421 to discuss this.

nemequ:
> One thing to note was that in the portable fallback loop, I had to cast each integer to its larger counterpart to stop truncation. I'm
> unsure if there's a faster way to do this, so if there is that you can think of off the top of your head, let me know.

There is the [`__builtin_convertvector`](https://clang.llvm.org/docs/LanguageExtensions.html#langext-builtin-convertvector) function, which SIMDe has a `SIMDE_CONVERT_VECTOR_` macro for.  It's generally a bit tricky when different sizes are used; the output has as many elements as the input, so if you're converting from `int8_t` to `int16_t` you're doubling the size of the vector.  Thankfully these functions take 64-bit inputs and return 128-bit outputs, so it works well.

It's not available in all compilers (IIRC really just GCC, clang, and clang-derived compilers), so you'll still need the code you currently have as a fallback, but for compilers which support `__builtin_convertvector` the result should be optimal.

Basically, the implementation would look something like

```c
#if defined(SIMDE_CONVERT_VECTOR_)
  __typeof__(r_.values) A, B;
  SIMDE_CONVERT_VECTOR_(A, a_.values);
  SIMDE_CONVERT_VECTOR_(B, b_.values);
  r_.values = A + B;
#else
  SIMDE_VECTORIZE
    r_.values[i] = HEDLEY_STATIC_CAST(int16_t, a_.values[i]) + HEDLEY_STATIC_CAST(int16_t, b_.values[i]);
  }
#endif
```

In this case, though, I'd suggest just implementing the `vmovl` functions first, then you can use those to implement these.  For example:

```
SIMDE_FUNCTION_ATTRIBUTES
simde_int16x8_t
simde_vaddl_s8(simde_int8x8_t a, simde_int8x8_t b) {
    return vaddl_s8(a, b);
  #else
    return simde_vaddq_s16(simde_vmovl_s8(a), simde_vmovl_s8(b));
  #endif
}
```

ISA extensions which support widening addition could still provide optimized versions in `simde_vaddl_s8`, but you can share the logic for just the widening (vmull needs to do the same thing, for example), which is more likely to have analogs in other ISA extensions.

seanptmaher:
Thanks for the comments. A lot of the PRs up right now rely on that, so I'll work that into the implementations.

nemequ:
> Thanks for the comments. A lot of the PRs up right now rely on that, so I'll work that into the implementations.

You're talking about the movl stuff, right?  I went ahead merged that yesterday for that reason.  I did include__builtin_convertvector implementations for vmovl, but not vmovl_high.  I'll add those in a bit; the implementations could really just be `vmovl_s32(vget_high_s32(a))`.

If you update this to use vmovl I'll go ahead and merge it.

Sorry about all these change requests; I'd just really like to reuse as much code as possible in the NEON implementations… that should make it *much* easier to maintain than the x86 stuff, which is a mess (largely because x86 just has tons of gaps in what it supports so doing stuff like this is much harder).

seanptmaher:
Alright, it uses movl now.

PR #415, opened 2020-06-23
svml: implement {mm,mm256}_erfinv_{ps,pd}

himanshi18037:


simde/x86/svml.h
```diff
@@ -4370,6 +4370,138 @@ simde_mm_erfinv_ps (simde__m128 a) {
   #define _mm_erfinv_ps(a) simde_mm_erfinv_ps(a)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128d
+simde_mm_erfinv_pd (simde__m128d a) {
+  #if defined(SIMDE_X86_SVML_NATIVE) && defined(SIMDE_X86_SSE_NATIVE)
+    return _mm_erfinv_pd(a);
+  #elif defined(SIMDE_ASSUME_VECTORIZATION)
+    simde__m128d one = simde_mm_set1_pd(1.0f);
```
nemequ, line 4379: These should be `SIMDE_FLOA64_C(1.0)`; 1.0f is a 32-bit float.  The rest looks good to me, so I'll just tweak this when I merge it.

simde/x86/avx.h
```diff
@@ -2797,6 +2797,48 @@ simde_mm256_cmp_ps (simde__m256 a, simde__m256 b, const int imm8)
...
+    dest_ = simde__m256_to_private(dest),
+    src_ = simde__m256_to_private(src);
+
+  #if defined(simde_math_copysignf)
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.f32) / sizeof(r_.f32[0])) ; i++) {
+      r_.f32[i] = simde_math_copysignf(dest_.f32[i], src_.f32[i]);
+    }
+  #else
+    simde__m256 sgnbit = simde_mm256_xor_ps(simde_mm256_set1_ps(SIMDE_FLOAT32_C(0.0)), simde_mm256_set1_ps(-SIMDE_FLOAT32_C(0.0)));
```
nemequ, line 2814: This is such an ugly hack, I hate to see it propogated ☹.  Luckily it shouldn't be reached very often, since `simde_math_copysignf` will almost certainly be defined…

himanshi18037: Yes in most of the cases it would not matter, but I don't see any other workaround for this.

PR #416, opened 2020-06-23
neon: implementation of paddl

seanptmaher:
I decided to copy-paste the implementation of addl and adjust it, because using uzp{1,2} wasn't feasible (as there is only one vector argument)

Otherwise, should be good.

(CC @tlively @ngzhian)


nemequ:
How about something like (untested)

```c
simde_int16x8_t tmp = simde_vmovl_s8(a);
return simde_vpadd_s16(simde_vget_low_s16(tmp), simde_vget_high_s16(tmp));
```

?

seanptmaher:
Actually, that's a good pattern, I didn't know there was an intrinsic that does what `vmovl` does, thanks! I'll see if I can get that swapped over

nemequ:
> EDIT: Oh, because movl isn't implemented. I'll see what I can do in a sec.

Yes, sorry, I should have mentioned that.  FWIW the implementation should be pretty straightforward; `SIMDE_CONVERT_VECTOR_(r_.val, a_.val)` if `SIMDE_CONVERT_VECTOR_` is defined, otherwise just a loop with a static cast.

seanptmaher:
Hm, I didn't realize that `SIMDE_CONVERT_VECTOR_` existed, so I just used the static cast.

Also, movl isn't defined for the right sizes to use it for `paddlq`, so I didn't. Is there another way you want me to do it? I could probably use the privates and some memcpy magic but it probably would not be cleaner than this at all.

nemequ:
> Also, movl isn't defined for the right sizes to use it for `paddlq`, so I didn't. Is there another way you want me to do it? I could probably use the privates and some memcpy magic but it probably would not be cleaner than this at all.

How about something like

```c
int16x8_t lo = simde_vmovl_s8(vget_low_s8(a));
int16x8_t hi = simde_vmovl_s8(vget_high_s8(a));
return vpaddq_s16(lo, hi);
```

?

nemequ:
Looks great, thanks!

PR #417, opened 2020-06-23
sse3,ssse3: refactored mm_h{add,sub}{,s}_{ps,pd,epi16,epi32}

masterchef2209:


nemequ:
This looks great, thanks!  I love seeing all that code removed ☺

PR #418, opened 2020-06-23
neon: Add implementation of uzp

seanptmaher:
This PR is mostly adapted from zip, I noticed uzp, which uses the x2 vectors, also wasn't implemented. I'm pretty sure ICC will have the same problems, but it should work on other platforms. 

(CC @tlively @ngzhian)


nemequ:
Thanks!  Yep, you called it, problems for ICC :/

I'll go ahead and do the same thing here; just ifdef them out of existence in ICC.

PR #420, opened 2020-06-23
neon: implementation of trn

seanptmaher:
Nothing special here, aside from maybe not working on ICC due to the same bug.

(CC @tlively @ngzhian)


simde/arm/neon/trn1.h
```diff
@@ -0,0 +1,500 @@
...
+  #else
+    simde_float32x2_private
+      r_,
+      a_ = simde_float32x2_to_private(a),
+      b_ = simde_float32x2_to_private(b);
+
+    const size_t halfway_point = sizeof(r_.values) / sizeof(r_.values[0]) / 2;
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < halfway_point ; i++) {
+    const size_t idx = i << 1;
```
ngzhian, line 51: formatting is off

ngzhian: and likewise for the other implementations

seanptmaher: Should be good now

PR #424, opened 2020-06-24
svml: implement mm512_{erfinv,floor,round}_{ps,pd}  avx512dq: implement mm512_xor_{ps,pd}

himanshi18037:


simde/x86/svml.h
```diff
@@ -6718,6 +6718,302 @@ simde_mm512_mask_ceil_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
   #define _mm512_mask_ceil_pd(src, k, a) simde_mm512_mask_ceil_pd(src, k, a)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128
+simde_mm_svml_floor_ps (simde__m128 a) {
+  #if defined(SIMDE_X86_SVML_NATIVE) && defined(SIMDE_X86_SSE_NATIVE)
+    return _mm_svml_floor_ps(a);
+  #else
+    simde__m128_private
```
nemequ, line 6727: It looks like these are basically just duplicates of the SSE 4.1 rounding functions, except they only require SSE.  What they require doesn't really matter to us since we already have portable versions to fall back on.

You could greatly simplify the implementation (and improve performance) my just using them.  For example, this whole function could just be

```c
SIMDE_FUNCTION_ATTRIBUTES
simde__m128
simde_mm_svml_floor_ps (simde__m128 a) {
  #if defined(SIMDE_X86_SVML_NATIVE) && defined(SIMDE_X86_SSE_NATIVE)
    return _mm_svml_floor_ps(a);
  #else
    return simde_mm_round_ps(a, _MM_FROUND_TO_NEG_INF);
  #endif 
}
```

PR #425, opened 2020-06-24
neon: Fix include guards

seanptmaher:
Hi, this is just me cleaning up after doing a s// wrong in my code, sorry about that. It caused a compile error when including paddl.h (I think because of the way the headers are set up it doesn't happen in any other of the files right now, but it could in the future) from other places in the code that didn't already include "add.h" (I just spent longer than I care to admit finding why it wasn't including the file, I've never had to run into this problem before :^) ) 

(CC @ngzhian @tlively)

### DIFF
--- a/simde/arm/neon/addw.h
+++ b/simde/arm/neon/addw.h
@@ -25,8 +25,8 @@
  *   2020      Sean Maher <seanptmaher@gmail.com>
  */
 
-#if !defined(SIMDE_ARM_NEON_ADD_H)
-#define SIMDE_ARM_NEON_ADD_H
+#if !defined(SIMDE_ARM_NEON_ADDW_H)
+#define SIMDE_ARM_NEON_ADDW_H
 
 #include "types.h"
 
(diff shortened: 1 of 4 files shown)


nemequ:
I noticed one of these yesterday, too.  I think I'll add a formatting test to try to check for this, it's a really easy mistake to make.

rosbif:
> I noticed one of these yesterday, too. I think I'll add a formatting test to try to check for this, it's a really easy mistake to make.
>

As each test is run in C and in C++, an easy way to detect this would be for one to include e.g. max.h and for the other to include neon.h.
If two tests have the same include guard, one of the two would fail when including neon.h.
This would also pick up the case I mentioned earlier when one forgets to include e.g. max.h in neon.h.
I know you added a formatting test for this but, as I also pointed out, this happens after a PR rather than in local tests causing wasted time and wasted CI resources.

nemequ:
That's a good idea. The additional complexity is only in the tests not the headers, which is fine with me.

The other thing I really need to do is move the formatting checks (which aren't really just for formatting anymore) to a script which can be run locally, perhaps even as a pre-commit hook…

PR #427, opened 2020-06-24
neon: Implementation of movl{,_high}

seanptmaher:
Hello, here's an implementation of movl. 

(CC @tlively @ngzhian)


simde/arm/neon/movl_high.h
```diff
@@ -0,0 +1,178 @@
...
+  #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
+    return vmovl_high_s16(a);
+  #else
+    simde_int32x4_private r_;
+    simde_int16x8_private a_ = simde_int16x8_to_private(a);
+
+    size_t half = (sizeof(a_.values) / sizeof(a_.values[0])) / 2;
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
+      r_.values[i] = HEDLEY_STATIC_CAST(int16_t, a_.values[i + half]);
```
nemequ, line 72: This should cast to `int32_t`.  Same problem in the several other functions in this file.

This in why the aarch64 build on Travis is failing; when it calls the native functions the results are actually correct, but they don't match the test vectors.

seanptmaher: Oh, oops. Thanks,

PR #431, opened 2020-06-25
avx2: added mm256_sra_epi{16,32}, mm{,256}_srav_epi32

masterchef2209:
avx2: added impl for mm256_sra_epi{16,32}
avx2: added impl for mm{,256}_srav_epi32


simde/x86/avx2.h
```diff
@@ -2842,6 +2842,80 @@ simde_mm256_sllv_epi64 (simde__m256i a, simde__m256i b) {
...
+  #else
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.i16) / sizeof(r_.i16[0])) ; i++) {
+      r_.i16[i] = a_.i16[i] >> shift;
+    }
+  #endif
+
+  return simde__m256i_from_private(r_);
+}
+#if defined(SIMDE_X86_AVX2_NATIVE)
```
nemequ, line 2869: These should go in the functions, instead of being implemented as macros.  The only time functions like this should be implemented as macros is when one of the arguments must be an immediate-mode constant (in which case the compiler will emit an error if we try to use it with a "normal" variable in a function, often even if that value is still always known at compile time).

simde/x86/avx2.h
```diff
@@ -2842,6 +2842,80 @@ simde_mm256_sllv_epi64 (simde__m256i a, simde__m256i b) {
...
+    for (size_t i = 0 ; i < (sizeof(r_.i32) / sizeof(r_.i32[0])) ; i++) {
+      r_.i32[i] = a_.i32[i] >> shift;
+    }
+  #endif
+
+  return simde__m256i_from_private(r_);
+}
+#if defined(SIMDE_X86_AVX2_NATIVE)
+#  define simde_mm256_sra_epi32(a, count) _mm256_sra_epi32(a, count)
+#elif defined(SIMDE_X86_SSE2_NATIVE)
```
masterchef2209, line 2908: @nemequ SSE2 fallbacks are present here, should I remove these?

masterchef2209:
> These look good, but in addition to moving the implementations into the functions it would be great if you could add fallbacks on the SSE2 versions.

For `sra` intrinisics I have moved them inside the functions but for `srav` there are no fallbacks in SSE2 (they are in AVX512* but haven't been implemented yet).

nemequ:
LGTM.  Assmuing CI passes I'll merge them soon, thanks!

test/x86/avx2.c
```diff
@@ -9312,6 +9312,108 @@ test_simde_mm256_sllv_epi64(SIMDE_MUNIT_TEST_ARGS) {
   return 0;
 }
 
+static int
+test_simde_mm256_sra_epi16 (SIMDE_MUNIT_TEST_ARGS) {
+  static const struct {
+    const int16_t a[16];
+    const int16_t count[8];
```
nemequ, line 9319: count is actually 64x2.  They basically just shove a count into the lower half of the vector.

Since values > 15 all do the same thing, even though it's a 64-bit value you really only want to use the 4 least significant bits, so to generate it I would do something like

```c
simde__m128i count = simde_mm_srli_epi64(simde_test_x86_random_i64x2(), 60);
```

test/x86/avx2.c
```diff
@@ -9824,6 +9926,92 @@ test_simde_mm256_srai_epi32(SIMDE_MUNIT_TEST_ARGS) {
...
+static int
+test_simde_mm_srav_epi32 (SIMDE_MUNIT_TEST_ARGS) {
+  static const struct {
+    const int32_t a[4];
+    const int32_t b[4];
+    const int32_t r[4];
+  } test_vec[] = {
+    { {  INT32_C(   956480390), -INT32_C(   327735143), -INT32_C(   355390379), -INT32_C(   386052259) },
+      { -INT32_C(  1743728675), -INT32_C(  1123534065), -INT32_C(   600872973),  INT32_C(  1413968589) },
+      {  INT32_C(           0), -INT32_C(           1), -INT32_C(           1), -INT32_C(           1) } },
```
nemequ, line 9938: You're only getting 0 and 1 for results because you're always shifting by absurd values so you basically only get the highest bit.  The shift values should be 5 bits here.

masterchef2209: oh, I didn't think about this, I have redone the tests, hopefully they are correct this time.
native-aliases test is failing but it is unrelated to this PR.
Thank You

PR #433, opened 2020-06-25
neon: Add implementation of ceqz

seanptmaher:
I wrote this a bit ago, and then finished up the implementation today, so it's possible I missed something due to the mental disconnect of working on half of it at different times, but I guess we shall see. 

(CC @ngzhian @tlively)


seanptmaher:
No prob, I'll update the implementation. I'm glad you're on the lookout to spot these.

PR #439, opened 2020-06-26
neon: Fix sub native aliases

seanptmaher:
I noticed a problem with sub's native aliases. This should fix it. 

(CC @tlively @ngzhian)


nemequ:
Nice catch, thanks!  I really need to put together a native-aliases test for NEON to catch these issues in CI…

nemequ:
882764e1b61d90305bd239d956d066ec13af52db adds NEON support to the native aliases test.  Hopefully this won't happen again :)

PR #443, opened 2020-06-28
svml: implement {mm,mm256,mm512}_cdfnorminv_{ps,pd}

himanshi18037:


simde/simde-math.h
```diff
@@ -1005,8 +1178,8 @@
     x = (1.0f - x) * (1.0f + x);
     lnx = simde_math_logf(x);
 
-    tt1 = 2.0f / (HEDLEY_STATIC_CAST(float, SIMDE_MATH_PI) * 0.147f) + 0.5f * lnx;
-    tt2 = (1.0f / 0.147f) * lnx;
+    tt1 = 2.0f / (HEDLEY_STATIC_CAST(float, SIMDE_MATH_PI) * 0.15449436008930206298828125f) + 0.5f * lnx;
```
nemequ, line 1181: I know you got this value from me when, but it is intended to reduce the *maximum* absolute error.  Since the problem turned out to be something else I think in this case it would be better to minimize the *average* absolute error.

In my tests, 0.147 had an average absolute error of ~ 0.0001465479290345683693885803, and the best value to minimize the average absolute error was 0.1482909470796585083007812500 which had an average absolute error of 0.0001410958211636170744895935.

What do you think?

himanshi18037: Even I feel that we can change it back to 0.147 because it is also minimizing the error, so there is no point of keeping such long values,  I may have updated it when you said and forgot to change back, should I update it? Or you can update it while merging?

nemequ: AFAICT there is no advantage to using 0.147; `0.1482909470796585083007812500` gives us significantly more accurate results on average.  The length of the value doesn't hurt anything; it might be a bit ugly in the source code, but it takes up the same amount of space once it's compiled.

I can change it.

PR #445, opened 2020-06-29
neon: Correct excess arguments and also test neon.h

rosbif:
Test that both neon.h (in C version) and xxx.h (in C++ version) work.

@nemequ I think we agreed that it was a good idea to do this.


rosbif:
Sorry, I issued a PR for skel.c and then pushed some real stuff.
There are some cast issues to clear up so the complete PR isn't ready yet.

rosbif:
Many of these need a cast to avoid warning-errors but I am not sure how to cast them :-(
(r_ .values needs to be an unsigned vector but a_ .values and b_ .values may be signed, float or double.)
```
    #if defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
      r_.values = a_.values > b_.values;
    #else
```
Any ideas?

simde/arm/neon/cgt.h
```diff
@@ -45,12 +45,13 @@ simde_vcgt_f32(simde_float32x2_t a, simde_float32x2_t b) {
       b_ = simde_float32x2_to_private(b);
     simde_uint32x2_private r_;
 
-    #if defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
+    #if defined(SIMDE_VECTOR_SUBSCRIPT_OPS) && 0
+      /* These need a cast to avoid warning-errors but I am not sure how to cast it :-( */
```
nemequ, line 49: `HEDLEY_REINTERPRET_CAST`.  You can even cheat and do `HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > b_.values)` since every compiler that supports GCC-style vector extensions also supports `__typeof__` :)

simde/arm/neon/cgtz.h
```diff
@@ -0,0 +1,379 @@
...
+HEDLEY_DIAGNOSTIC_PUSH
+SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
+SIMDE_BEGIN_DECLS_
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde_uint32x2_t
+simde_vcgtz_f32(simde_float32x2_t a) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+    return vcgtz_f32(a);
+  #elif defined(SIMDE_NATURAL_VECTOR_SIZE) && (SIMDE_NATURAL_VECTOR_SIZE >= 64)
```
nemequ, line 44: Why not just make the whole function

```c
    return vcgtz_f32(a);
  #else
    return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
  #endif
```

?  Or maybe

```c
    return vcgtz_f32(a);
  #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
    simde_float32x2_private a_ = simde_float32x2_to_private(a);
    simde_uint32x2_private r_;

    r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));

    return simde_uint32x2_from_private(r_);
  #else
    return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
  #endif
```

The only time I can think of where we wouldn't want to fall back on this implementation is on an architecture which supports comparison to zero without having to involve another register, and the compiler is smart enough to auto-vectorize the loop to that instruction.  ARM is the only architecture I'm aware of that has comparison to zero built in (though there could be others), but even then putting the vector extension version first should take care of that.

What do you think?  It would certainly simplify things, and I'm having trouble thinking of any disadvantages.

test/arm/neon/skel.c
```diff
@@ -19,7 +25,7 @@ test_simde_vxxx_f32 (SIMDE_MUNIT_TEST_ARGS) {
     simde_float32x2_t b = simde_vld1_f32(test_vec[i].b);
     simde_float32x2_t r = simde_vxxx_f32(a, b);
 
-    simde_test_arm_neon_assert_equal_f32x2(r, simde_vld1_f32(test_vec[i].r), 1);
+    simde_test_arm_neon_assert_equal_f32x2(r, simde_vld1_f32(test_vec[i].r));
```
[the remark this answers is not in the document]
rosbif: > Those aren't extra arguments; it's the precision for floating point tests (the number of decimal places we test to). They're only present on the floating point assertions, I'm guessing you thought they were extra because you were replacing them with integer assertions.
> 
> I'm going to drop these, but if I'm missing something please feel free to file another PR.

Sorry, I was missing something. I was using these for compare operations which, with float inputs, return an unsigned, not a float so I just changed e.g. f32 to u32 in the test without realizing that this was the reason for the extra argument error :-(

simde/arm/neon/cgtz.h
rosbif: > Why not just make the whole function
> 
> ```c
>   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
>     return vcgtz_f32(a);
>   #else
>     return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
>   #endif
> ```
> 

I didn't want to do this because if there is no SIMD, this will create a loop for each sub-function and I am not sure whether all compilers will auto-vectorize and combine them.

> ? Or maybe
> 
> ```c
>   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
>     return vcgtz_f32(a);
>   #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
>     simde_float32x2_private a_ = simde_float32x2_to_private(a);
>     simde_uint32x2_private r_;
> 
>     r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));
> 
>     return simde_uint32x2_from_private(r_);
>   #else
>     return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
>   #endif
> ```
> 
> The only time I can think of where we wouldn't want to fall back on this implementation is on an architecture which supports comparison to zero without having to involve another register, and the compiler is smart enough to auto-vectorize the loop to that instruction. ARM is the only architecture I'm aware of that has comparison to zero built in (though there could be others), but even then putting the vector extension version first should take care of that.
> 
> What do you think? It would certainly simplify things, and I'm having trouble thinking of any disadvantages.
>

Frankly I didn't understand what you can do in a `SIMDE_VECTOR_SUBSCRIPT_*`. But I would prefer to force my solution than to hope that the compiler vectorization works. I don't see how the compiler could realize how to decompose e.g. the vrsra intrinsics.

rosbif:
Should I cancel the PR and issue another one later.
Otherwise every push generates the full monty of CI stuff.

simde/arm/neon/cgtz.h
nemequ: > I didn't want to do this because if there is no SIMD, this will create a loop for each sub-function and I am not sure whether all compilers will auto-vectorize and combine them.

That's a valid concern; compilers aren't great about fusing loops across functions like that.  Moving the vector extension version should take care of that, at least for compilers which support GCC-style vector extensions (basically, "not MSVC").

I guess the question is whether that is a bigger concern than an unnecessary splat.  My instinct is that it would be better to optimize for platforms where we have some hope of running at a reasonable speed; if we can't vectorize this function I feel like the battle is already lost.  However, after thinking about it a bit more I'm starting to think you're right here; NEON is the only platform I'm aware of with compare to zero instructions instead of making you compare to a zeroed register…

> Frankly I didn't understand what you can do in a `SIMDE_VECTOR_SUBSCRIPT_*`. But I would prefer to force my solution than to hope that the compiler vectorization works. I don't see how the compiler could realize how to decompose e.g. the vrsra intrinsics.

GCC-style vector extensions are a bit limited, but if they are supported they are extremely reliable (probably because they're so limited).  If they are available it's safe to assume that they'll provide the best result.

> Edit: I know you don't like my implementation

It's not that I dislike it; with a couple very small exceptions I think the code is correct, I'm just trying to figure out how we could reuse other functions so we can take advantage of optimizations there.

> (originally I didn't realize that SIMDE_NATURAL_VECTOR_SIZE could be undefined),

Okay, that explains some things.  I was really confused about why you're testing to see if it is >= 64.

If all you're interested in is whether the platform supports SIMD, I think what you're looking for here is the `SIMDE_ASSUME_VECTORIZATION` macro, which has been part of SIMDe for a long time.  Now that `SIMDE_NATURAL_VECTOR_SIZE` is a thing I've been thinking I should just change it to

```
#if defined(SIMDE_NATURAL_VECTOR_SIZE)
  #define SIMDE_ASSUME_VECTORIZATION
#endif
```

Or maybe just `s/SIMDE_ASSUME_VECTORIZATION/SIMDE_NATURAL_VECTOR_SIZE/`.  IIRC that was actually one of the original use cases for `SIMDE_NATURAL_VECTOR_SIZE`, but I forgot about it when I merged the patch.

> but how would you have used SIMDE_PREFER_VECTOR_SIZE instead?

Just

```c
#elif SIMDE_PREFER_VECTOR_SIZE(128)
  return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
```

But I really don't think it adds anything here.  We don't want to restrict it to vector lengths <= 128, we want to always fall back.  That's what `SIMDE_ASSUME_VECTORIZATION` is for.

It seems like what we really want to do here is something like

```c
    return vcgtz_f32(a);
  #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
    simde_float32x2_private a_ = simde_float32x2_to_private(a);
    simde_uint32x2_private r_;

    r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));

    return simde_uint64x1_from_private(r_);
  #elif defined(SIMDE_ASSUME_VECTORIZATION)
    return simde_vcgt_f32(a, simde_vdup_n_f32(SIMDE_FLOAT32_C(0.0)));
  #else
    simde_float32x2_private a_ = simde_float32x2_to_private(a);
    simde_uint32x2_private r_;

    SIMDE_VECTORIZE
      r_.values[i] = (a_.values[i] > SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
    }

    return simde_uint32x2_from_private(r_);
  #endif
```

I'm a bit groggy right now, but I *think* we can safely change the vector extension part to just

```c
  #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
    HEDLEY_REINTERPRET_CAST(simde_uint32x2_t, a > SIMDE_FLOAT32_C(0.0));
```

which would make that a *lot* more palatable.

nemequ:
> Should I cancel the PR and issue another one later.
> Otherwise every push generates the full monty of CI stuff.

It's up to you.  My preference would be to leave it open so that when this does get merged the commit message will link back to this conversation instead of a "freash" PR with no background.

If you want to avoid noise in CI you could just create a new branch ("pr/445" or something) in your repo and use that until you're comfortable with the contents, then push it to the original branch.

rosbif:
> > Should I cancel the PR and issue another one later.
> > Otherwise every push generates the full monty of CI stuff.
> 
> It's up to you. My preference would be to leave it open so that when this does get merged the commit message will link back to this conversation instead of a "freash" PR with no background.
> 
> If you want to avoid noise in CI you could just create a new branch ("pr/445" or something) in your repo and use that until you're comfortable with the contents, then push it to the original branch.
>

OK, I'll do that.

simde/arm/neon/cgtz.h
nemequ: > I'm a bit groggy right now, but I think we can safely change the vector extension part to just

I just thought about this a bit more; it wouldn't work reliably.  WASM is a counte-example :(

Maybe we should add functions like simde_float32x2_to_f32 if SIMDE_VECTOR_SUBSCRIPT is defined?  It would still be a bit ugly, but much better than what I wrote above…

rosbif:
@nemequ Thank you for your help with the `SIMDE_VECTOR_SUBSCRIPT_SCALAR` stuff.
The reinterpret  `__typeof__(r_.values)` works fine.
I have implemented it in cgt.h and cgtz.h and added it in tst.h.
I'll check elsewhere.

However we still differ on how and when to enable the "fallback to simpler (and therefore more common) intrinsics".
I don't want to fall back from 128-bit to 64-bit.
The only use-case for this would be for a machine with only MMX but I doubt that it would be worth the effort.

Do we agree on not falling back from 128-bit to MMX?

My intention was to fallback to simpler intrinsics only if the machine supports at least 128-bit vectors (and the fallback will generate an optimal (or near optimal) combination of intrinsics) .
I don't want to fallback if the machine only supports 64-bit vectors or if the machine has no SIMD at all.

I don't think that any of your suggestions implement this.

nemequ:
> However we still differ on how and when to enable the "fallback to simpler (and therefore more common) intrinsics".
> I don't want to fall back from 128-bit to 64-bit.
> The only use-case for this would be for a machine with only MMX but I doubt that it would be worth the effort.

Yeah, I'm not too worried about MMX but no SSE/SSE2/etc.  I think a much more important use case, though, is NEON functions which have 64-bit versions but not 128-bit versions.  I don't know how many there are, but not many, at least when you exclude stuff like vget_high/low, vcombine, etc.

> Do we agree on not falling back from 128-bit to MMX?

I don't see anything wrong with falling back on 64-bit NEON if we need to, but MMX is definitely more fraught.  FWIW, we don't set `SIMDE_NATURAL_VECTOR_SIZE` to 64 if all we have is MMX; we just pretend it doesn't exist.

My understanding is that MMX is really only a problem on x86, not x86_64 where it uses XMM (SSE) registers.  I should really do some more research here, but assuming that's true I don't see a reason not to fall back on MMX on x86_64.  AFAIK MMX doesn't actually offer anything that SSE2 doesn't, so there should be no need to fall back on MMX.

The trouble is that a lot of 64-bit NEON functions are implemented using MMX (which maybe they should only do on x86_64?), so if we fall back on them we might be falling back on MMX.  Again, though, if we have SSE2 I think any path where we call a 128-bit function and it falls back on MMX is a bug (missing SSE/SSE2 implementation).

Does your understanding of MMX line up with that?  I've just always heard that you really shouldn't use MM, at least unless you don't have to worry about putting pressure on the x87 (definitely not in libraries), but I'm not totally clear on the details.

> My intention was to fallback to simpler intrinsics only if the machine supports at least 128-bit vectors (and the fallback will generate an optimal (or near optimal) combination of intrinsics) .
> I don't want to fallback if the machine only supports 64-bit vectors or if the machine has no SIMD at all.

Okay, I think I see the disconnect.  Often we *want* to fall back on 2 * shorter vectors instead of portable versions.  It's just like going from 512->256, or 256->128.  However, what you want to do is only fall back on multiple "simpler" instructions for the same vector length if the CPU spports that size natively.

Keeping in mind I'm trying to generalize this so it would work for stuff like AVX512 -> AVX -> SSE, maybe something like:

```c
  return vcgtzq_f32(a);
#elif deifned(SIMDE_IMAGINARY_ARCH_WITH_SIMILAR_FUNCTION_NATIVE)
  return foo_cmpgt_zero_f32(a);
#elif SIMDE_NATIVE_VECTOR_SIZE_GTE(128)
  return vcgtq_f32(a, vdupq_n_f32(SIMDE_FLOAT32_C(0.0)));
#elif SIMDE_NATIVE_VECTOR_SIZE_LTE(64) // <- currently never true on x86
  vcombine_f32(
      vcgtz_f32(vget_low_f32(a)),
      vcgtz_f32(vget_high_f32(a)));
#else
  simde_float32x4_private a_ = simde_float32x4_to_private(a);
  simde_uint32x4_private r_;

  #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
    r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));
  #else
    SIMDE_VECTORIZE
      r_.values[i] = (a_.values[i] > SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
    }
  #endif

  return simde_u32x4_from_private(r_);
#endif
```

?

I think it's probably obvious, but `SIMDE_NATIVE_VECTOR_SIZE_LTE` would also return false vectorization isn't supported at all.

I don't really know if the `SIMDE_NATIVE_VECTOR_SIZE_LTE` part is worth it; AFAIK NEON an MMX are the only extensions with 64-bit instructions, but maybe something I'm not really familiar with (SPARC, MIPS, etc.)?

rosbif:
@nemequ Thank you for your detailed reply.

> > However we still differ on how and when to enable the "fallback to simpler (and therefore more common) intrinsics".
> > I don't want to fall back from 128-bit to 64-bit.
> > The only use-case for this would be for a machine with only MMX but I doubt that it would be worth the effort.
> 
> Yeah, I'm not too worried about MMX but no SSE/SSE2/etc. I think a much more important use case, though, is NEON functions which have 64-bit versions but not 128-bit versions. I don't know how many there are, but not many, at least when you exclude stuff like vget_high/low, vcombine, etc.
> 

I haven't noticed any functions where it would be useful to fall back to 64-bit NEON. Are there machines with only 64-bit NEON? Was 64-bit NEON introduced before 128-bit (like MMX preceding SSE)?
If this is the case then I agree to fall back to 64-bit NEON, but it could be done selectively.

> > Do we agree on not falling back from 128-bit to MMX?
> 
> I don't see anything wrong with falling back on 64-bit NEON if we need to, but MMX is definitely more fraught. FWIW, we don't set `SIMDE_NATURAL_VECTOR_SIZE` to 64 if all we have is MMX; we just pretend it doesn't exist.
> 

I hadn't noticed this. Actually I would have preferred `SIMDE_NATURAL_VECTOR_SIZE` to always be defined. A scalar is akin to a vector of length 1, so I would have set it to 1 if there is no SIMD. Bur I think this may screw up your `SIMDE_PREFER_VECTOR_SIZE`.

> My understanding is that MMX is really only a problem on x86, not x86_64 where it uses XMM (SSE) registers. I should really do some more research here, but assuming that's true I don't see a reason not to fall back on MMX on x86_64. AFAIK MMX doesn't actually offer anything that SSE2 doesn't, so there should be no need to fall back on MMX.
> 
> The trouble is that a lot of 64-bit NEON functions are implemented using MMX (which maybe they should only do on x86_64?), so if we fall back on them we might be falling back on MMX. Again, though, if we have SSE2 I think any path where we call a 128-bit function and it falls back on MMX is a bug (missing SSE/SSE2 implementation).
> 
> Does your understanding of MMX line up with that? I've just always heard that you really shouldn't use MM, at least unless you don't have to worry about putting pressure on the x87 (definitely not in libraries), but I'm not totally clear on the details.
> 

If MMX is really as evil as you say then should we implement 64-bit NEON with 128-bit functions, dropping the top 64 bits?
This would also permit 64-bit NEON to use AltiVec or WASM which AFAIK lack 64-bit versions. It would be very easy, except for the few, if any, 64-bit intrinsics lacking 128-bit counterparts. Something like, if we don't have native NEON but do have 128-bit SIMD, implementing e.g. `vadd(a, b)` with `a128 = vcombine(a, a); b128 = vcombine(b, b); r128 = vaddq(a128, b128); return vget_low(r128);`
(I have omitted the types and I have never used vcombiine and vget_low so this is wrong but you get the idea.)

> > My intention was to fallback to simpler intrinsics only if the machine supports at least 128-bit vectors (and the fallback will generate an optimal (or near optimal) combination of intrinsics) .
> > I don't want to fallback if the machine only supports 64-bit vectors or if the machine has no SIMD at all.
> 
> Okay, I think I see the disconnect. Often we _want_ to fall back on 2 * shorter vectors instead of portable versions. It's just like going from 512->256, or 256->128. However, what you want to do is only fall back on multiple "simpler" instructions for the same vector length if the CPU spports that size natively.
> 

Yes, in this particular case, this is exactly what I want to do.
(But generally I am a fan of "divide and conquer" which has two aspects in SIMDe:
- falling back to half-length vectors until we find a size we can handle,
- falling back to simpler operations until we find one we can handle.)

> Keeping in mind I'm trying to generalize this so it would work for stuff like AVX512 -> AVX -> SSE, maybe something like:
> 
> ```c
> #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
>   return vcgtzq_f32(a);
> #elif deifned(SIMDE_IMAGINARY_ARCH_WITH_SIMILAR_FUNCTION_NATIVE)
>   return foo_cmpgt_zero_f32(a);
> #elif SIMDE_NATIVE_VECTOR_SIZE_GTE(128)
>   return vcgtq_f32(a, vdupq_n_f32(SIMDE_FLOAT32_C(0.0)));
> #elif SIMDE_NATIVE_VECTOR_SIZE_LTE(64) // <- currently never true on x86
>   vcombine_f32(
>       vcgtz_f32(vget_low_f32(a)),
>       vcgtz_f32(vget_high_f32(a)));
> #else
>   simde_float32x4_private a_ = simde_float32x4_to_private(a);
>   simde_uint32x4_private r_;
> 
>   #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
>     r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));
>   #else
>     SIMDE_VECTORIZE
>     for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
>       r_.values[i] = (a_.values[i] > SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
>     }
>   #endif
> 
>   return simde_u32x4_from_private(r_);
> #endif
> ```
> 
> ?
> 
> I think it's probably obvious, but `SIMDE_NATIVE_VECTOR_SIZE_LTE` would also return false vectorization isn't supported at all.
> 
> I don't really know if the `SIMDE_NATIVE_VECTOR_SIZE_LTE` part is worth it; AFAIK NEON an MMX are the only extensions with 64-bit instructions, but maybe something I'm not really familiar with (SPARC, MIPS, etc.)?
>

So what do we conclude?
I think I would be happy with the following modifications to what exists:
a) `SIMDE_NATURAL_VECTOR_SIZE` always defined (to 1 with no SIMD as above),
b) I am not sure about `SIMDE_PREFER_VECTOR_SIZE()` but I would prefer a `SIMDE_REDUCE_VECTOR_SIZE() `which indicates that we should fall back on half size vectors but which should return 0 for no SIMD (and probably also return 0 if there is only 64-bit SIMD).

nemequ:
> @nemequ Thank you for your detailed reply.
> 
> > > However we still differ on how and when to enable the "fallback to simpler (and therefore more common) intrinsics".
> > > I don't want to fall back from 128-bit to 64-bit.
> > > The only use-case for this would be for a machine with only MMX but I doubt that it would be worth the effort.
> > 
> > 
> > Yeah, I'm not too worried about MMX but no SSE/SSE2/etc. I think a much more important use case, though, is NEON functions which have 64-bit versions but not 128-bit versions. I don't know how many there are, but not many, at least when you exclude stuff like vget_high/low, vcombine, etc.
> 
> I haven't noticed any functions where it would be useful to fall back to 64-bit NEON. Are there machines with only 64-bit NEON? Was 64-bit NEON introduced before 128-bit (like MMX preceding SSE)?
> If this is the case then I agree to fall back to 64-bit NEON, but it could be done selectively.

I don't *think* machines with only a 64-bit NEON API are a thing, no.  My understanding is that ARMv7 NEON are really 64-bit registers in hardware, but there are 128-bit aliases you can use which span two 64-bit registers, and AFAIK that ability has been present since the first release.  ARMv8 NEON has real 128-bit registers which can also be used as 64-bit registers.

> > > Do we agree on not falling back from 128-bit to MMX?
> > 
> > 
> > I don't see anything wrong with falling back on 64-bit NEON if we need to, but MMX is definitely more fraught. FWIW, we don't set `SIMDE_NATURAL_VECTOR_SIZE` to 64 if all we have is MMX; we just pretend it doesn't exist.
> 
> I hadn't noticed this. Actually I would have preferred `SIMDE_NATURAL_VECTOR_SIZE` to always be defined. A scalar is akin to a vector of length 1, so I would have set it to 1 if there is no SIMD. Bur I think this may screw up your `SIMDE_PREFER_VECTOR_SIZE`.

We can always rewrite `SIMDE_PREFER_VECTOR_SIZE`, or drop it altogether.

I didn't define `SIMDE_NATURAL_VECTOR_SIZE` when there is no natural vector size because, well, there is no natural vector size.  I don't see why you would want to define it to 1, but 0 would make sense to me.  Why would you want to define it to 1?

> > My understanding is that MMX is really only a problem on x86, not x86_64 where it uses XMM (SSE) registers. I should really do some more research here, but assuming that's true I don't see a reason not to fall back on MMX on x86_64. AFAIK MMX doesn't actually offer anything that SSE2 doesn't, so there should be no need to fall back on MMX.
> > The trouble is that a lot of 64-bit NEON functions are implemented using MMX (which maybe they should only do on x86_64?), so if we fall back on them we might be falling back on MMX. Again, though, if we have SSE2 I think any path where we call a 128-bit function and it falls back on MMX is a bug (missing SSE/SSE2 implementation).
> > Does your understanding of MMX line up with that? I've just always heard that you really shouldn't use MM, at least unless you don't have to worry about putting pressure on the x87 (definitely not in libraries), but I'm not totally clear on the details.
> 
> If MMX is really as evil as you say

I don't think it's really evil; it's just aliased registers on the x87 (IIRC x87 uses an 80-bit extended precision format with a 64-bit mantissa internally, and MMX uses the mantissa bits), so if you use MMX registers it puts a lot of pressure on the FPU.  On x86_64, though, it uses the SSE registers.  If you don't do a lot of FP in your code, MMX is excellent.

> then should we implement 64-bit NEON with 128-bit functions, dropping the top 64 bits?

That could be a rather nice feature on some architectures, but I don't see how we could do it while preserving `sizeof(T) == 8`.  I think we'd have to hide it behind an opt-in macro.  It may actually work pretty well; until AArch64 Arm was pretty strict about alignment requirements, so stuff like `reinterpret_cast<int8x8_t*>(foo)` was pretty rare; I think most people actually use the vld1 / vst1 functions.

Last I heard, [GCC was working on something like this](https://www.phoronix.com/scan.php?page=news_item&px=GCC-MMX-Intrinsics-With-SSE).  IDK how much of that got merged, or if clang did the same thing.  It would definitely make sense, though.

> This would also permit 64-bit NEON to use AltiVec or WASM which AFAIK lack 64-bit versions. It would be very easy, except for the few, if any, 64-bit intrinsics lacking 128-bit counterparts. Something like, if we don't have native NEON but do have 128-bit SIMD, implementing e.g. `vadd(a, b)` with `a128 = vcombine(a, a); b128 = vcombine(b, b); r128 = vaddq(a128, b128); return vget_low(r128);`
> (I have omitted the types and I have never used vcombiine and vget_low so this is wrong but you get the idea.)
> 
> Edit: I really like this idea, but maybe auto-vectorization already does this automatically.
> Do you know if it does?

I don't think they're usually very good about this, no.

> > > My intention was to fallback to simpler intrinsics only if the machine supports at least 128-bit vectors (and the fallback will generate an optimal (or near optimal) combination of intrinsics) .
> > > I don't want to fallback if the machine only supports 64-bit vectors or if the machine has no SIMD at all.
> > 
> > 
> > Okay, I think I see the disconnect. Often we _want_ to fall back on 2 * shorter vectors instead of portable versions. It's just like going from 512->256, or 256->128. However, what you want to do is only fall back on multiple "simpler" instructions for the same vector length if the CPU spports that size natively.
> 
> Yes, in this particular case, this is exactly what I want to do.
> (But generally I am a fan of "divide and conquer" which has two aspects in SIMDe:
> 
>     * falling back to half-length vectors until we find a size we can handle,
> 
>     * falling back to simpler operations until we find one we can handle.)
> 
> 
> > Keeping in mind I'm trying to generalize this so it would work for stuff like AVX512 -> AVX -> SSE, maybe something like:
> > ```c
> > #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
> >   return vcgtzq_f32(a);
> > #elif deifned(SIMDE_IMAGINARY_ARCH_WITH_SIMILAR_FUNCTION_NATIVE)
> >   return foo_cmpgt_zero_f32(a);
> > #elif SIMDE_NATIVE_VECTOR_SIZE_GTE(128)
> >   return vcgtq_f32(a, vdupq_n_f32(SIMDE_FLOAT32_C(0.0)));
> > #elif SIMDE_NATIVE_VECTOR_SIZE_LTE(64) // <- currently never true on x86
> >   vcombine_f32(
> >       vcgtz_f32(vget_low_f32(a)),
> >       vcgtz_f32(vget_high_f32(a)));
> > #else
> >   simde_float32x4_private a_ = simde_float32x4_to_private(a);
> >   simde_uint32x4_private r_;
> > 
> >   #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
> >     r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > SIMDE_FLOAT32_C(0.0));
> >   #else
> >     SIMDE_VECTORIZE
> >     for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
> >       r_.values[i] = (a_.values[i] > SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
> >     }
> >   #endif
> > 
> >   return simde_u32x4_from_private(r_);
> > #endif
> > ```
> > 
> > 
> > ?
> > I think it's probably obvious, but `SIMDE_NATIVE_VECTOR_SIZE_LTE` would also return false vectorization isn't supported at all.
> > I don't really know if the `SIMDE_NATIVE_VECTOR_SIZE_LTE` part is worth it; AFAIK NEON an MMX are the only extensions with 64-bit instructions, but maybe something I'm not really familiar with (SPARC, MIPS, etc.)?
> 
> Edit: I like this last proposal although I think I would prefer:
> 
>     * `SIMDE_NATIVE_VECTOR_SIZE_LT(128)` to `SIMDE_NATIVE_VECTOR_SIZE_LTE(64)` or maybe `SIMDE_REDUCE_VECTOR_SIZE(128)` as `LT` isn't strictly true due to the omissions (MMX and no vectorization).
>
> 
>     * `SIMDE_NATIVE_VECTOR_SIZE_GE` rather than `SIMDE_NATIVE_VECTOR_SIZE_GTE` but this is nitpicking.

I think checking based on the destination size is slightly more sensible; we don't really wan to fall back on a 64-bit implementation if we have an 80-bit vector or something.  Of course I can't imagine we would ever have an 80-bit vector, but I'd argue that it's the more correct of the two.

Anyways, it sounds like you don't have a particularly strong opinion here; neither do I.  If you have an argument more persuasive than the 80-bit thing I'd be willing to go that route.

> 
> However I think I would also be happy with `SIMDE_NATURAL_VECTOR_SIZE` if it is always defined (to 1 with no SIMD as above),
> 
> So what do we conclude?

It sounds like always definining `SIMDE_NATURAL_VECTOR_SIZE` would give us the most flexibility and resolve the issue for now, we just need to decide between 0 and 1… do you have an argument for 1, or is it just habit for macros?

`SIMDE_NATIVE_VECTOR_SIZE_{LTE,GTE}` (or, as above, LT/GT) would be trivial to add on top of that, and I think the `SIMDE_PREFER_VECTOR_SIZE` name has proved sufficiently ambiguous that we should probably drop it.

rosbif:
@nemequ I think we almost agree now.

> It sounds like always definining `SIMDE_NATURAL_VECTOR_SIZE` would give us the most flexibility and resolve the issue for now, we just need to decide between 0 and 1… do you have an argument for 1, or is it just habit for macros?
>

I suggested 1 because a scalar is rather like a vector of length 1.  But if you prefer 0 that is OK with me.

nemequ:
> Edit: I now realize that my suggestion of 1 makes no sense: I was confusing the size with the number of elements :-(
> So I now agree with you that 0 is preferable.

Ah, yeah if it were number fo elements it 1 would have made sense.

> The other question is should I try to convert NEON 64-bit operations to the equivalent 128-bit ones (for AltiVec, WASM and possibly Intel) as I described above like `vadd(a, b) -> vget_low(vaddq(vcombine(a, a), vcombine(b, b)))` ?

I'm a bit skeptical that would perform well since you'd be constantly splitting and merging vectors.  I suspect that the only way to get it to perform acceptably would be to actually make the types 128 bits, so calls to vadd would be implemented with vaddq, no combine/get_low.  It would require the compiler to be smart enough to understand every intrinsic well enough to know that we throw away the high half and it should, too.  I think clang will actually be able to handle that for simple intrinsics (like addition, subtraction, etc.) which are implemented using GCC-style vector extensions, but I doubt it would be the case for most functions.  I'd be happy to be proven wrong, though, if you want to try to port a few functions to test. 

The drawback to doing it this way is that the size of each vector would be 16 instead of 8.  I don't actually think that would be a huge problem for NEON; people seem to me much better about using `vld1`/`vst1` instead of just aliasing, but I'm sure it would break some code, so we would have to hide it behind an off-by-default macro, something like `SIMDE_ENABLE_VECTOR_EXPANSION` (or whatever).  The good news is that for most functions it would be very easy to implement; just call the 128-bit NEON functions (well, the SIMDe versions).

If you're interested I would definitely accept a patch like that; I think it could make porting applications using 64-bit vectors much easier.

rosbif:
For all logical operations and for 64x1 vectors, the best solution would probably be to pass instead via 64-bit scalars (and maybe 32-bit scalars for arithmetic 32x2 vectors operations). However for arithmetic 8x8 and 16x4 operations I think passing via 128-bit vectors would be the best solution.
However I have no idea whether the compiler will find the same (or better!) transformations when auto-vectorizing. I guess also that different compilers will perform differently.
Do you know of any code making heavy use of 64-bit NEON which could be used as a benchmark to compare different implementations?
I shall add vget_low/combine implementations for the 64-bit intrinsics in my PR. Should I use `SIMDE_ENABLE_VECTOR_EXPANSION`  to condition this?
(In gitter I suggested `SIMDE_NEON_CONVERT_64_TO_128`.)
(I don't feel like really forcing the types to 128-bits: I fear that there will be dangerous side-effects.)

To decide whether to simplify operations I'll use `SIMDE_NATURAL_VECTOR_SIZE` (defined to zero if undefined).

For the moment I can fix up these macros in types.h which is always included.
Is this OK with you?

rosbif:
Incidentally the code that failed on clang on PPC64BE was this (I added the `&& 0` to force execution):
```
SIMDE_FUNCTION_ATTRIBUTES
simde_uint8x8_t
simde_vcgt_s8(simde_int8x8_t a, simde_int8x8_t b) {
  #if defined(SIMDE_ARM_NEON_A32V7_NATIVE) && 0
    return vget_low_u8(vcgtq_s8(vcombine_s8(a, a), vcombine_s8(b, b)));
    //    return vcgt_s8(a, b);
  #elif defined(SIMDE_X86_MMX_NATIVE) && 0
    return _mm_cmpgt_pi8(a, b);
  #elif defined(SIMDE_NATURAL_VECTOR_SIZE) && (SIMDE_NATURAL_VECTOR_SIZE >= 128)
    return simde_vget_low_u8(simde_vcgtq_s8(simde_vcombine_s8(a, a), simde_vcombine_s8(b, b)));
  #else
    simde_int8x8_private
      a_ = simde_int8x8_to_private(a),
      b_ = simde_int8x8_to_private(b);
    simde_uint8x8_private r_;

    #if defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
      r_.values = HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > b_.values);
    #else
      SIMDE_VECTORIZE
        r_.values[i] = (a_.values[i] > b_.values[i]) ? UINT8_MAX : 0;
      }
    #endif

    return simde_uint8x8_from_private(r_);
  #endif
}
#if defined(SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES)
  #undef vcgt_s8
  #define vcgt_s8(a, b) simde_vcgt_s8((a), (b))
#endif


```

nemequ:
> For all logical operations and for 64x1 vectors, the best solution would probably be to pass instead via 64-bit scalars (and maybe 32-bit scalars for arithmetic 32x2 vectors operations). However for arithmetic 8x8 and 16x4 operations I think passing via 128-bit vectors would be the best solution.

I don't think mixing like that is going to be an option; it would make `vreinterpret` operations (which are supposed to be free) require inserts/extracts, which will break a *lot* of performance assumptions, and you don't really solve anything since most types would still use 128-bit vectors.  People doing a lot of 64x1 operations aren't generaly doing it in NEON registers anyways, so you wouldn't gain much.

FWIW, I did [some quick tests](https://godbolt.org/z/kttP2t) with the inserting and extracting version.  clang actually does pretty well for the really simple stuff (lke add), but is quite bad for more complex operations.  GCC is just bad across the board.

> However I have no idea whether the compiler will find the same (or better!) transformations when auto-vectorizing. I guess also that different compilers will perform differently. (But if all compilers auto-vectorize well all this effort is wasted :-( )

It depends.  If GCC-style vector extensions are supported they'll generally work pretty well, at least when the operation is something they support (possibly with the help of `__builtin_convertvector`.  If you can actually pass around the data in 128-bit buffers, though, you'll easily beat the auto-vectorizer for more complicated operations, and will be no worse for the simple stuff.

> Do you know of any code making heavy use of 64-bit NEON which could be used as a benchmark to compare different implementations?

Nope, sorry.

> I shall add vget_low/combine implementations for the 64-bit intrinsics in my PR. Should I use `SIMDE_ENABLE_VECTOR_EXPANSION` to condition this?
> (In gitter I suggested `SIMDE_NEON_CONVERT_64_TO_128`.)

I'm not a fan of using "convert" here because it already has meaning in SIMD; basically, stuff that could be implemented with `SIMDE_CONVERT_VECTOR_` (the `_mm_cvt*` functions, for example).  "Widen" in similarly taken.  I don't think "expand" is in use yet, though.

> (I don't feel like really forcing the types to 128-bits: I fear that there will be dangerous side-effects.)

Neither do it, but that's why it would be hidden behind an off-by-default macro.

> To decide whether to simplify operations I'll use `SIMDE_NATURAL_VECTOR_SIZE` (defined to zero if undefined).
> 
> For the moment I can fix up these macros in types.h which is always included.
> Is this OK with you?

I think we'll eventually want to add a global setting here to cover MMX as well, but yes, something like `SIMDE_ARM_NEON_VECTOR_EXPANSION` in types.h would make sense.

nemequ:
You didn't force push to the branch in this PR, so I can't review the code in this PR, but I went ahead and merged most of what was in your pr/445 branch, I hope you don't mind.  I put what I didn't merge in the wip/rosbif branch (in the main repo) as two patches.

One of them is for the *_n functions.  I'm wondering if there was a reason you implemented all those with macros outside the function instead of using the CONSTIFY macros, which were intended to remove the need for code like this.  They would also provide an easy way to improve the test cases since the constness checks are disabled for the tests.  I'm willing to merge that patch as-is if you really want, I'm just not sure what's going on here.

The other patch is for the EXPANSION stuff.  I did some testing, and performance for what you had was pretty bad.  I think we really need to pass the vectors as 128-bit types like we discussed.  I thought we were on the same page about that, but based on the patch it seems like you're still trying to push ahead with creating a 128 bit vector, performing the operation, then extracting 64 bits of it to return.  Did you try to get it working with 128 bit vectors and run into an unforseen problem?

rosbif:
> You didn't force push to the branch in this PR, so I can't review the code in this PR, but I went ahead and merged most of what was in your pr/445 branch, I hope you don't mind. I put what I didn't merge in the wip/rosbif branch (in the main repo) as two patches.
> 

No that's fine. I thought the plan for me was to work in pr/445 to avoid CI noise then merge here when all was agreed.
But I admit that I was worried about screwing up this merge so I'm glad you did it ;-)

> One of them is for the *_n functions. I'm wondering if there was a reason you implemented all those with macros outside the function instead of using the CONSTIFY macros, which were intended to remove the need for code like this. They would also provide an easy way to improve the test cases since the constness checks are disabled for the tests. I'm willing to merge that patch as-is if you really want, I'm just not sure what's going on here.
> 

I had already written the test cases (or at least similar ones I could C&P) so I didn't need to use CONSTIFY. But if by "improve the test cases" you mean test undocumented, out of range, shift counts, I did try this but I ran into compile failures even if the code with the out of range value couldn't be executed (it was protected by an ? : test.)
 
> The other patch is for the EXPANSION stuff. I did some testing, and performance for what you had was pretty bad. I think we really need to pass the vectors as 128-bit types like we discussed. I thought we were on the same page about that, but based on the patch it seems like you're still trying to push ahead with creating a 128 bit vector, performing the operation, then extracting 64 bits of it to return. Did you try to get it working with 128 bit vectors and run into an unforseen problem?
>

Sorry there is obviously a misunderstanding. I thought we had agreed that using 128-bit vectors directly for 64-bit stuff was too dangerous.

nemequ:
> No that's fine. I thought the plan for me was to work in pr/445 to avoid CI noise then merge here when all was agreed.
> But I admit that I was worried about screwing up this merge so I'm glad you did it ;-)

Oh, I thought the idea was to work in pr/445 to avoid CI noise here, then do a push to your neon2 branch (a force push would have been okay, no need to merge) once you were ready for me to review and I would merge it into master from there.  Anyways, it's done now.

> I had already written the test cases (or at least similar ones I could C&P) so I didn't need to use CONSTIFY. But if by "improve the test cases" you mean test undocumented, out of range, shift counts, I did try this but I ran into compile failures even if the code with the out of range value couldn't be executed (it was protected by an ? : test.)

No, not really test out-of-range values.  The SIMDE_REQUIRE_CONSTANT_RANGE macros should protect us from that on clang (and derivatives), and at least document the requirements on other compilers.  I just meant testing different values instead of hard-coding a few values, and making the code a bit easier to manage since there is no need for constant values.

Unfortunately I think the macros have a tendency to get pretty out of hand; some of the x86 implementations (not the tests, the stuff in the headers) are absolutely insane.  CONSTIFY is a bit of an ugly hack, but the resulting code is a lot cleaner.  Take a look at get_lane.h, you should see what I mean.

> Sorry there is obviously a misunderstanding. I thought we had agreed that using 128-bit vectors directly for 64-bit stuff was too dangerous.

I agree that it can break code, that's why I suggested hiding it behind an off-by-default macro.

Based on my tests, I don't actually think the combine/get_low approach is significantly faster than the existing code… it's probably better for more complicated instructions, and maybe for smaller lanes (and therefore more lanes), but for simple stuff like adds, shifts, bitwise operations, etc., it seems to actually be a bit slower.

Passing around a 128-bit type, OTOH, is very fast.

As for breaking code, I admit I was initially pretty skeptical about the idea, but I think it actually works for NEON.  Since arm was pretty strict about unaligned access until Aarch64, the vast majority of NEON code I've seen actually uses load and store functions instead of relying on aliasing like people tend to do on x86.  You don't really see much NEON code where there is just a big array of data with an `int32x2_t*` that people use to iterate over it, so there is actually quite a good chance that NEON code in the wild would just work with.  Even most of the code that wouldn't work woud only require slight modifications.

rosbif:
I thought the idea was to agree on the implementation *before* merging pr/445 to neon2.

While you were writing this I added a couple of edits above giving additional explanations.
However frankly, other than adding the selective 8x8, ... macros I suggested, I don't want to spend more time on these.
If someone is using SIMDe, he must have the source code. So if he is worried about performance, he should convert from a 64-bit to a 128-bit implementation. It is fairly simple as I have yet to come across a 64-bit NEON intrinsic without a direct 128-bit counterpart.

nemequ:
> I thought the idea was to agree on the implementation _before_ merging pr/445 to neon2.

Sorry, I should have been more clear.  Either way, it worked out so it doesn't really matter :)

> It is fairly simple as I have yet to come across a 64-bit NEON intrinsic without a direct 128-bit counterpart.

Yeah, I'd be surprised if there are any.  Since the 64-bit registers are just half of a 128-bit register the hardware is already there.  My only real worry is stuff like shuffles, scatter/gather, etc.  I don't know if they would be a problem, I just haven't looked at them with this question in mind, and worry they might be.

> However frankly, other than adding the selective 8x8, ... macros I suggested, I don't want to spend more time on these.

Okay, give me some time to put together a prototype implementation using 128-bit types and see how it works, then we can do some more direct comparisons and see what works better.  Your code isn't going anywhere; I might put it in another branch, but I won't get ird of it.  I'll make sure to look at different element sizes when testing, and I'll try to find some code which uses 64-bit NEON instructions to test with.

Meanwhile, do you want me to just go ahead and merge the other patch as-is and just add a todo to try switching to constify?  I'd prefer to use constify for new code in the future, but since this is already written I don't think there is sufficient reason to hold off…

rosbif:
> > It is fairly simple as I have yet to come across a 64-bit NEON intrinsic without a direct 128-bit counterpart.
> 
> Yeah, I'd be surprised if there are any. Since the 64-bit registers are just half of a 128-bit register the hardware is already there. My only real worry is stuff like shuffles, scatter/gather, etc. I don't know if they would be a problem, I just haven't looked at them with this question in mind, and worry they might be.
> 

OK, I didn't think about these.

> > However frankly, other than adding the selective 8x8, ... macros I suggested, I don't want to spend more time on these.
> 
> Okay, give me some time to put together a prototype implementation using 128-bit types and see how it works, then we can do some more direct comparisons and see what works better. Your code isn't going anywhere; I might put it in another branch, but I won't get ird of it. I'll make sure to look at different element sizes when testing, and I'll try to find some code which uses 64-bit NEON instructions to test with.
> 

Agreed.

> Meanwhile, do you want me to just go ahead and merge the other patch as-is and just add a todo to try switching to constify? I'd prefer to use constify for new code in the future, but since this is already written I don't think there is sufficient reason to hold off…
>

Yes, I would like it to be merged. AFAIK it isn't actually *wrong*. (Release early, release often.)
And it is certainly easier to modify it than to start over.

OK, I shall use CONSTIFY, where applicable, in future.

PR #450, opened 2020-06-30
avx2: added impl for mm256_permutevar8x32_{epi32,ps}

masterchef2209:
avx2: added impl for mm256_permutevar8x32_{epi32,ps}


simde/x86/avx2.h
```diff
@@ -2335,6 +2335,57 @@ SIMDE_REQUIRE_RANGE(imm8, 0, 255) {
...
+  #else
+    simde__m256i_private
+      r_,
+      a_ = simde__m256i_to_private(a),
+      idx_ = simde__m256i_to_private(idx);
+
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.i32) / sizeof(r_.i32[0])) ; i++) {
+      const int id = idx_.i32[i] & 7;
+      r_.i32[i] = a_.i32[id];
```
nemequ, line 2352: I'm going to just do `r_.i32[i] = a_.i32[idx_.i32[i] & 7];`.  I think it's simple enough that we can get away with not using a temporary variable here.

You don't need to do anything, I'll tweak it when I merge.  Just thought I should mention it for future reference.

PR #451, opened 2020-06-30
neon: implementation of st1_lane

seanptmaher:
Hello,

I had to heavily modify the skel.c test file to do the tests (seeing as it's not really producing a result, and it's got a bunch of differently typed arguments), so if you could just double check everything there to make sure it's kosher, that'd be great. 

(CC @tlively @ngzhian)


test/arm/neon/st1_lane.c
```diff
@@ -0,0 +1,1244 @@
+#define SIMDE_TEST_ARM_NEON_INSN st1_lane
+
+#include "test-neon.h"
+#include "../../../simde/arm/neon/st1_lane.h"
+
+static int
+test_simde_vst1_lane_f32 (SIMDE_MUNIT_TEST_ARGS) {
+#if 1
```
ngzhian, line 8: leftover from debugging?

seanptmaher: Nah, we (Evan, rosbif, and I) decided to leave the test generation code in there in case we need to revisit it later.

seanptmaher:
Hm, so there's a mistake here with the way that I do the tests, it's complaining that I am not using constant immediates. I had assumed that since I'm referencing a `const` struct it would be fine, the constant would be folded or something, but I guess not. Do you know how exactly I should go about fixing that?

ngzhian:
Unfortunately, it looks like you have to specify the lane immediate as a constant: see https://godbolt.org/z/9JdDYp
I could get closer by making `test_vec` constexpr, but the problem then falls on `i`, which is the loop counter and hence needs to be mutated. If there's a way to do `constexpr` loop, maybe it can be done?

ngzhian:
I think you could use templates to hack around it (found http://mmore500.com/2019/07/07/cpp-compile-loop.html to try to loop over ints at compile time), an alternative is to wrap the test code up into a templated function, something like https://godbolt.org/z/WPadMD.

nemequ:
SIMDe works in C and C++, so unfortunately templates aren't an option.  However, we already have [a "CONSTIFY" macro for this](https://github.com/simd-everywhere/simde/blob/master/simde/simde-constify.h) which just puts everything in a giant switch.  Since the value is still a compile-time constant this doesn't introduce any runtime penalty.

We use the `SIMDE_REQUIRE_CONSTANT` (or `SIMDE_REQUIRE_CONSTANT_RANGE`) macro to verify that we're receiving a compile-time constant (in the `__builtin_constant_p()` sense).  It only works on clang-derived compliers (including emscripten) right now, but since "not working" just means we accept code we don't really want to it's not a huge problem.

For an example, see [neon/get_lane](https://github.com/simd-everywhere/simde/blob/72d1da815a6227c706381fdcfbcd492f91a8f4c0/simde/arm/neon/get_lane.h),

I don't know why compilers require integer constant expressions here instead of just compile-time constants, but it's not something we can change so we have to live with it (and work around it).

The other possibility, which is what we usually did until recently, was just implement functions like these as macros, at least on the targets which require an ICE.  There are some annoying problems with that, though, when dealing with non-trivial implementations (I can provide details, but I doubt anyone cares).

test/arm/neon/st1_lane.c
nemequ: Yeah, in the NEON implementations the test files are small enough that it's not a problem.  For x86 we still delete the test generation code, though, since those are massive.

It also helps that we have [a separate repository](https://github.com/simd-everywhere/simde-no-tests) without the tests, so the bloat shouldn't cause problems for projects embedding SIMDe.

seanptmaher:
Since `SIMDE_CONSTIFY` took a result, and I didn't really feel like passing a value I'd ignore, I wrote up a "SIMDE_CONSTIFY_X_NORESULT_` set of macros that do the exact same thing but without assigning a variable, seems to work well enough here (at least, the tests run on ARM).

nemequ:
Seems reasonable enough.

I hope you don't mind, but I tweaked it to use NO_RESULT instead of NORESULT.

PR #458, opened 2020-07-02
avx2: added mm256_sad_epu8, mm256_mpsadbw_epu8, mm256_maddubs_epi16

masterchef2209:
avx2: added impl for mm256_sad_epu8
avx2: added impl for mm256_mpsadbw_epu8
avx2: added impl for mm256_maddubs_epi16


masterchef2209:
I have done a straightforward implementation for mm256_maddubs_epi16 in this PR. For the parallel version, what I understood from [this comment](https://github.com/simd-everywhere/simde/pull/410#issuecomment-647680724) is that we have to first perform a widening multiplication after which we will get a 256 bit block. After that we have to use parallel operation. But I have few doubts, firstly I am confused about how to do the widening multiplication? Apart from this after we do the widening multiplication we need to do saturated parallel addition and return the lower 128 bit block. If we want to do it like that we need to write new deinterleave operations as existing x_mm256_deinterleave_even returns a0,a2,a4,a6,b0,b2,b4,b6,a8,a10,a12,a14,b8,b10,b12,b14 and not a0,a2,a4,a6,a8,a10,a12,a14,b0,b2,b4,b6,b8,b10,b12,b14. So in my opinion if we have to use parallel operation we will need new parallel functions and return the `'lower 128  bit block' of mm256_adds_epi16( new_mm256_deinterleaveeven(x) , new_mm256_deinterleaveodd(x) ) where x is result of widening multiplication. Am I thinking in right direction? I am a little confused about this. Please tell me what I need to do for parallel impl of maddubs or if the current one is good enough?

nemequ:
I think that at the time I was under the impression that `hadd` would do the sensible thing instead of switching back and forth between a and b, in which case it would have just been a matter of

```c
return simde_mm256_hadd_epi16(
    simde_mm256_mul_epi16(simde_mm256_cvtepu8_epi16(a_.m128i[0]), simde_mm256_cvtepu8_epi16(b_.m128i[0])),
    simde_mm256_mul_epi16(simde_mm256_cvtepu8_epi16(a_.m128i[1]), simde_mm256_cvtepu8_epi16(b_.m128i[1])));
```

Unfortunately this is another example of Intel's APIs lacking any sort of consistency.  In this case, it's probably because they are trying to reuse hardware.

I'm actually thinking it might be a good idea to rename the existing simde_x_mm256_deinterleave* functions, since what they're doing isn't really deinterleaving but something rather more complicated.  They're still useful as *internal* APIs, but they're pretty useless as a public API.

Anyways, for now I think what you have is about as good as we're going to get.  A SIMDE_SHUFFLE_VECTOR-based implementation would be nice, but I think the fallbacks that will mainly be here is what is currently the SSSE3 fallback (I'm planning on changing it soon to be a fallback for all 128-bit ISA extensions), so it probably doesn't matter much.  I'll go ahead and merge this soon.

BTW, you keep using "parallel", I think what you mean is "pairwise" (as in operates on an adjacent pair of lanes).

masterchef2209:
> BTW, you keep using "parallel", I think what you mean is "pairwise" (as in operates on an adjacent pair of lanes).

yes, my bad, I meant pairwise.

PR #461, opened 2020-07-03
avx2: added impl for mm{,256}_maskload_epi{32,64}

masterchef2209:
avx2: added impl for mm{,256}_maskload_epi{32,64}


masterchef2209:
```mem_add``` for ```epi64``` functions must be long long for native call but int64_t is long. This is causing native-aliases to fail. What should be the correct way of handling this ?
Thank You

**Comment by nemequ** [comment]
You should use `SIMDE_CHECKED_REINTERPRET_CAST` to cast in the native aliases:

```c
#define _mm256_maskload_epi64(mem_addr, mask) simde_mm256_maskload_epi64(SIMDE_CHECKED_REINTERPRET_CAST(long long*, int64_t*, mem_addr), mask)
```

However, you might have to cast to a different type for some compilers.  For example, ICC and MSVC use `__int64*`, so

```c
#if defined(HEDLEY_INTEL_VERSION) || defined(HEDLEY_MSVC_VERSION)
  #define _mm256_maskload_epi64(mem_addr, mask) simde_mm256_maskload_epi64(SIMDE_CHECKED_REINTERPRET_CAST(__int64*, int64_t*, mem_addr), mask)
#else 
  #define _mm256_maskload_epi64(mem_addr, mask) simde_mm256_maskload_epi64(SIMDE_CHECKED_REINTERPRET_CAST(long long*, int64_t*, mem_addr), mask)
#endif
```

The checked cast allows us to accept either `long long*` or `int64_t*`, but still if you try to pass a different type such as `int16_t*` (at least in some configurations, sometimes it just degrades to a regular static cast that you can pass anything to).

Since we use `int64_t` in the test suite, you may have to add a cast there as well in order to get the real `_mm256_maskload_epi64` to accept the parameter; see [test/x86/sse.c:3224-3228](https://github.com/simd-everywhere/simde/blob/0b99a5d63413fb4ce7d786610ee811be210a6c63/test/x86/sse.c#L3224-L3228) for an example.

PR #463, opened 2020-07-04
avx2: added impl for mm{,256}_maskload_epi{32,64}

masterchef2209:
avx2: added impl for mm{,256}_maskload_epi{32,64}


masterchef2209:
I tried everything as suggested and I managed to pass the native aliases tests, only a little change in ```test_simde_mm{,256}_maskload_epi64``` was required

PR #464, opened 2020-07-04
svml: optimize simde_mm_cdfnorm_{ps,pd}

himanshi18037:


simde/x86/svml.h
```diff
@@ -3661,6 +3616,91 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+    simde__m128 p = simde_mm_set1_ps(SIMDE_FLOAT32_C(0.3275911));
+    simde__m128 one = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.0));
+
+    simde__m128 sign = a;
+
+    a = simde_mm_div_ps(simde_x_mm_abs_ps(a), simde_mm_sqrt_ps(simde_mm_set1_ps(SIMDE_FLOAT32_C(2.0))));
+
+    simde__m128 t = simde_mm_div_ps(one, simde_mm_add_ps(one, simde_mm_mul_ps(p, a)));
+    simde__m128 temp = simde_mm_add_ps(simde_mm_mul_ps(simde_mm_add_ps(simde_mm_mul_ps(simde_mm_add_ps(simde_mm_mul_ps(simde_mm_add_ps(simde_mm_mul_ps(a5, t), a4), t), a3), t), a2), t), a1);
+    simde__m128 y = simde_mm_sub_ps(one, simde_mm_mul_ps(simde_mm_exp_ps(simde_mm_mul_ps(a, (simde_x_mm_negate_ps(a)))),simde_mm_mul_ps(temp,t)));
```
nemequ, line 3640: You should break these up into multiple statements so they're easier to follow.  Looking at the original code, I would probably do something like

```c
simde__m128 y = simde_mm_mul_ps(a5, t);
y = simde_mm_add_ps(y, a4);
y = simde_mm_mul_ps(y, t);
y = simde_mm_add_ps(y, a3);
...
```

You can also do the same thing with the scalar version during debugging:

```c
double y = a5 * t;
y = y + a4;
y = y * t;
y = y + a3;
...
```

Once you have verified that the scalar version is correct, porting it to an explicitly vectorized version should be much more straightforward… it's basically an intermediate step that reduces the differences between the two versions so errors are easier to catch.

simde/x86/svml.h
```diff
@@ -3661,6 +3616,91 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+
+    simde__m128 r_ = simde_mm_mul_ps(simde_mm_set1_ps(SIMDE_FLOAT32_C(0.5)), simde_mm_add_ps(one, simde_x_mm_xorsign_ps(sign, y)));
```
nemequ, line 3642: The `simde_x_mm_xorsign_ps` call is wrong; the arguments are reversed.

simde/x86/svml.h
```diff
@@ -3661,6 +3616,91 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+
+    simde__m128 a1 = simde_mm_set1_ps(SIMDE_FLOAT32_C( 0.254829592));
+    simde__m128 a2 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-0.284496736));
+    simde__m128 a3 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.421413741));
+    simde__m128 a4 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-1.453152027));
+    simde__m128 a5 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.061405429));
+    simde__m128 p = simde_mm_set1_ps(SIMDE_FLOAT32_C(0.3275911));
+    simde__m128 one = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.0));
+
+    simde__m128 sign = a;
```
nemequ, line 3634: Copying a to a variable called "sign" here is misleading; you're copying the entire value, not just the sign.  You could choose another variable name, but my suggestion would actually be leave `a` untouched during the entire function, and instead of assigning to `a` create a new `x` variable.  That would match what the `simde_math_cdfnormf` function does.

simde/x86/svml.h
```diff
@@ -3661,6 +3616,91 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+    return _mm_cdfnorm_ps(a);
+  #elif (SIMDE_NATURAL_VECTOR_SIZE > 0)
+
+    simde__m128 a1 = simde_mm_set1_ps(SIMDE_FLOAT32_C( 0.254829592));
+    simde__m128 a2 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-0.284496736));
+    simde__m128 a3 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.421413741));
+    simde__m128 a4 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-1.453152027));
+    simde__m128 a5 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.061405429));
+    simde__m128 p = simde_mm_set1_ps(SIMDE_FLOAT32_C(0.3275911));
+    simde__m128 one = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.0));
```
nemequ, line 3632: These should really all be `static const`.  TBH it probably doesn't matter from the compiler's perspective since it should be able to determine that you never touch the variables anyways, but it also serves as documentation for human readers, and lets the compiler provide you with a bit of protection; if you accidentally try to assign to a `const` variable the compiler will emit an error.  It's just a good habit to get into.

simde/x86/svml.h
himanshi18037: I copied a to a function x and performed operations on it.

simde/x86/svml.h
himanshi18037: This was the one causing failures:(

simde/x86/svml.h
```diff
@@ -3661,6 +3616,110 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+
+    const simde__m128 a1 = simde_mm_set1_ps(SIMDE_FLOAT32_C( 0.254829592));
+    const simde__m128 a2 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-0.284496736));
+    const simde__m128 a3 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.421413741));
+    const simde__m128 a4 = simde_mm_set1_ps(SIMDE_FLOAT32_C(-1.453152027));
+    const simde__m128 a5 = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.061405429));
+    const simde__m128 p = simde_mm_set1_ps(SIMDE_FLOAT32_C(0.3275911));
+    const simde__m128 one = simde_mm_set1_ps(SIMDE_FLOAT32_C(1.0));
+
+    simde__m128 x = a;
```
nemequ, line 3634: This is unnecassary; you write to x only to overwrite it in the next statement.  Just do `simde__m128 x = simde_mm_div_ps(...);`

simde/x86/svml.h
```diff
@@ -3661,6 +3616,110 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+    y = simde_mm_mul_ps(y, t);
+    y = simde_mm_add_ps(y, a3);
+    y = simde_mm_mul_ps(y, t);
+    y = simde_mm_add_ps(y, a2);
+    y = simde_mm_mul_ps(y, t);
+    y = simde_mm_add_ps(y, a1);
+    y = simde_mm_mul_ps(y, t);
+    simde__m128 ex = simde_mm_exp_ps(simde_mm_mul_ps(x, simde_x_mm_negate_ps(x)));
+    y = simde_mm_mul_ps(y, ex);
+    y = simde_mm_sub_ps(one, y);
```
nemequ, line 3647: This looks much better!  I suggest you copy the code from the scalar version and put it in comments, though… something like

```c
    /* fabs(x) / sqrt(2.0); */
    x = simde_mm_div_ps(simde_x_mm_abs_ps(x), simde_mm_sqrt_ps(simde_mm_set1_ps(SIMDE_FLOAT32_C(2.0))));

    /* 1.0 / (1.0 + p * x) */
    simde__m128 t = simde_mm_div_ps(one, simde_mm_add_ps(one, simde_mm_mul_ps(p, x)));

    /*  1.0 - (((((a5 * t + a4) * t) + a3) * t + a2) * t + a1) * t * exp(-x * x)*/
    simde__m128 y = simde_mm_mul_ps(a5, t);
    y = simde_mm_add_ps(y, a4);
    y = simde_mm_mul_ps(y, t);
    y = simde_mm_add_ps(y, a3);
    y = simde_mm_mul_ps(y, t);
    y = simde_mm_add_ps(y, a2);
    y = simde_mm_mul_ps(y, t);
    y = simde_mm_add_ps(y, a1);
    y = simde_mm_mul_ps(y, t);
    y = simde_mm_mul_ps(y, simde_mm_exp_ps(simde_mm_mul_ps(x, simde_x_mm_negate_ps(x))));
    y = simde_mm_sub_ps(one, y);
```

simde/x86/svml.h
```diff
@@ -3661,6 +3616,110 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+
+    simde__m128 r_ = simde_mm_mul_ps(simde_mm_set1_ps(SIMDE_FLOAT32_C(0.5)), simde_mm_add_ps(one, simde_x_mm_xorsign_ps(y, a)));
```
nemequ, line 3649: Here, too.  Just add a comment with the scalar version.

simde/x86/svml.h
```diff
@@ -3661,6 +3616,110 @@ simde_mm512_mask_exp_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
   #define _mm512_mask_exp_pd(src, k, a) simde_mm512_mask_exp_pd(src, k, a)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128
+simde_mm_cdfnorm_ps (simde__m128 a) {
+  #if defined(SIMDE_X86_SVML_NATIVE) && defined(SIMDE_X86_SSE_NATIVE)
+    return _mm_cdfnorm_ps(a);
+  #elif (SIMDE_NATURAL_VECTOR_SIZE > 0)
+
```
nemequ, line 3625: You don't need a blank line here.  Though you should add a link to https://www.johndcook.com/blog/cpp_phi/.

simde/x86/svml.h
himanshi18037: Updated.

PR #466, opened 2020-07-05
svml: optimize {mm256,mm512}_cdfnorm_{ps,pd}

himanshi18037:


nemequ:
Looks good, thanks!

I'm going to merge this as it is, but if you want a little speed bost a good optimization might be to use FMA instructions here instead of separate multiply + add.

PR #467, opened 2020-07-05
svml: optimize mm_cdfnorminv_{ps,pd}

himanshi18037:


simde/x86/svml.h
```diff
@@ -4768,13 +4722,213 @@ simde_mm512_mask_log_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+
+    const simde__m128 low  = simde_mm_set1_ps(SIMDE_FLOAT32_C( 0.02425));
+    const simde__m128 high = simde_mm_set1_ps(SIMDE_FLOAT32_C( 0.97575));
+
+    simde__m128 q, r;
+
+    simde__m128 matched;
+
+    /* if (a < 0 || a > 1) */
+    simde__m128 m1 = simde_mm_or_ps(simde_mm_cmplt_ps(a, simde_mm_set1_ps(SIMDE_FLOAT32_C(0.0))), simde_mm_cmpnlt_ps(a, simde_mm_set1_ps(SIMDE_FLOAT32_C(1.0))));
```
nemequ, line 4768: You should use `simde_mm_cmpgt_ps` instead of `simde_mm_cmpnlt_ps`.  The latter would be the equivalent of `>=` not `>` (though you could also use `simde_mm_cmpge`).

simde/x86/svml.h
```diff
@@ -4768,13 +4722,213 @@ simde_mm512_mask_log_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+    matched = m1;
+
+    /* if (a < low) */
+    simde__m128 m2 = simde_mm_andnot_ps(matched, simde_mm_min_ps(a, low));
```
nemequ, line 4772: You have `simde_mm_min_ps`, but that return the lower value for each lane of the inputs.  Based on the comment above, what you want is `simde_mm_cmplt_ps` which returns a mask of lanes where the first argument (a) is less than the second (low).

In other words, max is `(a < low) ? a : low`.  cmplt is `(a < low) ? ~0 : 0`.

simde/x86/svml.h
```diff
@@ -4768,13 +4722,213 @@ simde_mm512_mask_log_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
...
+    matched = simde_mm_or_ps(matched, m2);
+
+    /* if (a > high) */
+    simde__m128 m3 = simde_mm_andnot_ps(matched, simde_mm_max_ps(a, high));
```
nemequ, line 4776: Same thing here; you want `simde_mm_cmpgt_ps` not `simde_mm_max_ps`.

PR #473, opened 2020-07-07
neon: implementation of qadd

seanptmaher:
This one's got a lot of stuff I was unsure how to implement. 

For unsigned saturation, I think it's all good, but for signed saturation, I think this could be done better. I use casts to saturate in other intrinsics, and I'm not sure if that's the wrong way to do things (it might, but in reality, you're just doing a `movse`, but it stops proper vectorization).

Also, I'm unsure why, but on my x86 machine, SIMDE_NATURAL_VECTOR_SIZE is not being defined to some number, which is not ideal. `lscpu` says I support avx, so, I'm unsure what's going on there. 

Anyways, this patch might require some light rework. 

(CC @tlively @ngzhian)


nemequ:
I spent some time a while back trying to find a decent way to implement saturated math, and the conclusion I came to is that compilers just can't handle it well, and the only option is to take advantage of platform-specific intrinsics.

I've merged this and added a few additional implementations which should help improve things a bit, it's working its way through CI now.

Thanks for the patch!

> Also, I'm unsure why, but on my x86 machine, SIMDE_NATURAL_VECTOR_SIZE is not being defined to some number, which is not ideal. `lscpu` says I support avx, so, I'm unsure what's going on there.

You're probably looking at it when the emul tests are being compiled.  When `SIMDE_NO_NATIVE` is no native instructions are used, so `SIMDE_NATURAL_VECTOR_SIZE` is 0.

PR #477, opened 2020-07-08
avx2: added impl for mm{,256}_{,mask}_i{32,64}gather_epi32 and small bug fix

masterchef2209:
I was using the wrong scale factor, scale factor should have been 4 instead of 1 to test with int32_t buffer array and I didn't understand at first that this instruction uses byte addressing.


masterchef2209:
fixed this, it is hopefully fine now

PR #479, opened 2020-07-09
neon: Implementation of qshl

seanptmaher:
I didn't really create optimized versions of the instruction, but other platforms don't support saturating shifting (at least, not to my knowledge), so I'm not too worried about it. 

Feel free to change up the style, the way I do it right now was the one which was easiest to debug for me, but it could be written slightly differently. 

Thanks

(CC @tlively @ngzhian)


simde/arm/neon/qshl.h
```diff
@@ -0,0 +1,812 @@
...
+      a_ = simde_int8x8_to_private(a),
+      b_ = simde_int8x8_to_private(b);
+
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
+      int8_t b_value = HEDLEY_STATIC_CAST(int8_t, b_.values[i] & 0xff); // HEDLEY_STATIC_CAST(int8_t, b_.values[i]);
+      uint8_t mask = UINT8_MAX << SIMDE_MIN_((7 - b_value), 8);
+      if (b_value > 0) {
+        if (b_value >= 8) {
+          if (a_.values[i]) {
```
ngzhian, line 56: The overall logic is a bit confusing for me. E.g. I don't get why we need to check `if (a_.values[i])`.
If `b_value is >= 8` then we can immediate set `r_.values[i]` to be 0, can't we?

Not trying to add more work for you, maybe you can consider implementing sshl first, then you qshl can be sshl + saturate, which may simpilfy the logic here a bit.
See https://git.linaro.org/arm/vixl.git/tree/src/aarch64/simulator-aarch64.cc#n5301

seanptmaher: If `b_value is >= 8`, then either we're saturating, or the value was 0 to begin with, so I've gotta make another check.

seanptmaher:
@nemequ I saw you thumbs-up-ed Zhi's comment, do you want me to do `sshl` first? I'm pretty sure that this is probably close to as direct of an implementation (as in, without fluff/extraneous checks) that I can make, and it passes all the tests, so I'm not exactly in a rush to reimplement it, but if it'd be better, I'm not against it, either.

nemequ:
I know this isn't going to be a very satisfying answer, but I'm not sure what I want to do here yet.  I feel like the current implementation is way more complicated than it needs to be, but I need to spend some time with it and test some stuff to see how compilers handle it… sometimes they do a really good job of auto-vectorizing stuff like this, and if you try help and optimize it a bit by hand you can make things a lot worse.

If it's blocking you on something I can go ahead and merge it as-is, then come back to it later.  If not, I'll try to get to this soon (likely later this week).

seanptmaher:
I don't think I'm blocked, no rush.

aqrit:
PR superseded by [this commit](https://github.com/simd-everywhere/simde/commit/b266b2b65b87d4529576bca33bd76d1fb3fa84b4)?

nemequ:
Oops, sorry Sean :(

I'll take a look at this again soon to see which version performs better; if this one does I'll replace the current one with it.

seanptmaher:
Don't give it too much trouble, a glance makes me nearly certain that yours is superior; I think mine might be pathological wrt the branch predictor.

Either way, or if I'm wrong now, feel free but don't feel forced to.

PR #481, opened 2020-07-10
Meson packaging

SoapGentoo:
@mr-c some canonical Meson cleanups, in preparation for your PR in https://github.com/PacificBiosciences/pbcopper/pull/11


mr-c:
@SoapGentoo Super cool! I'll let @nemequ review this one.

meson.build
```diff
@@ -6,31 +6,132 @@ project('SIMDe', 'c', 'cpp',
...
-		'simde/x86/mmx.h',
-		'simde/x86/sse.h',
-		'simde/x86/sse2.h',
-		'simde/x86/sse3.h',
-		'simde/x86/sse4.1.h',
-		'simde/x86/sse4.2.h',
-		'simde/x86/ssse3.h',
-		'simde/x86/svml.h',
-    ],
+if not meson.is_subproject() and get_option('tests')
```
nemequ, line 9: If it's being built as a subproject the tests won't be build, so I'm not sure we need an option here.  It's not like we're building a shared library or something; the *only* thing to build is the tests, so why would anyone want to "build" SIMDe without building the tests?

The only thing I can think of is that you could install the headers without building the tests, but that doesn't seem particularly useful…  Do you need this feature, or is it just the pattern you use for Meson projects?

nemequ: Oh, nevermind, I just realized: Gentoo.  I guess you want to have a simde package which doesn't actually build anything and only installs to cut down on build times.

SoapGentoo: Correct, we need to be able to disable tests. 95% of Gentoo users don't have tests enabled.

nemequ:
Merged, thanks!

PR #485, opened 2020-07-11
avx2: added impl for mm{,256}_{,mask}_i{32,64}gather_epi64

masterchef2209:
avx2: added impl for mm{,256}_{,mask}_i{32,64}gather_epi64


nemequ:
LGTM, thanks!  I'll merge as soon as it gets through CI.

mr-c:
@masterchef2209 So close to getting avx2 finished, thank you!

masterchef2209:
I could do it only because of the guidance from the mentors.
Thank You :)
@mr-c

PR #486, opened 2020-07-12
neon: Add rshl

rosbif:
Add vrshl and vrshlq intrinsics.


rosbif:
I don't understand the  `Meson install headers` failure ;-(

PR #487, opened 2020-07-13
avx2: added impl for mm{,256}_{,mask}_i{32,64}gather_p{s,d}

masterchef2209:
avx2: added impl for mm{,256}_{,mask}_i{32,64}gather_p{s,d}


nemequ:
Merged, thanks!

PR #488, opened 2020-07-13
neon: implementation of ext

seanptmaher:
Hello, this seems like a function that might be implementable using `SIMDE_SHUFFLE_VECTOR_`, but I couldn't think of a way to do it without having to use a `CONSTIFY`- style case statement (because depending on the lane you passed, you'd pass arguments like `0 -> (0), 1 -> (0, 1), 2 -> (0, 1, 2)` (not those values specifically, but it illustrates the issue I see). Thoughts on that? 

It's still functional as is, but I think it might be slightly more efficient that way. (It wouldn't use `simde_private` types, then, and it'd probably be a lot easier to replace with a shuffle on various platforms, I wouldn't be surprised if the autovectorizer failed to catch it as is)

(CC @ngzhian @tlively)


nemequ:
`SIMDE_SHUFFLE_VECTOR_` would indeed be a good way to implement this for platforms which support it.  I can't think of a better way to use a constify macro than what you have, but at that point I think it would be better to just implement the whole thing as a macro when `SIMDE_SHUFFLE_VECTOR_` is available.

On most platforms, though, it seems like this could be better implemented with shifts.  POWER has `vec_sld`, which I think is exactly what this function does (except you have to multiply n by sizeof(element_type)).  For SSE2, `_mm_or_si128(_mm_bslli_si128(a, n), _mm_bsrli_si128(b, elem_count - n))` (though that might be a bit tricky to get working with constify).  I'll try to play around with that a bit.

You're right that the current implementation is too complicated to rely on autovectorization.  Going through the private types doesn't really concern me (compilers are really good at optimizing that away), but having two separate loops is a bit much, and even if you combined them compilers tend to be a bit iffy at autovectorizing shuffles.

nemequ:
How about something like this:

```c
SIMDE_FUNCTION_ATTRIBUTES
simde_int8x8_t
simde_vext_s8(simde_int8x8_t a, simde_int8x8_t b, const int n)
    SIMDE_REQUIRE_CONSTANT_RANGE(n, 0, 7) {
    simde_int8x8_t r;
    SIMDE_CONSTIFY_8_(vext_s8, r, (HEDLEY_UNREACHABLE(), a), n, a, b);
    return r;
  #else
    simde_int8x8_private
      a_ = simde_int8x8_to_private(a),
      b_ = simde_int8x8_to_private(b),
      r_;
    const size_t n_ = HEDLEY_STATIC_CAST(size_t, n);

      size_t src = i + n_;
      r_.values[i] = (src < (sizeof(r_.values) / sizeof(r_.values[0]))) ? a_.values[src] : b_.values[src & 7];
    }

    return simde_int8x8_from_private(r_);
  #endif
}
#if defined(SIMDE_SHUFFLE_VECTOR_) && \
    !defined(SIMDE_ARM_NEON_A32V7_NATIVE)
  #define simde_vext_s8(a, b, n) (__extension__ ({ \
      const int simde_vext_s8_n = HEDLEY_STATIC_CAST(int8_t, n); \
      simde_int8x8_private simde_vext_s8_r_; \
      simde_vext_s8_r_.values = SIMDE_SHUFFLE_VECTOR_(8, 8, simde_int8x8_to_private(a).values, simde_int8x8_to_private(b).values, \
        HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 0), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 1), \
        HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 2), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 3), \
        HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 4), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 5), \
        HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 6), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 7)); \
      simde_int8x8_from_private(simde_vext_s8_r_); \
    }))
#endif
#if defined(SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES)
  #undef vext_s8
  #define vext_s8(a, b, n) simde_vext_s8((a), (b), (n))
#endif
```

?

seanptmaher:
How's that look?

seanptmaher:
Hm, the SDE server says it needs to be a constant integer (I assume that means an immediate integer constant, the passed integer is already declared `const`) -- Can we fix this with a constify-style thing? (I can't immediately see how that would work, seeing as we're doing an addition, or is an immediate addition foldable by the compiler?)

nemequ:
Here is my understanding:

Immediate really comes from the encoding of the instruction; the values are actually encoded into the instruction itself as opposed to referencing a register.

At the C level, what compilers seem to require for most the immediate mode parameter of intrinsics are integer constant expressions, which is defined by the C standard (§ 6.6 p6 of C18, but IIRC it goes back to at least C99):

> An integer constant expression shall have integer type and shall only have operands that are integer constants, enumeration constants, character constants, sizeof expressions whose results are integer constants, _Alignof expressions, and floating constants that are the immediate operands of casts. Cast operators in an integer constant expression shall only convert arithmetic types to integer types, except as part of an operand to the sizeof or _Alignof operator.

Then there is something that the compiler is able to infer the value of at compile time and can therefore perform constant folding. Basically, something for which `__builtin_constant_p` will return true.

> Can we fix this with a constify-style thing? (I can't immediately see how that would work, seeing as we're doing an addition, or is an immediate addition foldable by the compiler?)

I don't see a straightforward solution with a constify macro, either :(

I don't think the problem is the addition, but rather using a variable.  In the macro we could resolve this by just getting rid of the temporary variable and passing `n + 0`, `n + 1`, `n + 2`, etc. to `SIMDE_SHUFFLE_VECTOR_`, but that may not even be necessary (it depends on whether clang is insisting on an ICE or just constant in the sense of `__builtin_constant_p`).  The bigger problem is that the tests use variables, too, and those aren't trivial for the compiler to discard.  To work around that we would have to hard code different values for each test.

However, if clang only requires a `__builtin_constant_p`-style constant, something like this could do the trick:

```c
#if defined(SIMDE_SHUFFLE_VECTOR_) && \
    !defined(SIMDE_ARM_NEON_A32V7_NATIVE)
  #define simde_vext_s8(a, b, n) (__extension__ ({ \
      simde_int8x8_t simde_vext_s8_r;
      if (!__builtin_constant_p(n)) { \
        simde_vext_s8_r = simde_vext_s8(a, b, n); \
      } else { \
        const int simde_vext_s8_n = HEDLEY_STATIC_CAST(int8_t, n); \
        simde_int8x8_private simde_vext_s8_r_; \
        simde_vext_s8_r_.values = SIMDE_SHUFFLE_VECTOR_(8, 8, simde_int8x8_to_private(a).values, simde_int8x8_to_private(b).values, \
          HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 0), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 1), \
          HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 2), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 3), \
          HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 4), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 5), \
          HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 6), HEDLEY_STATIC_CAST(int8_t, simde_vext_s8_n + 7)); \
        simde_vext_s8_r = simde_int8x8_from_private(simde_vext_s8_r_); \
      } \
      simde_vext_s8_r; \
    }))
#endif
```

If you want to just get this in and worry about a clang version later, just changing the `defined(SIMDE_SHUFFLE_VECTOR_)` test to `defined(SIMDE_SHUFFLE_VECTOR_) && !defined(__clang__)` would do the trick.  If the `if(__builtin_constant_p)` thing doesn't work I'd be okay with that, then we can re-work the tests and remove the `&& !defined(__clang__)` later.

seanptmaher:
Looks like it didn't like that, so I suppose we'll have to `!defined(__clang__)` for now. 

Slight curiosity, I noticed you used `(__extension__ ([...]` to implement the macros, and as far as I know, `__extension__` is just a way to use compiler extensions without getting warnings, but I don't really see you using any in here. Why did you add the `__extension__`? I wouldn't be surprised if I was missing something obvious, but

nemequ:
Okay, I'll go ahead and merge this, and I'll try to redo the tests soon.

> Slight curiosity, I noticed you used `(__extension__ ([...]` to implement the macros, and as far as I know, `__extension__` is just a way to use compiler extensions without getting warnings, but I don't really see you using any in here. Why did you add the `__extension__`? I wouldn't be surprised if I was missing something obvious.

[Statement expressions](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) are an extension.  They're supported by at least GCC, clang, TinyCC, ICC, PGI, suncc, and xlc, but not by at least MSVC, Pelles, or DMC (I don't think by IAR either, but I'd have to check).  Luckily every compiler than supports __builtin_shuffle or __builtin_shufflevector supports them.

PR #491, opened 2020-07-14
neon: implementation of dot

seanptmaher:
Hello,

This one's got a problem that I'm unsure how to fix, passes most of the tests on my Travis, but it's throwing me an error about inlining on `aarch64`, but not on `armv8-a`.

```c
usr/lib/gcc/aarch64-linux-gnu/7/include/arm_neon.h:31565:1: error: inlining failed in call to always_inline ‘vdot_s32’: target specific option mismatch

 vdot_s32 (int32x2_t __r, int8x8_t __a, int8x8_t __b)

 ^~~~~~~~

In file included from ../arm/neon/dot.c:4:0:

../arm/neon/../../../simde/arm/neon/dot.h:46:12: note: called from here

     return vdot_s32(a, b, c);

            ^~~~~~~~~~~~~~~~~
```
It should be good otherwise, any thoughts are welcome. 

A quick google search seems to point to the proper compiler flags not being enabled, but I check for `SIMDE[...]A32V8[..]`, which is what it's supposed to require in [the documentation here](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=dot). Is there a mistake in the `#define`s? 

(CC @tlively @ngzhian)


nemequ:
vdot isn't really part of the core of AArch64, it's an extension.  You can check for support with `defined(__ARM_FEATURE_DOTPROD)` (in addition to checking for AArch64 as you already are).

nemequ:
Here is an implementation which should be pretty quick on ARM w/o dotprod: `simde_vadd_s32(a, simde_vpaddl_s16(simde_vmovn_s32(simde_vpaddlq_s16(simde_vmull_s8(b, c)))))`

For the portable versions, clang does really well with what you have, but this is about twice as fast on GCC (and very slightly slower on clang):

```c
    simde_int32x2_private
      a_ = simde_int32x2_to_private(a),
      r_;
    simde_int8x8_private
      b_ = simde_int8x8_to_private(b),
      c_ = simde_int8x8_to_private(c);

    for (int i = 0 ; i < 2 ; i++) {
      int32_t acc = 0;
      SIMDE_VECTORIZE_REDUCTION(+:acc)
      for (int j = 0 ; j < 4 ; j++) {
        const int idx = j + (i << 2);
        acc += HEDLEY_STATIC_CAST(int32_t, b_.values[idx]) * HEDLEY_STATIC_CAST(int32_t, c_.values[idx]);
      }
      r_.values[i] = acc * a_.values[i];
    }

    return simde_int32x2_from_private(r_);
```

PR #497, opened 2020-07-16
neon: Correct SSE translations of tbl3 and tbl4

rosbif:
Correct SSE translations of tbl3 and tbl4, sorry.


nemequ:
Did you push the tbx patch on accident?

I'm merging the tbl fixes (working through CI now), but it looks like tbx isn't working yet so I'll hold off on that.

rosbif:
Sorry I thought tbx was working as it passed local tests on x86, ARM and PPC64BE. But unfortunately I hadn't noticed that on x86 I was still testing tbl not tbx :-(
I also had some native alias errors which I couldn't find. They all look OK to me. Other than this tbx is now working (I hope).
I shall push it in another branch and PR it if it is OK.

nemequ:
NP; I just figured it was an accident since the PR title didn't mention it.

I've made the same mistake with the CI builds.  I really wish they were faster, but it's hard to complain given that they're free… Maybe it's time to add another provider or two to shift the burden off of Travis a bit more.  Azure Pipelines hasn't been doing anything since we moved the repo, and it's pretty snappy.

PR #501, opened 2020-07-16
Add .dir-locals.el

seanptmaher:
This is pretty useful, makes sure the offset is 2, shows trailing whitespace, should stop me (and anyone else using emacs) from failing the formatting bot, indent with spaces, etc.

It's not really super necessary, but it's helpful, a "convenience merge," if you will.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,4 @@
+((nil . ((indent-tabs-mode . nil)
+         (show-trailing-whitespace . t)))
+ (c-mode . ((c-basic-offset . 2)))
+ (c++-mode . ((c-basic-offset . 2))))


nemequ:
I don't really have a problem merging this, but it might be better to use [editorconfig-emacs](https://github.com/editorconfig/editorconfig-emacs).  `.editorconfig` not being supported by default is annoying, but adding different configuration files for emacs, vim, VS Code, VS, CLion, Sublime Text, etc. isn't a a very good solution either.

Does anyone else have an opinion on at least trying to add tons of editor-specific configurations vs. just encouraging people to install the editorconfig plugin for their editor of choice?

seanptmaher:
Honestly, I don't mind either way, this is a good solution too. I actually didn't notice that `.editorconfig` existed... I probably wouldn't have written this PR if I did...

nemequ:
I should probably add a note to CONTRIBUTING.md; I think a lot of people don't know about it… I had no idea it was a thing until Michael suggested it a while back.

If that's good with you, I think I'll just close this without merging, but please feel free to file issues like this in the future.  If it weren't for EditorConfig I'd definitely want to merge this.  Just among the active contributors I know Christopher uses emacs as well, but I use VS Code, and Hidayat uses CLion (not sure about anyone else).  Adding editor-specific files for each of those is already a bit much, then you have other people using vim, Sublime Text, VS, XCode, Atom, GNOME Builder… it just seems like a rabit hole I'm not particularly keen on going down.

If anyone wants to try to change my mind feel free to re-open.  I don't feel particularly strongly about this, so if you have a decent argument your odds are good.

seanptmaher:
Sounds good. Closing...

PR #503, opened 2020-07-16
neon: implementation of rhadd

seanptmaher:
There's a failure in ICC. I'm still looking into why it's failing, but it's failing many of the tests. 

Works on all other compilers and platforms, though, I believe... 

(CC @tlively @ngzhian)


simde/arm/neon/rhadd.h
```diff
@@ -0,0 +1,406 @@
...
+simde_vrhadd_s8(simde_int8x8_t a, simde_int8x8_t b) {
+  #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
+    return vrhadd_s8(a, b);
+  #else
+    simde_int8x8_private
+      r_,
+      a_ = simde_int8x8_to_private(a),
+      b_ = simde_int8x8_to_private(b);
+
+    #if defined(SIMDE_VECTOR_SUBSCRIPT_OPS) && !defined(__clang__)
```
nemequ, line 48: Why ` && !defined(__clang__)`?

seanptmaher: Clang's got a warning, it's not letting me do shifts or bit operations with integers (even with casted integers)

seanptmaher: Full disclosure, I could be doing something wrong there, but I can show you compile output in a sec

seanptmaher:
```c
../arm/neon/../../../simde/arm/neon/rhadd.h:49:32: error: cannot convert between vector values of different size ('__attribute__((__vector_size__(8 * sizeof(int8_t)))) int8_t' (vector of 8 'int8_t' values) and 'int8_t' (aka 'signed char'))

      r_.values = (((a_.values >> HEDLEY_STATIC_CAST(int8_t, 1)) + (b_.values >> HEDLEY_STATIC_CAST(int8_t, 1))) + ((a_.values | b_.values) & HEDLEY_STATIC_CAST(int8_t, 1)));

                     ~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

seanptmaher: Hm, I misread this earlier actually, I think... However, I'm still not actually 100% sure how best to do this with the subscript ops, do I have to create a vector of ones and shift it by that?

nemequ: There is a `SIMDE_VECTOR_SUBSCRIPT_SCALAR` instead of `SIMDE_VECTOR_SUBSCRIPT_OPS` for detecting whether the compiler supports that.

What version of clang are you using, though?  GCC has supported it since 4.8.0, and I don't even have a clang check on there but we test back to 3.8 and it works.  I just tried removing the ` && !defined(__clang__)` on this patch and it works on clang 10…

seanptmaher: Hm, it was giving me the error on clang 3.8. I had assumed it was a clang thing. Sounds like the right thing would be to sub that out to `SIMDE_VECTOR_SUBSCRIPT_SCALAR`, then.

nemequ: Sorry, I really need to document the macros in simde-common.h better.  There is some really useful stuff in there, but it's a bit hard to figure out what everything does…

It turns out we did actually have a test for clang, but we were effectively testing for version 4 and it [looks like](https://godbolt.org/z/eKKonG) support was added sometime between clang 4 and 5 (4.0.1 doesn't work, 5.0.0 does), so I'll adjust the check to use `SIMDE_DETECT_CLANG_VERSION_CHECK(5,0,0)`.

nemequ:
Sorry, it looks like I forgot to close this, but it was merged as dde1eca8a013b98570cbc6d8b7d43607ae7d2d1c, thanks.

PR #510, opened 2020-07-19
avx512bw,sse4.1,sse4.2: added impl for mm512_sra{,i,v}_epi16 + NEON impl for mm_testz_si128 and CRC intrinisics

masterchef2209:
I had to add ```+crc``` to ```-march=armv8-a+simd``` flag for testing NEON CRC code. We may need to add this flag to some of the travis tests but I am not sure.


simde/x86/sse4.2.h
```diff
@@ -275,15 +275,22 @@ simde_mm_crc32_u8(uint32_t prevcrc, uint8_t v) {
...
-    crc ^= v;
-    for(int bit = 0 ; bit < 8 ; bit++) {
-      if (crc & 1)
-        crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-      else
-        crc = (crc >> 1);
-    }
-    return crc;
+    #if defined(SIMDE_ARM_NEON_A64V8_NATIVE) && defined(__ARM_FEATURE_CRC32)
+      __asm__ __volatile__("crc32cb %w[c], %w[c], %w[v]\n\t"
```
nemequ, line 279: I'd like to avoid inline assembly.  Obviously this is GCC-style inline assembly, so it won't work in MSVC (which yes, can target ARM).  The __volatile__ is also a barrier for optimizations.

There are intrinsics for this in `arm_acle.h`.  If the compiler supports ACLE `__ARM_ACLE` should be defined.  However, GCC doesn't yet implement all of ACLE, though it does have the __crc32* intrinsics, so you can do something like

```c
#if defined(__ARM_ACLE) || (defined(__GNUC__) && defined(__ARM_FEATURE_CRC32))
  #include <arm_acle.h>
#endif
```

Then instead of using inline assembly here you can use the `__crc32cb` function.

masterchef2209: I have tried to do this, but for ```mm_crc32_u64``` __crc32cx is not available do I need to include some other header file for accessing that function?
Thank You

masterchef2209: fixed this

simde/x86/avx512bw.h
```diff
@@ -1895,6 +1895,101 @@ simde_mm512_packus_epi32 (simde__m512i a, simde__m512i b) {
...
+simde__m512i
+simde_mm512_sra_epi16 (simde__m512i a, simde__m128i count) {
+  #if defined(SIMDE_X86_AVX512BW_NATIVE)
+    return _mm512_sra_epi16(a, count);
+  #else
+    simde__m512i_private
+      r_,
+      a_ = simde__m512i_to_private(a);
+
+    #if SIMDE_NATURAL_VECTOR_SIZE_LE(128)
```
nemequ, line 1908:
```c
    #if SIMDE_NATURAL_VECTOR_SIZE_LE(256)
      for (size_t i = 0 ; i < (sizeof(r_.m256i) / sizeof(r_.m256i[0])) ; i++) {
        r_.m128i[0] = simde_mm_sra_epi16(a_.m128i[i], count);
      }
```

masterchef2209: fixed this

simde/x86/sse4.2.h
```diff
@@ -275,15 +279,19 @@ simde_mm_crc32_u8(uint32_t prevcrc, uint8_t v) {
...
-    uint32_t crc = prevcrc;
-    crc ^= v;
-    for(int bit = 0 ; bit < 8 ; bit++) {
-      if (crc & 1)
-        crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-      else
-        crc = (crc >> 1);
-    }
-    return crc;
+    #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
```
nemequ, line 282: You also need to check `defined(__ARM_FEATURE_CRC32)` before calling the `__crc32*`, not just before including the header.  If `__ARM_FEATURE_CRC32` isn't set some compilers will not define the functions, which is why this is still giving an error on the AArch64 builds on Drone and Travis.

You should probably also add at least one AArch64 build to include CRC32 support (`-march=armv8-a+simd+crc` instead of just `-march=armv8-a+simd`).  See https://github.com/simd-everywhere/simde/blob/de8b41f876cf44a184e16ec57e4ed0d5c49f7944/.travis.yml#L75

masterchef2209: fixed this

nemequ:
Just FYI, you should be able to restart builds on Travis. Just log in using your GitHub account; for each job there should be a button all the way on the right to reset that specific job, or you can trigger a rebuild of the whole build.

masterchef2209:
> Just FYI, you should be able to restart builds on Travis. Just log in using your GitHub account; for each job there should be a button all the way on the right to reset that specific job, or you can trigger a rebuild of the whole build.

I restarted build a couple of times for this PR, but for some reason, it kept failing at the same aarch64 test, I read the job log it showed "ninja-build" error so I thought something was wrong.

nemequ:
Okay, just wanted to make sure you knew you could do it.  I restarted it and it worked, so it was probably some sort of temporary issue (likely a problem with some dependency repository).

PR #511, opened 2020-07-20
avx512bw,sse4.1: added impl for mm512_sad_epu8, mm512_srlv_epi16 + NEON impl of mm_test* functions

masterchef2209:


nemequ:
In general you should try to limit PRs to similar patches when feasible.  For example, this should probably be one PR for the NEON stuff and one for the AVX-512BW stuff.

Honestly I don't mind if you do it this way, but other open-source projects may not be as accepting, so it it would probably be a good habit to get into.

simde/x86/sse4.1.h
```diff
@@ -1902,11 +1902,17 @@ simde_mm_test_mix_ones_zeros (simde__m128i a, simde__m128i mask) {
     a_ = simde__m128i_to_private(a),
     mask_ = simde__m128i_to_private(mask);
 
-  for (size_t i = 0 ; i < (sizeof(a_.u64) / sizeof(a_.u64[0])) ; i++)
-    if (((a_.u64[i] & mask_.u64[i]) != 0) && ((~a_.u64[i] & mask_.u64[i]) != 0))
-      return 1;
+  #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
+    int64x2_t s640 = vandq_s64(a_.neon_i64, mask_.neon_i64);
+    int64x2_t s641 = vandq_s64(~a_.neon_i64, mask_.neon_i64);
```
nemequ, line 1907: `~a_.neon_i64` is actually [a GCC extension](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html); with normal NEON you can't use the NOT operator on a vector.

What you want here is `vmvnq_s64`, but of course that doesn't exist (which is surprising, NEON is normally really good about that type of thing).  Luckily since this is just a simple bitwise operation the types don't really matter, so you can do something like

```c
int64x2_t s641 = vandq_s64(vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s664(a_.neon_i64))), mask_.neon_i64)
```

It's a bit ugly, but it works :)

masterchef2209: I see, I will fix this.
Thank You

masterchef2209:
> In general you should try to limit PRs to similar patches when feasible. For example, this should probably be one PR for the NEON stuff and one for the AVX-512BW stuff.

Sure, I will remember this and make separate PR from now on.
Thank You

simde/x86/sse4.1.h
masterchef2209: fixed this in #533

masterchef2209:
I see that all commits from this PR are already merged in master, if anything else needs  to be changed please tell me.
Thank You

nemequ:
> I will remember this and make separate PR from now on.

You don't have to do a separate PR for each patch, it would just be good to keep new implementations (mostly of AVX-512 functions) separate from adding optimized versions for NEON/AltiVec/etc.

> I see that all commits from this PR are already merged in master, if anything else needs to be changed please tell me.

Nope, if they're all merged we're good.  I'll close the issue, thanks!

Sorry, sometimes when there are a bunch of commits in a PR I don't close them.  When it's just a single commit I edit the commit message to say "Fixes #X", and GitHub automatically closes the PR when it hits the master branch.  When there are multiple patches it's not really fair to say any specific patch fixes the issue, and sometimes I forget to go in and close the PR manually after the patch finally makes it through CI.

PR #513, opened 2020-07-20
neon: Add qtbx with vqtbx[1234]{,q}_[su]8 intrinsics

rosbif:
Add qtbx with vqtbx[1234]{,q}_[su]8 intrinsics


rosbif:
Sorry, I forgit to change A32V7 to A64V8 :-(

PR #516, opened 2020-07-21
sse2,fma,avx512bw: added impl for mm512_sll{i,v}_epi16 + NEON impls

masterchef2209:


nemequ:
In C, shifting by more than the width of the type is undefined behavior.  That's why it is failing on non-x86 CPUs.

It's good that you have a test since we want to match Intel's behavior, but that means we need to handle a shift of greater than the width with a special case.

simde/x86/avx512bw.h
```diff
@@ -1895,6 +1895,60 @@ simde_mm512_packus_epi32 (simde__m512i a, simde__m512i b) {
...
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m512i
+simde_mm512_slli_epi16 (simde__m512i a, const int imm8)
+    SIMDE_REQUIRE_RANGE(imm8, 0, 255) {
+  simde__m512i_private
+    r_,
+    a_ = simde__m512i_to_private(a);
+
+  #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
+    r_.i16 = a_.i16 << HEDLEY_STATIC_CAST(int16_t, imm8);
```
nemequ, line 1907: So if `imm8 > 16` this is undefined behavior.

simde/x86/avx512bw.h
```diff
@@ -1895,6 +1895,60 @@ simde_mm512_packus_epi32 (simde__m512i a, simde__m512i b) {
...
+  #else
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(r_.i16) / sizeof(r_.i16[0])) ; i++) {
+      r_.i16[i] = HEDLEY_STATIC_CAST(int16_t, a_.i16[i] << (imm8 & 0xff));
```
nemequ, line 1911: Similarly here, if `(imm8 & 0xff) > 16`.

masterchef2209:
fixed this

PR #517, opened 2020-07-21
neon: Add rev64 with vrev64* intrinsics

rosbif:
vrev64_f16 and vrev64q_f16 are omitted as SIMDe has no 16-bit floating point support.


nemequ:
An implementation based on `SIMDE_SHUFFLE_VECTOR_` would go good, too.  If you're interested in adding it that would be great, otherwise we can add it down the road.

It's basically a hybrid of clang's __builtin_shufflevector GCC's __builtin_shuffle:

```
#define SIMDE_SHUFFLE_VECTOR_(elem_size_in_bits, vec_size_in_bytes, a, b, ...)
```

Where the ... is the lane number (like [`__builtin_shufflevector`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-shufflevector))

rosbif:
> An implementation based on `SIMDE_SHUFFLE_VECTOR_` would go good, too. If you're interested in adding it that would be great, otherwise we can add it down the road.
> 
> It's basically a hybrid of clang's __builtin_shufflevector GCC's __builtin_shuffle:
> 
> ```
> #define SIMDE_SHUFFLE_VECTOR_(elem_size_in_bits, vec_size_in_bytes, a, b, ...)
> ```
> 
> Where the ... is the lane number (like [`__builtin_shufflevector`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-shufflevector))

I hadn't noticed that there was a SIMDE_SHUFFLE_VECTOR. I deliberately try to avoid builtins. In what order do I place it wrt the native translations?
Anyway I would prefer to merge as is and maybe add SIMDE_SHUFFLE_VECTOR later, if that is OK with you.

nemequ:
> > An implementation based on `SIMDE_SHUFFLE_VECTOR_` would go good, too. If you're interested in adding it that would be great, otherwise we can add it down the road.
> > It's basically a hybrid of clang's __builtin_shufflevector GCC's __builtin_shuffle:
> > ```
> > #define SIMDE_SHUFFLE_VECTOR_(elem_size_in_bits, vec_size_in_bytes, a, b, ...)
> > ```
> > 
> > 
> > Where the ... is the lane number (like [`__builtin_shufflevector`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-shufflevector))
> 
> I hadn't noticed that there was a SIMDE_SHUFFLE_VECTOR. I deliberately try to avoid builtins.

Then you're missing out on some big optimizations. For compilers which support the builtin, the implementation will almost always be optimal.  Clang in particular tends to use `__builtin_shufflevector` to implement their functions instead of relying on instruction-specific builtins.  A lot of the native versions I've written have actually just been based on looking at the output from `SIMDE_SHUFFLE_VECTOR_`.

> In what order do I place it wrt the native translations?

After the native translations, but if it's available it would be fastest (semi-)portable version.

> Anyway I would prefer to merge as is and maybe add SIMDE_SHUFFLE_VECTOR later, if that is OK with you.

That's fine.  I actually already have this patch working through CI as is, so I'd kind of prefer it that way :)

rosbif:
> > I hadn't noticed that there was a SIMDE_SHUFFLE_VECTOR. I deliberately try to avoid builtins.
> 
> Then you're missing out on some big optimizations. For compilers which support the builtin, the implementation will almost always be optimal. Clang in particular tends to use `__builtin_shufflevector` to implement their functions instead of relying on instruction-specific builtins. A lot of the native versions I've written have actually just been based on looking at the output from `SIMDE_SHUFFLE_VECTOR_`.
> 

Maybe I should have explained that I try to avoid builtins, not because I doubt the quality of the generated code, but  because they are not portable and they are prone to causing FTBFS failures in multi-platform software like SIMDe if you don't correctly account for which versions of which compiler support them.

nemequ:
> Maybe I should have explained that I try to avoid builtins, not because I doubt the quality of the generated code, but because they are not portable and they are prone to causing FTBFS failures in multi-platform software like SIMDe if you don't correctly account for which versions of which compiler support them.

Ah, in that case, `SIMDE_SHUFFLE_VECTOR` should be good; we already have pretty solid logic for which versions of which compilers support it.  They can definitely complicate things, but usually I feel like the performance improvements are worth the effort.

For other projects, you might be interested in one of my other projects: the [builtin module in portable-snippets](https://github.com/nemequ/portable-snippets/tree/master/builtin).

FWIW, on clang (and clang-derived compilers) you can do `#if __has_builtin(__builtin_shufflevector)`. I really wish other compilers would adopt that as well…

rosbif:
Thank you for pointing out your portable-snippets project. There is a lot of interesting stuff.

PR #525, opened 2020-07-22
neon: implementation of cvt

seanptmaher:
I'm pretty sure I ironed out the rounding properties properly on ARM, but I might be surprised. 

(CC @tlively @ngzhian)


seanptmaher:
Hm, says it's failing on clang arm64 on drone, but I can't reproduce locally. I'll look into it.

nemequ:
The build which is failing has `-O2 -g`. If you can't otherwise reproduce it locally when running clang 10 that's probably a good place to start.

nemequ:
FWIW, I'm able to reproduce it with clang-10 and -O2 under qemu.

I'm using the [Docker development environment](https://github.com/simd-everywhere/simde/tree/master/docker), built with something like

```
mkdir /opt/simde/aarch64-clang-10-O2 && cd mkdir /opt/simde/aarch64-clang-10-O2
CC=clang-10 CXX=clang++-10 CFLAGS='--target=aarch64-linux-gnu -march=armv8-a+simd -isystem=/usr/aarch64-linux-gnu/include -Weverything -Werror -O2' CXXFLAGS='--target=aarch64-linux-gnu -march=armv8-a+simd -isystem=/usr/aarch64-linux-gnu/include -Weverything -Werror -O2' cmake /usr/local/src/simde/test -G Ninja
ninja && qemu-aarch64-static ./run-tests /arm/neon/vcvt
```

nemequ:
https://bugs.llvm.org/show_bug.cgi?id=46844

Since the problem is with the native call, the workaround should just be using the fallback code until the bug is fixed.

To do this, define a `SIMDE_BUG_CLANG_46844` in `simde-common.h` and just change `defined(SIMDE_ARM_NEON_A32V7_NATIVE)` in the problematic functions to `defined(SIMDE_ARM_NEON_A32V7_NATIVE) && !defined(SIMDE_BUG_CLANG_46844)`.

seanptmaher:
Alright, so, if you notice, in `test.h`, I changed the comparison  in `simde_test_equal_f32` and `64` from `((b > (a - slop)) && (b < (a + slop)))` to `((b >= (a - slop)) && (b <= (a + slop)))`. Is that fine? When I was doing conversion from integer to floating point, adding and removing 1 from the numbers wasn't sufficient because the exponent was too high, and I didn't want to only look at small integers, because if there was rounding in making it a floating point, I wanted to make sure they rounded in the same manner. 

Is there another way I should do this // does it break something?

nemequ:
> Alright, so, if you notice, in `test.h`, I changed the comparison in `simde_test_equal_f32` and `64` from `((b > (a - slop)) && (b < (a + slop)))` to `((b >= (a - slop)) && (b <= (a + slop)))`. Is that fine? When I was doing conversion from integer to floating point, adding and removing 1 from the numbers wasn't sufficient because the exponent was too high, and I didn't want to only look at small integers, because if there was rounding in making it a floating point, I wanted to make sure they rounded in the same manner.
> 
> Is there another way I should do this // does it break something?

I think it would be better to change the precision argument from 1 to 0.  The precision is converted to slop by doing `1e-##precision`, so precision(1) = 1e-1 = 0.1 and precision(0) = 1e-0 = 1.

We should probably also support negative precision values (-1 = 10.0, -2 = 100.0, etc.), I'll add a patch for that right now (unfortunately I don't think I can do it in a macro, but using simde_math_pow/powf shouldn't cause any problems.

nemequ:
Okay, I took another look at this, and now I see why you used `<=`/`>=`; the numbers are big enough that the next integer can't be represented, so adding 1 is basically a noop.  Sorry, I should have understood that from your comment :(.

You're right, `<=`/`>=` should be good, though I'd like to keep the ability to use precision values >= 0 for other code.

I'll go ahead and merge this, thanks as always, and sorry about the confusion.

seanptmaher:
Oh, thanks. 

Sorry, I haven't had as much time recently to look at this, I didn't have a chance to take a look carefully at what you were saying. 

I noticed the MSVC issue, I'm not sure how to solve it, so long as they're rounding floating point numbers the way they are.

nemequ:
No problem.

Yeah, I'm going to need to look at MSVC; I'd actually forgotten about that until I just sent this off to CI again.  One possibility might be to use `max(nextafter(value, SIMDE_MATH_INFINITY), value + slop)` instead of just `value + slop` (and similar code for the other direction, obviously).

nemequ:
That seems to have done the trick.  I'm not sure *why* it worked, probably MSVC being buggy, but it does.  Should be hitting master soon.

Thanks for the patch, sorry it took me so long to get to this :(

PR #534, opened 2020-07-24
svml: adding sleef to {hypot,log,log1p,log2,log10,pow} functions

himanshi18037:


nemequ:
This patch seems to do a *lot* more than the commit message says. It adds 2043 lines of code and removes 1884 lines.

Is this intentional, or maybe you were working off a different branch? If it's intentional, can you please explain the purpose of the other changes?

simde/x86/svml.h
```diff
@@ -4876,1164 +4950,837 @@ simde_mm512_mask_log_pd(simde__m512d src, simde__mmask8 k, simde__m512d a) {
 
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128
-simde_mm_cdfnorminv_ps (simde__m128 a) {
```
nemequ, line 4879: This is where things start to get weird.

himanshi18037: It was just to have all the log functions together, the cdfnorminv,erfcinv,erfinv all required log functions and maybe that is why I just added those below initially, however, while adding Sleef optimisations I realized that all these should be after the entire category of logarithmic functions. If you don't think so we can change it back.

nemequ: It just makes it very difficult to tell what is going on in the patch.

It probably wouldn't be a bad idea to re-sort the file a bit so all the functions are in order, but with the non-libm functions (which have implementations which depend on other functions) at the end.  However, this should really be a separate patch.

PR #541, opened 2020-07-28
sse4.1: added NEON impl for mm_dp_ps

masterchef2209:


masterchef2209:
icc tests are not running due to some problem which is not related to the patch, I will try rerunning tests at some later time, maybe it will work then

PR #542, opened 2020-07-28
svml: adding sleef to {mm,mm256,mm512}_cosd_{ps,pd}

himanshi18037:


simde/x86/avx2.h
```diff
@@ -1567,6 +1567,36 @@ simde_mm256_cvtepu32_epi64 (simde__m128i a) {
   #define _mm256_cvtepu32_epi64(a) simde_mm256_cvtepu32_epi64(a)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
```
nemequ, line 1570: I'm not sure how useful these will be outside of SVML. Can you move them over to svml.h (and the tests to svml.c)?

In the future we can always move them out of SVML if they're needed elsewhere since svml.h includes basically every x86 SIMDe header.  However, once we commit to providing these functions in sse/sse2/avx/etc.h moving them would be an API break.

PR #543, opened 2020-07-28
neon: Add create with vcreate_* intrinsics

rosbif:
Add create with vcreate_* intrinsics.


nemequ:
Wow, I hate these. Not your implementations, I don't like the API.

I see you've already accounted for endianness by using vrev*, but instead of explicitly reversing the bytes it would be better to just do something like `simde_test_arm_neon_assert_equal_u64x1(simde_vreinterpret_u64_s8(r), test_vec[i].r);`?

That would let us handle PDP-endian as well, not that I think SIMDe would actually work on a PDP-endian machine without some tweaks.  It just feels "more correct" to me to restructure the code a bit so the endianness issue goes away instead of having to handle it explicitly.

I think I'd also rather use, for example, `return vreinterpret_s8_u64(vdup_n_u64(a));` for the implementations instead of using memcpy. What do you think?

rosbif:
> Wow, I hate these. Not your implementations, I don't like the API.
> 
> I see you've already accounted for endianness by using vrev*, but instead of explicitly reversing the bytes it would be better to just do something like `simde_test_arm_neon_assert_equal_u64x1(simde_vreinterpret_u64_s8(r), test_vec[i].r);`?
> 

OK I'll try.

> That would let us handle PDP-endian as well, not that I think SIMDe would actually work on a PDP-endian machine without some tweaks. It just feels "more correct" to me to restructure the code a bit so the endianness issue goes away instead of having to handle it explicitly.
> 

I didn't realize that you were old enough to remember the PDP11 ;-) It must date back to the 1970s as it was superseded by the VAX by 1980 IIRC. The VAX was pure little-endian (although perhaps it had a mode to emulate the PDP11, but I don't remember).

> I think I'd also rather use, for example, `return vreinterpret_s8_u64(vdup_n_u64(a));` for the implementations instead of using memcpy. What do you think?

OK, I'll try.

nemequ:
> I didn't realize that you were old enough to remember the PDP11 ;-) It must date back to the 1970s as it was superseded by the VAX by 1980 IIRC. The VAX was pure little-endian (although perhaps it had a mode to emulate the PDP11, but I don't remember).

I'm not (I was born in 1984).  I've never had the ~pleasure~ experience of coding for a PDP11 (or VAX), but I do know a bit about it and its weird endianness.

Really handling PDP-endian is just an excuse; the real reason is that I generally feel like explicit endian swaps are a hack (unless you're serializing/deserializing to a format which requires a specific endianness), and IMHO it's generally best to avoid hacks like that if you can instead avoid the *need* for hacks like that.

In this case, I don't think we would really gain anything by explicitly swapping endianness, especially since these are in the tests not the actual implementations.

rosbif:
> Wow, I hate these. Not your implementations, I don't like the API.
> 

I see what you mean. This API invites endian problems.

I have now implemented your suggestions.

nemequ:
It's not just about the endianness (though obviously that is a concern), it's the punning in general. NEON already has a way to cast which is type-safe (if that's the right word): `vreinterpret*`

If you're really loading a 64-bit unsigned integer, `vdup_n_s64(value)` should generally be what you want anyways. If not, it's likely that you just have 64 bits of stuff shoved into a `uint64_t`, in which case you should probably be providing the original data type instead of using `uint64_t` as an intermediary.

I think the API encourages bad code just to make things easier for a very small number of cases for where you really do want to do exactly what these functions do. In those cases something like `vreinterpretq_s8_u64(vdupq_n_u64(foo))` should really be enough.  A good API isn't just easy to use, it is also hard to *mis*use. These functions are easy to misuse, and don't provide a substantial benefit.

Also, the name "create"… if I was trying to figure out how to create a NEON vector, I'd think to search for "create" long before "ld1" or "dup_n". I shudder to think how many people found these functions and just thought that's how you create vectors in NEON.

rosbif:
> Also, the name "create"… if I was trying to figure out how to create a NEON vector, I'd think to search for "create" long before "ld1" or "dup_n". I shudder to think how many people found these functions and just thought that's how you create vectors in NEON.
>

Yes, I started out with NEON by manually translating the SSE intrinsics in my application to NEON. To replace _mm_set_epi64x() I used a vcombine() of two vcreates().

PR #547, opened 2020-07-31
sse2: added NEON impl for mm_{min,max}_{sd,pd}

masterchef2209:


simde/x86/sse2.h
```diff
@@ -3586,7 +3586,9 @@ simde_mm_min_pd (simde__m128d a, simde__m128d b) {
       a_ = simde__m128d_to_private(a),
       b_ = simde__m128d_to_private(b);
 
-    #if defined(SIMDE_POWER_ALTIVEC_P8_NATIVE)
+    #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
```
nemequ, line 3589: There is no need for the NEON implementation to come first.  If there is already something else (in this case, an AltiVec implementanion) you should add an `#elif` for NEON instead of changing the `#if`. 

There are several (relatively minor) advantages

 * It minimizes the diff (instead of adding three lines and removing one, you would just add one).
 * It's easier to understand the diff.
 * It preserve the git blame. If you don't already know about git blame, take a look at `git blame simde/x86/sse2.h` (or whatever file you want). Basically you can see exactly when a line was added. That's extremely valuable because you can go back and look at the commit to see *why* that change was made.

It's really not a big deal, but it's definitely a good habit to get into, especially when you're working with large codebases there there can be a lot of knowledge encoded into the source code that isn't necessarily documented in comments.

simde/x86/sse2.h
```diff
@@ -3694,7 +3701,9 @@ simde_mm_max_pd (simde__m128d a, simde__m128d b) {
       a_ = simde__m128d_to_private(a),
       b_ = simde__m128d_to_private(b);
 
-    #if defined(SIMDE_POWER_ALTIVEC_P7_NATIVE)
```
nemequ, line 3697: Same here.

I'm guessing you were trying to preserve the order from other functions, but you don't need to worry about that.  The only reason NEON is usually first is because that's usually the first implementation we add.  In this case it just so happens that we added AltiVec first.

simde/x86/sse2.h
masterchef2209: I see, I will remember this.
Thank You

simde/x86/sse2.h
masterchef2209: > I'm guessing you were trying to preserve the order from other functions, but you don't need to worry about that. The only reason NEON is usually first is because that's usually the first implementation we add. In this case it just so happens that we added AltiVec first.

Yes, I was trying to preserve the order so that it is consistent with all the functions.

masterchef2209: Sorry, I didn't realize I have to fix this, I have fixed it now.
Thank You

PR #550, opened 2020-08-02
sse: optimising simde_mm_rcp_ps

himanshi18037:


nemequ:
Merged as 313bcf836f4eb470f8bfe5752aaca8ebfa70392b, thanks!

I did tweak the feature test to be based on IEEE 754 (aka IEC 559), but I think that's it.

And I just realized it should have a `SIMDE_VECTORIZE` on the loop.  Oops.  I'll add it, sorry :(

PR #551, opened 2020-08-02
allow skipping the inclusion of the complex header

kitegi:
the c++ `<complex>` header can add quite a bit of compilation overhead, since it includes the `io` headers to define serializing operations.  
this PR allows the user to define a macro `SIMDE_NO_COMPLEX` to skip that header, which reduces the compilation time of a trivial c++ program from ~0.46s to ~0.28s on my machine. not a massive gain by any means, but i thought it'd be nice to give this option.

```cpp
#define SIMDE_NO_COMPLEX
#include "simde/x86/avx2.h"
int main() {}
```

### DIFF
--- a/simde/simde-math.h
+++ b/simde/simde-math.h
@@ -123,6 +123,7 @@
   #endif
 #endif
 
+#ifndef SIMDE_NO_COMPLEX
 #if !defined(__cplusplus)
   /* If this is a problem we *might* be able to avoid including
    * <complex.h> on some compilers (gcc, clang, and others which
@@ -213,6 +214,7 @@
   #if !defined(simde_math_cimagf)
     #define simde_math_cimagf(z) ((z).imag())
   #endif
+#endif
 #endif
 
 #if !defined(SIMDE_MATH_INFINITY)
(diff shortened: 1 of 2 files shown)


nemequ:
Wow, I had no idea <complex> was that heavy, thanks for looking into it! I'm guessing that with sse.h compilation time would be completely dominated by <complex>…

I would definitely like a way to avoid <complex>, especially since it's really only used in SVML right now, and I'd be willing to go with the `SIMDE_NO_COMPLEX` idea if need be, but I think it may be possible to avoid including <complex>, at least on GCC/clang, by using builtins and other extensions.  That way we wouldn't have to rely on people knowing about this issue and using `SIMDE_NO_COMPLEX`, it would just be fast by default.

I'll try to look into this soon, if I can't figure out a way to make it work with builtins and/or extensions I'll add the `SIMDE_NO_COMPLEX` option.

nemequ:
I added 1b048505cac3f92a66c4bc0ee98dfd1541301e65 for this, does that work for you?

It still doesn't allow you to skip <complex> on MSVC, but most other modern compilers should work.  If you need this to work with MSVC I might be able to find a way, but TBH I'd probably just add something like `SIMDE_NO_COMPLEX` at that point.

PR #553, opened 2020-08-03
neon: Avoid -Wstrict-aliasing in tbl, tbx, qtbl and qtbx

rosbif:
Also correct indentation.


nemequ:
LGTM, thanks.

I'm always tempted to do something like

```c
#define SIMDE_REINTERPRET_AS(T, value) \
  (__extension__ ({ \
    HEDLEY_STATIC_ASSERT(sizeof(T) == sizeof(value), "size of destination type must match size of source type"); \
    __typeof__(value) simde_reinterpret_as_src_ = (value); \
    T simde_reinterpret_as_dest_; \
    simde_memcpy(&simde_reinterpret_as_dest_, &simde_reinterpret_as_src_, sizeof(simde_reinterpret_as_dest_)); \
    simde_reinterpret_as_dest_; \
  }))
```

It would just be nice to be able to write stuff like `SIMDE_REINTERPRET_AS(simde_uint8x16x2_t, t)`

Unfortunately I don't know of a way to do it in MSVC in C mode.  MSVC ruins everything.  The best I can think of would be a macro to easily create conversion functions, but IMHO that's pretty ugly :(

PR #555, opened 2020-08-03
avx512vl: implement mm_abs_{mask,maskz}_{epi32,epi64}

himanshi18037:


nemequ:
I think this is a duplicate of #556, with a slightly different name?  Closing, but if it was intentional feel free to reopen.

BTW, you can edit the title of PRs.  Next to the title, all the way on the right, there is an "Edit" button.

himanshi18037:
> I think this is a duplicate of #556, with a slightly different name? Closing, but if it was intentional feel free to reopen.
> 
> BTW, you can edit the title of PRs. Next to the title, all the way on the right, there is an "Edit" button.

These are different functions, the previous do not include for epi32 and epi64.

PR #557, opened 2020-08-04
sse2: added NEON impl for mm_sub_{sd,pd,epu32,si64}, mm_{add,sqrt}_sd

masterchef2209:


simde/x86/sse2.h
```diff
@@ -5968,8 +5980,13 @@ simde_mm_sub_sd (simde__m128d a, simde__m128d b) {
       a_ = simde__m128d_to_private(a),
       b_ = simde__m128d_to_private(b);
 
-    r_.f64[0] = a_.f64[0] - b_.f64[0];
-    r_.f64[1] = a_.f64[1];
+    #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
```
nemequ, line 5983: You don't really need to worry about NEON implementations of the *_sd functions.  This should do the same thing as `simde_mm_move_sd(a, simde_mm_sub_pd(a, b))`, having a separate implementation here only complicates the code unnecessarily.

That said, I believe this code is actually wrong; the second argument to `vsetq_lane_f64` should be a, `a_.neon_f64` not `temp`.  The compiler just never hits this case because it uses the `simde_mm_move_sd`-based path above.

masterchef2209: > You don't really need to worry about NEON implementations of the *_sd functions. This should do the same thing as `simde_mm_move_sd(a, simde_mm_sub_pd(a, b))`, having a separate implementation here only complicates the code unnecessarily.

Thank You, I got it, so if a function is calling other simde function directly and the other function has a NEON implementation so we don't need to write NEON implementation separately for this function. Am I right?

> That said, I believe this code is actually wrong; the second argument to `vsetq_lane_f64` should be a, `a_.neon_f64` not `temp`. The compiler just never hits this case because it uses the `simde_mm_move_sd`-based path above.

I think it is correct because what I am doing is
temp[i]=a[i]-b[i], then replaceing temp[1] by a[1], so we get
temp[0]=a[0]-b[0]
and temp[1]=a[1], sorry if the code was not clear but I get your point of not needing NEON implementation for *_sd functions.
Thank You

PR #560, opened 2020-08-05
neon: implementation of {ld,st}{3,4}

seanptmaher:
Hello,

I wrote ld3/4 and st3/4.

To test them, I used ld3 and ld4 to do some loads/stores to make sure it was working properly, but I only wrote test files for st{3,4}, as those test files use ld{3,4}. That ok?

(CC @tlively @ngzhian)


seanptmaher:
There's actually a bug here in my handling of multidimensional arrays, so don't touch this yet

seanptmaher:
That emscripten fail doesn't seem to be related to my code, is that a known problem?

seanptmaher:
Also, after having rewritten, this, I think it's good for you to take a look at.

nemequ:
> That emscripten fail doesn't seem to be related to my code

Agreed.

> is that a known problem?

An old problem. I've seen the issue before, but I thought we disabled that warning before including µnit.  Maybe an updated package, and the warning used to be part of -Wcpp98-compat-pedantic or something?  I'll take a look in a bit, don't worry about it.

It didn't occur on [the previous build](https://github.com/simd-everywhere/simde/runs/947091078), though, if that's what you're asking.

mr-c:
@To test them, I used ld3 and ld4 to do some loads/stores to make sure it was working properly, but I only wrote test files for st{3,4}, as those test files use ld{3,4}. That ok?

I strongly prefer individual tests, as we use them to ensure that native aliases work (see https://github.com/simd-everywhere/simde/blob/master/test/native-aliases.sh )

nemequ:
Yes, if possible the load and store tests should be separate.  Native aliases aren't really a problem here since the tests contain both functions, so the native aliases tests will just replace both prefixed functions with the native versions and we do still test everything.

The problem with this is that all you're testing is that store does the opposite of load, you're not testing the layout of the memory. These load functions functions are non-trivial; they de-interleave as they load.  I think the easiest way would probably be something like

```c
  static const struct {
    simde_float32 a[3][2];
    simde_float32 r0[2];
    simde_float32 r1[2];
    simde_float32 r2[2];
 } test_vec = ...
```

Then

```c
simde_float32x2x3_t r = simde_vld3_f32(test_vec[i].a);
simde_test_arm_neon_assert_equal_f32x2(r.val[0], simde_vld1_f32(test_vec[i].r0), 1);
simde_test_arm_neon_assert_equal_f32x2(r.val[1], simde_vld1_f32(test_vec[i].r1), 1);
simde_test_arm_neon_assert_equal_f32x2(r.val[2], simde_vld1_f32(test_vec[i].r2), 1);
```

seanptmaher:
> ```c
>   static const struct {
>     simde_float32 a[3][2];
>     simde_float32 r0[2];
>     simde_float32 r1[2];
>     simde_float32 r2[2];
>  } test_vec = ...
> ```
> 
> Then
> 
> ```c
> simde_float32x2x3_t r = simde_vld3_f32(test_vec[i].a);
> simde_test_arm_neon_assert_equal_f32x2(r.val[0], simde_vld1_f32(test_vec[i].r0), 1);
> simde_test_arm_neon_assert_equal_f32x2(r.val[1], simde_vld1_f32(test_vec[i].r1), 1);
> simde_test_arm_neon_assert_equal_f32x2(r.val[2], simde_vld1_f32(test_vec[i].r2), 1);
> ```

What exactly do you mean here? I'm slightly confused. Do you mean that I'd store the same values in `a` and `r[0-3]`, and then use `r[0-3]` to verify the ordering?

seanptmaher:
Or, I could do something like:
```c
  static const struct {
    union {
      simde_float32 a[3][2];
      struct {
        simde_float32 r0[2];
        simde_float32 r1[2];
        simde_float32 r2[2];
      } r;
   } test_vec = ...
```
and test it that way?

nemequ:
> Do you mean that I'd store the same values in `a` and `r[0-3]`, and then use `r[0-3]` to verify the ordering?

Basically. yes. It would be the same values, but they'll be in a different order since the load de-interleaves them.

> and test it that way?

Yeah, that would be fine.  Whatever you want to do; I'm not picky about the implementation as long as they test the function, so if that's easier by all means feel free.

nemequ:
To be clear, `vld3_*` is *not* the same thing as a memcpy. I believe `vld1_*_x3` is, but I'd have to check to be certain.

Also, the s64/u64 functions use A32V7 for the test, but they should use A64V8.  Drone.io actually picked up the problem when I pushed this to a brach, but for some reason PRs don't seem to trigger Drone.  I'm not sure what's going on there, but until I figure it out please feel free to push to a branch in the main repo, that should trigger Drone.

Or, of course, you can use Docker to test. `./docker/simde-dev.sh unstable` should drop you into a shell, then `simde-reset-build.sh` to configure the builds, and `ninja -C armv8-gcc-10 test` (for example) to compile and test on gcc targeting arm.

seanptmaher:
Alright, I rewrote ld3 and st3, I had misunderstood what they did I will rewrite ld4 and st4 tomorrow, doing both was a bit too much for the time I had today.

You say that s64/u64 should use A64V8, but [the isa reference](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vst3_s64) says they're supported on V7/A32/A64. Am I missing something here?

seanptmaher:
Should be good now. Let me know if anything else is off about it.

seanptmaher:
Hm, looks like icc is failing, but I'm not too sure why. Looks like the x3 vectors aren't defined, which is odd... Were they disabled on icc, which I missed?

nemequ:
> Alright, I rewrote ld3 and st3, I had misunderstood what they did I will rewrite ld4 and st4 tomorrow, doing both was a bit too much for the time I had today.

IMHO they're confusingly named.  I would expect them to do what you thought they did as well, and Arm's documentation doesn't really help.  Overall I think Arm did a really good job with NEON, but there are a few functions where they went in a weird direction, including these.

> You say that s64/u64 should use A64V8, but the isa reference says they're supported on V7/A32/A64. Am I missing something here?

Interesting, I thought it was all the s64/u64 functions for this, but it looks like it's just the 128-bit versions.  I'll tweak them.

https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vst3q_s64

> Hm, looks like icc is failing, but I'm not too sure why. Looks like the x3 vectors aren't defined, which is odd... Were they disabled on icc, which I missed?

Yes, due to https://software.intel.com/en-us/forums/intel-c-compiler/topic/857088.  I'll take care of the ifdefs to disable them; I'm guessing you're pretty sick of this issue by now :/

seanptmaher:
Thanks for your help!

PR #561, opened 2020-08-05
avx512vl: implement mm_{mask,maskz}_add_epi{8,16,32,64}

himanshi18037:


simde/x86/avx512vl.h
```diff
@@ -352,6 +352,110 @@ simde_mm256_maskz_abs_epi64(simde__mmask8 k, simde__m256i a) {
 #define _mm256_maskz_abs_epi64(k, a) simde_mm256_maskz_abs_epi64(k, a)
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128i
+simde_mm_mask_add_epi8(simde__m128i src, simde__mmask16 k, simde__m128i a, simde__m128i b) {
+#if defined(SIMDE_X86_AVX512VL_NATIVE)
```
nemequ, line 358: Please be careful about indenting; the preprocessor directives should be indented in the function just like regular C code.

simde/x86/avx512vl.h
```diff
@@ -352,6 +352,110 @@ simde_mm256_maskz_abs_epi64(simde__mmask8 k, simde__m256i a) {
...
+  return _mm_mask_add_epi8(src, k, a, b);
+#else
+  return simde_mm_mask_mov_epi8(src, k, simde_mm_add_epi8(a, b));
+#endif
+}
+#if defined(SIMDE_X86_AVX512VL_ENABLE_NATIVE_ALIASES)
+#define _mm_mask_add_epi8(src, k, a, b) simde_mm_mask_add_epi8(src, k, a, b)
```
nemequ, line 365: Here too; if/elif/else preprocessor directives should increase the indentation of code inside of them (both other preprocessor macros and normal code).

I'll fix these when I merge, just something to look out for in the future.

simde/x86/avx512vl.h
```diff
@@ -352,6 +352,110 @@ simde_mm256_maskz_abs_epi64(simde__mmask8 k, simde__m256i a) {
...
+simde__m128i
+simde_mm_mask_add_epi16(simde__m128i src, simde__mmask8 k, simde__m128i a, simde__m128i b) {
+#if defined(SIMDE_X86_AVX512VL_NATIVE)
+  return _mm_mask_add_epi16(src, k, a, b);
+#else
+  return simde_mm_mask_mov_epi16(src, k, simde_mm_add_epi16(a, b));
+#endif
+}
+#if defined(SIMDE_X86_AVX512VL_ENABLE_NATIVE_ALIASES)
+#define _mm_mask_add_epi16(src, k, a, b) simde_mm_mask_add_epi16(src, k, a, b)
```
nemequ, line 391: Also, these are missing `#undef _mm_mask_add_epi16` directives before defininig the macros.  They are necessary on everything after SSE 4.2 (since Intel started using <immintrin.h> for everything instead of having separate includes for each ISA extension).

PR #562, opened 2020-08-05
neon: SSE2 is needed for _mm_castps_si128 in ceq, cge, cgt, cle and clt

rosbif:
The Intel compare float32 intrinsics only need SSE but the casts to integer require SSE2 :-(
If necessary it would be possible to get round this stupid situation with a memcpy instead of a cast.


nemequ:
> If necessary it would be possible to get round this stupid situation with a memcpy instead of a cast.

Meh, SSE and SSE2 basically have the same adoption rate (pretty much 100%) since they're both included in baseline x86_64. I'm not too worried.

PR #573, opened 2020-08-10
sse2,sse4.1: added NEON impl for mm_x_copysign_{ps,pd}, mm_round_s{d,s}

masterchef2209:


nemequ:
I'm pulling in the round_ss/sd stuff, but for copysign I put together a different implementation using vbsl which should be a bit more efficient.

PR #575, opened 2020-08-12
avx512dq: implement mm{256,512}_broadcast_f32x{2,8}

himanshi18037:


simde/x86/avx512dq.h
```diff
@@ -204,6 +204,122 @@ simde_mm512_maskz_and_pd(simde__mmask8 k, simde__m512d a, simde__m512d b) {
...
+  #else
+    simde__m512_private r_;
+    simde__m256_private a_ = simde__m256_to_private(a);
+
+    #if defined(SIMDE_VECTOR_SUBSCRIPT) && HEDLEY_HAS_BUILTIN(__builtin_shufflevector)
+      r_.f32 = __builtin_shufflevector(a_.f32, a_.f32, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7);
+    #else
+      SIMDE_VECTORIZE
+      for (size_t i = 0 ; i < (sizeof(r_.f32) / sizeof(r_.f32[0])) ; i+=8) {
+         r_.f32[i]     = a_.f32[0];
```
nemequ, line 276: The patch looks great, merging now.  One trivial style suggestion here:

```c
r_.f32[  i  ] = a_.f32[0];
r_.f32[i + 1] = a_.f32[1];
...
```

Instead of

```c
r_.f32[i]     = a_.f32[0];
r_.f32[i + 1] = a_.f32[1];
...
```

Lining up the square brackets definitely makes it easier to read.  An argument could be made for `i    ` instead of `  i  ` so the + lines up, too, though I prefer centering the `i`.  Another style I see sometimes is `i + 0`, which can also be a good way to go.

Obviously it doesn't really change anything, but anything we can do to keep the code readable helps, so as long as it doesn't hurt performance…

PR #576, opened 2020-08-12
seanptmaher: Add Google copyright note

seanptmaher:
Hello, 

I noticed I hadn't put in the copyright that the work I've done is owned by Google, so I added that in here. 

Doesn't really change anything else.

(cc @tlively @ngzhian)

### DIFF
--- a/simde/arm/neon/addl.h
+++ b/simde/arm/neon/addl.h
@@ -22,7 +22,7 @@
  *
  * Copyright:
  *   2020      Evan Nemerson <evan@nemerson.com>
- *   2020      Sean Maher <seanptmaher@gmail.com>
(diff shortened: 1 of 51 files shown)


nemequ:
Since this is a copyright declaration, not an AUTHORS file, I think it would be more appropriate to reverse this so it's:

> 2020      Google, LLC (Author: Sean Maher <seanptmaher@gmail.com>)

I like the idea of keeping your name there, though, even if it's not technically necessary in a copyright declaration.

I went ahead and tweaked the patch, I hope you don't mind.  It should hit master soon (assuming CI is working today…).

seanptmaher:
Nice, sounds good. Thanks for the help.

PR #585, opened 2020-08-17
svml: adding {128,256} bit fallback to simde_mm{256,512}_acos_{ps,pd}

himanshi18037:


simde/x86/svml.h
```diff
@@ -117,10 +117,16 @@ simde_mm256_acos_ps (simde__m256 a) {
       r_,
       a_ = simde__m256_to_private(a);
 
+    #if SIMDE_NATURAL_VECTOR_SIZE_LE(128)
+      for (size_t i = 0 ; i < (sizeof(r_.m128) / sizeof(r_.m128[0])) ; i++) {
+        r_.m128[i] = simde_mm_acos_ps(a_.m128[i]);
+    }
+    #else
     SIMDE_VECTORIZE
```
nemequ, line 125: The inside of the else blocks (lines 125-128 here) should be indented to the same level as the for loop above.

simde/x86/svml.h
```diff
@@ -117,10 +117,16 @@ simde_mm256_acos_ps (simde__m256 a) {
       r_,
       a_ = simde__m256_to_private(a);
 
+    #if SIMDE_NATURAL_VECTOR_SIZE_LE(128)
+      for (size_t i = 0 ; i < (sizeof(r_.m128) / sizeof(r_.m128[0])) ; i++) {
+        r_.m128[i] = simde_mm_acos_ps(a_.m128[i]);
+    }
```
nemequ, line 123: This should be indented to the same level as the beginning of the for loop.

PR #601, opened 2020-08-28
avx512f: implement mm_{mask,maskz}_add_ss

himanshi18037:


simde/x86/avx512f.h
```diff
@@ -3692,6 +3692,34 @@ simde_mm512_maskz_add_pd(simde__mmask8 k, simde__m512d a, simde__m512d b) {
...
+simde_mm_maskz_add_ss(simde__mmask8 k, simde__m128 a, simde__m128 b) {
+  #if defined(SIMDE_X86_AVX128F_NATIVE)
+    return _mm_maskz_add_ss(k, a, b);
+  #else
+    return simde_mm_maskz_mov_ps(k, simde_mm_add_ss(a, b));
+  #endif
+}
+#if defined(SIMDE_X86_AVX128F_ENABLE_NATIVE_ALIASES)
+  #undef _mm_maskz_add_ss
+  #define _mm_maskz_add_ss(src, k, a, b) simde_mm_maskz_add_ss(src, k, a, b)
```
masterchef2209, line 3720: @himanshi18037 arguments are wrong here, please fix this.

himanshi18037: Thanks!

simde/x86/avx512f.h
```diff
@@ -3692,6 +3692,34 @@ simde_mm512_maskz_add_pd(simde__mmask8 k, simde__m512d a, simde__m512d b) {
...
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m128
+simde_mm_mask_add_ss(simde__m128 src, simde__mmask8 k, simde__m128 a, simde__m128 b) {
+  #if defined(SIMDE_X86_AVX128F_NATIVE)
+    return _mm_mask_add_ss(src, k, a, b);
+  #else
+    return simde_mm_mask_mov_ps(src, k, simde_mm_add_ss(a, b));
+  #endif
+}
+#if defined(SIMDE_X86_AVX128F_ENABLE_NATIVE_ALIASES)
```
nemequ, line 3704: You have `AVX128F` instead of `AVX512F` here (and on the maskz version), that's why the native aliases test is still failing; the `_mm_mask_add_ss` macro is never defined because `SIMDE_X86_AVX128F_ENABLE_NATIVE_ALIASES` is never defined.

nemequ (changes requested):
I'm in the middle of refactoring AVX-512 to be structured a bit more like NEON (one file per function family instead of one file per ISA extension). Rebasing that would be a huge pain, so I'll merge the AVX-512 patches when I"m done (probably tomorrow).

PR #612, opened 2020-09-04
sse/sse2: _mm512_abs_ps introduced in GCC7.1, simde_x_mm_abs_pd does not work before GCC 7.4

milot-mirdita:
simde_x_mm_abs_ps was breaking MMseqs2 build on GCC 4.9. This should fix compilation on GCC < 7.1.

### DIFF
--- a/simde/x86/sse.h
+++ b/simde/x86/sse.h
@@ -922,7 +922,8 @@ simde_mm_avg_pu8 (simde__m64 a, simde__m64 b) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128
 simde_x_mm_abs_ps(simde__m128 a) {
-  #if defined(SIMDE_X86_AVX512F_NATIVE)
+  #if defined(SIMDE_X86_AVX512F_NATIVE) && \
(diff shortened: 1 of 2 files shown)


mr-c:
I guess we need to add a gcc < 7.1 with AVX512F native to the test matrix

nemequ:
> I guess we need to add a gcc < 7.1 with AVX512F native to the test matrix

The AVX-512F part is the problem; SDE is terribly slow.

Travis has AVX-512 hardware, maybe we could add gcc-4.7 / clang-3.5 builds there? I think we can move the SIMDE_NO_SHUFFLE_VECTOR build over to GitHub Actions or Azure Pipelines, and the armv7/armv8 builds are really duplicates of some stuff on Drone…

nemequ:
Merged a modified version (there are are abs functions in NEON which had the same problem when implemented using AVX-512), thanks.

I also added a GCC 6 build on Travis (which supports AVX-512) so this doesn't happen again, as well as some other CI improvements. We're now up to ~ 61 builds in CI ☹.

milot-mirdita:
We try to still support GCC 4.9 (4.8 could theoretically also work). Would it be possible to either use that instead of 6 or add one more?

PR #617, opened 2020-09-23
docker: add tigerlake flag to sdl in docker .cross files

ashnewmanjones:
SDE was failing when using Tigerlake instruction sets, adding flag to SDE to target the same as -march.

### DIFF
--- a/docker/cross-files/i686-all-gcc-10.cross
+++ b/docker/cross-files/i686-all-gcc-10.cross
@@ -5,7 +5,7 @@ ar = '/usr/bin/ar'
 strip = '/usr/bin/strip'
 objcopy = '/usr/bin/objcopy'
 ld = '/usr/bin/ld'
-exe_wrapper = ['sde', '--']
+exe_wrapper = ['sde', '-future', '--']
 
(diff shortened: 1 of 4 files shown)


nemequ:
Oops, I thought omitting the flag was the same as just saying to enable all extensions… Glad you caught this, thanks!

Instead of -tgl, would you have any objection to using -future? We're really just using Tiger Lake because compilers don't have a way (AFAIK) to just enable all the extensions. IIRC some of the older builds have to use icelake server (icx) or older because the compiler doesn't support the newer targets.

ashnewmanjones:
No objection to -future instead, just thought to use the same as the compiler flag.

mr-c (approved):

nemequ:
> No objection to -future instead, just thought to use the same as the compiler flag.

That is completely sensible; if I had just looked at most of the CI jobs that's probably what I would have done, too.

Sorry, I really meant to update the patch myself not wait for you to do it.  Thanks for updating it, though, and of course for tracking down the issue and fixing it.  It should hit the default branch soon.

PR #618, opened 2020-09-23
README: add two related projects

Artoria2e5:
The loongson one is really fringe, so it's fine if you want to take it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,10 @@ We would love to add more, so patches are extremely welcome!
      [powerveclib](https://www.ibm.com/developerworks/community/groups/community/powerveclib/)
    * [SSE-to-NEON](https://github.com/otim/SSE-to-NEON) — implementing
      SSE with NEON.  Non-free, C++.
+   * [loongson-community GCC](https://github.com/loongson-community/gcc/commit/99cb272)  —
+     implementing SSE2 with Loongson SIMD.
+   * [Emscripten](https://emscripten.org/docs/porting/simd.html) —
+     implementing SSE and 128-bit AVX with WebAssembly SIMD.
  * [arm-neon-tests](https://github.com/christophe-lyon/arm-neon-tests)
    contains tests to verify NEON implementations.
 


nemequ:
Thanks!  I wasn't aware of the Loongson work, that's awesome. I think I'll reach out to the authors to see if they might be interested in collaborating; I'd be happy to add Loongson support to SIMDe that way everyone can benefit, not just GCC.

As for emscripten, I believe the current plan is actually to get all that code merged into SIMDe then have Emscripten just use SIMDe; they are actually already doing that for NEON.

Anyways, instead of this patch I think it would be better to just add a link in the README to the [Alternatives](https://github.com/simd-everywhere/simde/wiki/Alternatives) page in the Wiki, and add these there.  If you're willing to update this PR to just point to the wiki I'd be happy to accept it, and of course feel free to add any relevant information to the wiki.

nemequ:
Regarding Loongson, it looks like the MMX-compatible functions were removed pretty quickly (in [d65b9c](https://github.com/loongson-community/gcc/commit/d65b9c4e78b1bca57ad7bd669cbb9373cd3234da)), so we probably shouldn't focus on that too much.  It seems like the preferred path for Loongson is to use the functions in [loongson-mmiintrin.h](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/mips/loongson-mmiintrin.h), which aren't x86-compatible.

Maybe @FlyGoat has some insight into where that code came from and why it disappeared so quickly? If it is still alive somewhere I'd be happy to link to them from the wiki, or maybe even merge them into SIMDe (depending on licensing).

FlyGoat:
> Regarding Loongson, it looks like the MMX-compatible functions were removed pretty quickly (in [d65b9c](https://github.com/loongson-community/gcc/commit/d65b9c4e78b1bca57ad7bd669cbb9373cd3234da)), so we probably shouldn't focus on that too much. It seems like the preferred path for Loongson is to use the functions in [loongson-mmiintrin.h](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/mips/loongson-mmiintrin.h), which aren't x86-compatible.
> 
> Maybe @FlyGoat has some insight into where that code came from and why it disappeared so quickly? If it is still alive somewhere I'd be happy to link to them from the wiki, or maybe even merge them into SIMDe (depending on licensing).

Well because upstream suggested we shouldn't do it in GCC.

nemequ:
@Artoria2e5, do you want to submit a patch to tweak the README to link to the alternatives page, or would you like me to do it?

Either way, thanks again for filing this; it looks like we'll merge Loongson support into SIMDe soon thanks to this issue (and @FlyGoat's patches).

Artoria2e5:
I think it would be more appropriate for you to do it.

PR #619, opened 2020-09-28
Loongson MMI support

FlyGoat:
Need instructions on how to add CI tasks :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,21 @@ jobs:
         - gcc-10
         - g++-10
 
+  - name: "MIPS Loongson-MMI (Compile Only)"
+    if: branch != master OR type == pull_request
+    arch: amd64
+    compiler: mips64el-linux-gnuabi64-gcc
+    env:
+    - C_COMPILER=mips64el-linux-gnuabi64-gcc
+    - CXX_COMPILER=mips64el-linux-gnuabi64-g++
+    - ARCH_FLAGS="-march=loongson3a"
+    - RUN_TESTS=false
+    addons:
+      apt:
+        packages:
+        - gcc-mips64el-linux-gnuabi64
+        - g++-mips64el-linux-gnuabi64
+
   # These next two ere largely to detect missing AVX-512 functions
   # since Travis is the only CI right now with AVX-512 hardware and SDE
   # is too slow.
--- a/simde/simde-arch.h
+++ b/simde/simde-arch.h
@@ -361,6 +361,10 @@
 #  define SIMDE_ARCH_MIPS_CHECK(version) (0)
 #endif
 
+#if defined(__mips_loongson_mmi)
+#  define SIMDE_ARCH_MIPS_LOONGSON_MMI 1
+#endif
+
 /* Matsushita MN10300
    <https://en.wikipedia.org/wiki/MN103> */
 #if defined(__MN10300__) || defined(__mn10300__)
--- a/simde/simde-features.h
+++ b/simde/simde-features.h
@@ -348,6 +348,15 @@
   #endif
 #endif
 
+#if !defined(SIMDE_MIPS_LOONGSON_MMI_NATIVE) && !defined(SIMDE_MIPS_LOONGSON_MMI_NO_NATIVE) && !defined(SIMDE_NO_NATIVE)
+  #if defined(SIMDE_ARCH_MIPS_LOONGSON_MMI)
+    #define SIMDE_MIPS_LOONGSON_MMI_NATIVE  1
+  #endif
+#endif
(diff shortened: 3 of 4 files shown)


nemequ:
Wow, this looks fantastic, thanks!

For CI, is there an emulator we can use to run the tests, preferably something with a Debian package (maybe qemu-mips64 has a flag)?  We could do a compile-only test for Loongson, but I'd strongly prefer to be able to run them.  Also, I'd really like to make sure the [Docker development container](https://github.com/simd-everywhere/simde/tree/master/docker).

We actually have a few options for CI; we currently use [GitHub Actions](https://github.com/simd-everywhere/simde/blob/master/.github/workflows/ci.yml), [Travis CI](https://github.com/simd-everywhere/simde/blob/master/.travis.yml), [AppVeyor](https://github.com/simd-everywhere/simde/blob/master/.appveyor.yml), [Azure Pipelines](https://github.com/simd-everywhere/simde/blob/master/.azure-pipelines.yml), and [Drone.io](https://github.com/simd-everywhere/simde/blob/master/.drone.star), but I can add another (Semaphore, CirceCI, etc.) if there is something you're already comfortable with.

FlyGoat:
> Wow, this looks fantastic, thanks!
> 
> For CI, is there an emulator we can use to run the tests, preferably something with a Debian package (maybe qemu-mips64 has a flag)? We could do a compile-only test for Loongson, but I'd strongly prefer to be able to run them. Also, I'd really like to make sure the [Docker development container](https://github.com/simd-everywhere/simde/tree/master/docker).
> 
> We actually have a few options for CI; we currently use [GitHub Actions](https://github.com/simd-everywhere/simde/blob/master/.github/workflows/ci.yml), [Travis CI](https://github.com/simd-everywhere/simde/blob/master/.travis.yml), [AppVeyor](https://github.com/simd-everywhere/simde/blob/master/.appveyor.yml), [Azure Pipelines](https://github.com/simd-everywhere/simde/blob/master/.azure-pipelines.yml), and [Drone.io](https://github.com/simd-everywhere/simde/blob/master/.drone.star), but I can add another (Semaphore, CirceCI, etc.) if there is something you're already comfortable with, or if you can point me to an emulator we can use I should be able to set it up.

Unfortunately QEMU loongson support is not available at the time. We're upstreaming it and it's expected to come in next major version.

I'll take a look at compile test, thanks~

nemequ:
For a compile-only test, I'd suggest using Travis CI; it's the only place where we still use CMake instead of Meson, and Meson really wants to be able to run the compiled executables.

You should be able to just install the relevant packages (`gcc-mips64-linux-gnuabi64` and `g++-mips64-linux-gnuabi64`, I guess?) and set the `RUN_TESTS` environment variable to "false".  It's untested, but something like:

```
  - name: "loongson"
    if: branch != master OR type == pull_request
    arch: amd64
    compiler: mips64-linux-gnuabi64-gcc
    env:
    - C_COMPILER=mips64-linux-gnuabi64-gcc
    - CXX_COMPILER=mips64-linux-gnuabi64-g++
    - ARCH_FLAGS="-march=loongson3a"
    - RUN_TESTS=false
    addons:
      apt:
        packages:
        - gcc-mips64-linux-gnuabi64
        - g++-mips64-linux-gnuabi64
```

I was actually trying to get qemu working earlier, and `qemu-mips64el-static -cpu help` does show Loongson-2E, Loongson-2F, Loongson-3A1000, and Loongson-3A4000, at least on Debian Sid.  Maybe they're running a pre-release version?  Unfortunately every time I tried to run a MIPS executable (even just the default target instead of Loongson) I just got an "Invalid ELF image for this architecture" error which I'm not sure what to do about.  If a Debian Sid build would do the trick and I'm just missing some flags I think I can make that work…

FlyGoat:
@nemequ We should compile with mips64el-linux-gnuabi64 toolchain :-(
Is it possible to run CI in docker?

nemequ:
I got `-march=loongson3a` working with `qemu-mips64el-static -cpu Loongson-3A4000`.  Unfortunately the tests fail; I'm not sure if the problem lies with qemu or gcc.  I'm going to generate some reduced test cases, I don't suppose you have access to any hardware we could use to test just to see if the tests pass with real hardware instead of qemu?

> @nemequ We should compile with mips64el-linux-gnuabi64 toolchain :-(

Yep, you're right.  It mostly works with mips64el.

>  Is it possible to run CI in docker?

Some of the CI providers actually allow you to specify a docker image… GitHub Actions, for example, works that way (the `runs-on` propery for each job)…

nemequ:
Merged, thanks @FlyGoat!

Issue #231, opened 2020-04-24
Some functions don't have their own tests


mr-c:
I've added some in https://github.com/nemequ/simde/tree/missing_tests

- [ ] _mm256_alignr_epi8
- [ ] _mm256_castpd_ps
- [ ] _mm256_castpd_si256
- [ ] _mm256_castps128_ps256
- [ ] _mm256_castps256_ps128
- [ ] _mm256_castps_pd
- [ ] _mm256_castps_si256
- [ ] _mm256_castsi128_si256
- [ ] _mm256_castsi256_pd
- [ ] _mm256_castsi256_ps
- [ ] _mm256_castsi256_si128
- [ ] _mm256_permute2f128_pd
- [ ] _mm256_permute2f128_si256
- [ ] _mm256_undefined_pd
- [ ] _mm256_undefined_ps
- [ ] _mm256_undefined_si256
- [ ] _mm256_xor_pd
- [ ] _mm256_xor_ps
- [ ] _mm512_add_epi16
- [ ] _mm512_add_epi8
- [ ] _mm512_castpd128_pd512
- [ ] _mm512_castpd256_pd512
- [ ] _mm512_castpd_ps
- [ ] _mm512_castpd_si512
- [ ] _mm512_castps128_ps512
- [ ] _mm512_castps256_ps512
- [ ] _mm512_castps_pd
- [ ] _mm512_castps_si512
- [ ] _mm512_castsi128_si512
- [ ] _mm512_castsi256_si512
- [ ] _mm512_castsi512_pd
- [ ] _mm512_castsi512_ps
- [ ] _mm512_set1_epi16
- [ ] _mm512_set1_epi32
- [ ] _mm512_set1_epi64
- [ ] _mm512_set1_epi8
- [ ] _mm512_set1_pd
- [ ] _mm512_set1_ps
- [ ] _mm512_set4_epi32
- [ ] _mm512_set4_epi64
- [ ] _mm512_set4_pd
- [ ] _mm512_set4_ps
- [ ] _mm512_set_epi16
- [ ] _mm512_set_epi32
- [ ] _mm512_set_epi64
- [ ] _mm512_set_epi8
- [ ] _mm512_set_pd
- [ ] _mm512_set_ps
- [ ] _mm512_setr4_epi32
- [ ] _mm512_setr4_epi64
- [ ] _mm512_setr4_pd
- [ ] _mm512_setr4_ps
- [ ] _mm512_setr_epi32
[...]

PR #627, opened 2020-10-16
avx512: add some missing tests

ashnewmanjones:
I have added some of the missing tests for `set` functions mentioned in #231. I see it was discussed them not being strictly necessary however I have added them for completeness.


nemequ:
Excellent, thank you!

They look good to me, I'll get the merging process started.

PR #628, opened 2020-10-28
New package name for intel oneapi beta10

rscohn2:


nemequ:
It's really to see someone from Intel reaching out ahead of a change like this instead of just letting our build break; thank you!

PR #629, opened 2020-10-30
avx512bw: Implement madd functions

ashnewmanjones:

PR #630, opened 2020-10-30
implement maddubs functions

ashnewmanjones:


simde/x86/avx512/maddubs.h
```diff
@@ -0,0 +1,156 @@
...
+
+    #if SIMDE_NATURAL_VECTOR_SIZE_LE(256)
+      for (size_t i = 0 ; i < (sizeof(r_.m256i) / sizeof(r_.m256i[0])) ; i++) {
+        r_.m256i[i] = simde_mm256_maddubs_epi16(a_.m256i[i], b_.m256i[i]);
+      }
+    #else
+      SIMDE_VECTORIZE
+      for (size_t i = 0 ; i < (sizeof(r_.m256i) / sizeof(r_.m256i[0])) ; i++) {
+        r_.m256i[i] = simde_mm256_maddubs_epi16(a_.m256i[i], b_.m256i[i]);
+      }
```
nemequ, line 114: Both of these implementations are the same, but the `#else` condition shouldn't be.  `#if SIMDE_NATURAL_VECTOR_SIZE_LE(256)` will be true if the vector implementation we're using natively supports 256-bit vectors or smaller (i.e., 128-bit or 64-bit) but not 256-bit or larger.

The idea is that we want to fall back on smaller implementations until we hit the "natural" size, then provide either a native implementation or something we hope the compiler can auto-vectorize into something decent for the targeted ISA.

So the `#else` condition should basically just be a copy of the 128-bit implementation, except instead of looping over 128 bits of data we loop over 512-bits of data.  Since we actually use `sizeof` to calculate the number of elements in the data, no code changes are necessary.

I went ahead and made the change, but I wanted to explain the motivation.

nemequ (commented):
Awesome, thanks! Apart from some very trivial issues this looks great.

Aside from the small change I mentioned, I also went ahead and regenerated the test cases so all the numbers line up; it just makes it a bit easier to debug any issues since you can look at the code and see which inputs correlate to one another as well as the output a bit easier.

Also, I don't think it's documented anywhere, but for the AVX-512 functions you can leave the test-case generation code in place in case we need to regenerate it for some reason later.  It's a recent change, so most tests in AVX-512 don't follow that convention yet (and I don't think it's generally worthwhile to go back in and add the code unless we need to regenerate the tests), but we've been doing that in NEON for a while and it's working great; it's just not really an option for earlier ISAs since it would make the test files unreasonably large.  If you take a look at the version of this patch I'm merging you should see what I'm talking about.

Anyways, thanks for the patches! This one and #629 should be in merged soon ☺

simde/x86/avx512/maddubs.h
ashnewmanjones: Noted. Thanks for sorting and merging.

PR #635, opened 2020-11-09
sse2: add missing mm_cmpngt_{pd,sd}

ktgw0316:


mr-c:
Great catch! We claimed that these functions were implemented in https://github.com/simd-everywhere/simde/issues/3 but searching the git history I never did find them.

Thanks!

nemequ:
Merged, thank you!

PR #637, opened 2020-11-09
OSTYPE requires BASH

mr-c:
and -cap-add is for podman only


nemequ:
Thanks for this.  The docker container is *really* nice once you get it working, but that can be a heavy lift so anything we can do to make that easier is a very good thing ☺

PR #639, opened 2020-11-13
neon: correct C&P errors in maxnm and minnm

rosbif:
Correct _mm_max_ps to _mm_max_pd in maxnm.h
Correct _mm_min_ps to _mm_min_pd in minnm.h


nemequ:
Argh, thanks ☺

I need to think about how to test SIMDE_FAST_NANS better; right now only ICC tests it… it just feels like a waste to add a bunch more tests when only a very few functions change…

PR #645, opened 2020-11-22
avx512: add kshift.[ch]

rosbif:
Add kshift.[ch] with _kshift[lr]i_mask{8,16,32,64} intrinsics.


rosbif:
@nemequ: Maybe I have forgotten to update some files: a lot has changed in the few months I have been away.
I am not sure about the gcc version checks but I know that gcc 6.5.0 doesn't accept any of the _kshift* intrinsics.
I used the CONSTIFY macros in the test file (kshift.c) but I prefer just a simple macro in the definition file (kshift.h). You may not agree ;-)

rosbif:
In a few places I assumed that SIMDE_REQUIRE_CONSTANT_RANGE protected against out of range constants by giving a compilation error. Maybe it doesn't. Does it?

nemequ:
> In a few places I assumed that SIMDE_REQUIRE_CONSTANT_RANGE protected against out of range constants by giving a compilation error. Maybe it doesn't. Does it?

It requires clang, and in the tests the constant check is disabled for the test suite.

Passing an invalid value is undefined behavior; there is no need to handle such cases in the code, *especially* when doing so would have a performance impact.  If you're using clang (or a clang-based compiler) you'll get a compile-time error, otherwise you just get undefined behavior.

rosbif:
@nemequ In this case the performance impact is minute: I am just masking off bits in an index (bits that would never be set if the value were in range).

mr-c:
Apparently fixed in https://github.com/simd-everywhere/simde/commit/6bf0dfd33e53799e97c260e2a9038e9fddae8c39

PR #687, opened 2020-12-29
Fix git_id when amalgamating from another git repo

tlively:
To update the amalgamated SIMDe header included in Emscripten, we run a wrapper
script around amalgamate.py. Since this wrapper script is typically run from the
Emscripten directory, the git id picked up by amalgamate.py was Emscripten's git
id rather than SIMDe's git id as intended. This PR changes the working directory
of the git command run by amalgamate.py so that it correctly reads SIMDe's git
id even when run from a different git repo.


mr-c:
Thanks for catching this! I'll update my PR in a moment.

PR #695, opened 2021-01-12
Fix for incorrect MAX alignment on Windows

jsbache:
We notice that we fault when using 128 bit wide SSE/neon instructions on Windows Arm.
The problem is simde-align.h which incorrectly sets maximum alignment to 8 when compiling with Visual Studio.
The Max needs to be at least 16, but as it turns out newer versions of Visual Studio (2017 and newer) does not have any of the described issues related to alignas, and with these tools we should simply compile SIMDE as other tool chains.


nemequ:
Nice, thanks! Going through CI now, should be merged soon.

PR #696, opened 2021-01-18
common,X86,NEON: everything involving pd or si128 needs at least SSE2

rosbif:
Correct SSE to SSE2 in sse2.h, uzp1.h and uzp2.h
Also correct C&P error in simde-common.h

### DIFF
--- a/simde/simde-common.h
+++ b/simde/simde-common.h
@@ -496,7 +496,7 @@ typedef SIMDE_FLOAT32_TYPE simde_float32;
 #  define SIMDE_FLOAT64_TYPE double
 #  define SIMDE_FLOAT64_C(value) value
 #else
-#  define SIMDE_FLOAT32_C(value) ((SIMDE_FLOAT64_TYPE) value)
+#  define SIMDE_FLOAT64_C(value) ((SIMDE_FLOAT64_TYPE) value)
 #endif
 typedef SIMDE_FLOAT64_TYPE simde_float64;
 
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -7290,7 +7290,7 @@ simde_mm_unpacklo_pd (simde__m128d a, simde__m128d b) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128d
 simde_x_mm_negate_pd(simde__m128d a) {
-  #if defined(SIMDE_X86_SSE_NATIVE)
+  #if defined(SIMDE_X86_SSE2_NATIVE)
     return simde_mm_xor_pd(a, _mm_set1_pd(SIMDE_FLOAT64_C(-0.0)));
   #else
     simde__m128d_private
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - 'master'
       - 'ci/**'
       - '!ci/gha**'
+  pull_request:
+    branches:
+      - 'master'
 
 jobs:
   formatting:
--- a/simde/arm/neon/uzp1.h
+++ b/simde/arm/neon/uzp1.h
@@ -427,7 +427,7 @@ simde_vuzp1q_s32(simde_int32x4_t a, simde_int32x4_t b) {
     return t.val[0];
   #elif defined(SIMDE_WASM_SIMD128_NATIVE)
     return wasm_v32x4_shuffle(a, b, 0, 2, 4, 6);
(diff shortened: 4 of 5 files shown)


simde/x86/sse2.h
```diff
@@ -7290,7 +7290,7 @@ simde_mm_unpacklo_pd (simde__m128d a, simde__m128d b) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128d
 simde_x_mm_negate_pd(simde__m128d a) {
-  #if defined(SIMDE_X86_SSE_NATIVE)
+  #if defined(SIMDE_X86_SSE2_NATIVE)
```
mr-c, line 7293: huh, I wonder why our testing didn't catch this..

rosbif: I also get some very surprising errors in several NEON functions involving f64 if I disable SSE2.
I am still trying to understand why they occur :-(

rosbif: Actually it doesn't seem to be as simple as that.
In this case where the failing addition gives -NAN, if the same addition is repeated in the loop it seems to give the correct result. Could it be a race condition between data being loaded into a_ and b_ (or r_) and their use ?
Maybe a tiny sleep would make the problem disappear.
Any better ideas?

simde/simde-common.h
```diff
@@ -496,7 +496,7 @@ typedef SIMDE_FLOAT32_TYPE simde_float32;
 #  define SIMDE_FLOAT64_TYPE double
 #  define SIMDE_FLOAT64_C(value) value
 #else
-#  define SIMDE_FLOAT32_C(value) ((SIMDE_FLOAT64_TYPE) value)
+#  define SIMDE_FLOAT64_C(value) ((SIMDE_FLOAT64_TYPE) value)
```
nemequ, line 499: This is the one I'm surprised testing didn't catch… with -Weverything I'm surprised we didn't see a *ton* of -Wdouble-promotion diagnostics.

simde/x86/sse2.h
nemequ: A race condition wouldn't make sense; the code isn't multi-threaded (the tests are run in parallel, but the parallelism is process-level not thread-level).

It sounds like a compiler bug, which I wouldn't really find surprising here.  I'm guessing you're targeting x86_64, and the baseline there includes SSE2 so I have a feeling -mno-sse2 isn't particularly well-tested.  If I can reproduce I'll try to set c-reduce loose on it and see what it comes up with.

rosbif: The sleep didn't help.
However if I simply duplicate the  `r_.values[i] = a_.values[i] + b_.values[i];` line:
- The error goes away in C
- I get a different equally strange error in C++:
`test/arm/neon/add.cpp:638: assertion failed: r[0] ~= simde_vld1q_f64(test_vec[i].r)[0] (99.060000 ~= 99.060000)`

Incidentally I force the portable code with -DSIMDE_NO_NATIVE but the initial error is also present without it.

nemequ: > Concerning the race condition: I imagined that something might cause the hardware to ignore the latency of loading from memory. IIRC some asynchronous load and cache pre-load instructions exist.

Thanks for explaining, that makes more sense though it would still be a compiler bug since it violates sequential consistency.

> However if I simply duplicate the r_.values[i] = a_.values[i] + b_.values[i]; line:

Good data point, thanks.  If manually unrolling the loop fixes the issue in C I'm even more convinced of a compiler bug.

> I get a different, but equally strange, error in C++:

Yep, I've seen that before. Compiler bug.  I have two others to try to generate a reduced test case for, I'll try to take a look at this one once those are done.  Thanks for letting me know.

rosbif: This is the list of functions with a similar problem, all in the *q_f64 intrinsic:
`arm/neon/{add,bsl,ceqz,cgez,cle,clez,clt,cltz,min,mls,mul,mul_n,padd,sub}`

rosbif: My impression is that sometimes, but consistently, some values are read incorrectly the first time but a subsequent read gives the correct value. This would also explain why two apparently equal values are declared unequal in the assertion.

nemequ:
Sorry, I misread the notification and thought Michael had merged this already. It should be in shortly.  And, of course, thanks ☺

rosbif:
@nemequ NP. I also misread the notification :-(

PR #698, opened 2021-01-22
Use SIMDE_REQUIRE_RANGE instead of SIMDE_REQUIRE_CONSTANT_RANGE for s…

alexeyvo:
…imde_mm_srai_epi16 and simde_mm_srai_epi32

It looks like const is not needed - vshlq_s16 and vshlq_s32 dont require const argument


nemequ:
Sorry, this got a bit more verbose than I intended. It's probably TMI, but writing it helped me think through the issue… feel free to stop reading if you get bored ;)

Thanks for the patch!  Unfortunately, it's not quite that simple (why are things never simple?!?!?).  There are some subtle differences that can effect performance and create a trap which I think a lot of people are likely to fall into :(

> It looks like const is not needed - vshlq_s16 and vshlq_s32 dont require const argument

Thanks for calling my attention to this; it should really be using `vshrq_n_s16` for NEON instead of `vshrq_s16`.  I'll fix that soon.

That said, the constant requirement isn't actually about the NEON implementation, it's because the parameter is an 8-bit immediate value in SSE2.  Some compilers may allow you to pass a non-constant value *as an extension*, but others don't.

For compilers which *do* allow you to specify a non-constant value, the `_mm_srai_epi16(a, count)` basically becomes `_mm_sra_epi16(a, _mm_cvtsi32_si128(count))`, which has very different performance characteristics.  Even if we discount the `_mm_cvtsi32_si128` (which generates a `movd` instruction) it's quite different; [`psraw xmm xmm`](https://uops.info/html-instr/PSRAW_XMM_XMM.html) (the non-constant version) has worse latency and throughput than [`psraw xmm i8`](https://uops.info/html-instr/PSRAW_XMM_I8.html), and of course you may add some pressure on the xmm registers.  On the NEON side I'm not sure what the performance implications of the immediate form vs. the dynamic one are (I'm not aware of a convenient resource for timing data on Arm like uops.info), but at the very least you have to add a `vdupq_n_s16` / `dup`.  Basically, allowing this can lead to performance regressions which are hard to see from just looking at the code.

If you want, you can define `SIMDE_NO_CHECK_IMMEDIATE_CONSTANT` prior to including SIMDe and SIMDe will disable the check, but you should be aware that you're likely to lose some portability. Unfortunately, once I switch the code to `vshrq_n_s16` that may no longer work on NEON (SIMDe won't generate an error, but `vshrq_n_s16` might), at least unless #699 happens (in which case you could define the new macro instead of `SIMDE_NO_CHECK_IMMEDIATE_CONSTANT`).

Another option would be to change your code to use `_mm_sra_epi16(a, _mm_cvtsi32_si128(imm8))`. This has the benefit of working on all compilers, and makes it clear that this is really two instructions, as well as which variants of the two instructions you're using.

Interestingly, this has some curious performance implications on other architectures which I didn't really think about until now, but make me think that providing a good way to disable the check might be good idea… on Arm, for example, if you do `_mm_sra_epi16(a, _mm_cvtsi32_si128(imm8))`, SIMDe should convert the `_mm_cvtsi32_si128` to a `vmovq_s32`, and `_mm_sra_epi16` to a shuffle (to broadcast the first element to all lanes) and a `vshrq_s16`.  However, a `vdupq_s16` and a `vshrq_s16` would be better.

The difference is that SSE doesn't have a great way to broadcast a single element to all lanes (`_mm_set1_epi16` is basically a mov and a shuffle), and the `_mm_sra_epi16` shifts all elements by the value in the first 32-bit lane.  `_mm_srav_epi16`, which is equivalent to `vshrq_s16`, isn't available until AVX-512VL+BW. I suspect that clang (and possibly GCC) can handle this elegantly, but I wouldn't hold my breath for MSVC. Basically, it's a mismatch between NEON and SSE which we could smooth out by allowing a non-constant value, but then the code isn't necessarily compatible with true SSE implementations :(

rosbif:
@nemequ Thanks for pointing out that the` _mm_set1_*` intrinsics generate a sequence of instructions. They are so simple that I have always assumed that they were native intrinsics and I had never bothered to check. For zero there are` _mm_setzero_*` intrinsics (which in fact generate ` _mm_xor_*(foo, foo)`). For 0xFF there are `_mm_cmpeq_*(foo, foo)` but I have always been afraid that the compiler (or valgrind) will complain that foo is uninitialized if no suitable foo is available. Do you know whether the compiler will generate this automatically for 0xFF?

nemequ:
> They are so simple that I have always assumed that they were native intrinsics and I had never bothered to check

I'm not surprised. I've honestly never understood why Intel didn't add instructions; it's such a common operation, I would think it would be worthwhile…

> Do you know whether the compiler will generate this automatically for 0xFF?

I haven't checked every compiler, but yeah at least clang, GCC, and MSVC will generate a cmpeq instruction for `set1(~0)` and have done so for a really long time. It shoud be safe to use that in your code.

alexeyvo:
Thanks for explanation. In practice all major compilers(tested on gcc, clang, msvc. intel) allows to pass a non-constant value. I use SIMDE to port my project on apple silicon and only one real issue I faced is compile-time const requirement. To solve it I just define new functions using vshlq + vdupq, btw #699 would be helpful.

nemequ:
Thank you very much for the patch.  I'm sorry really sorry it took so long for me to merge it.  I'm not sure how this slipped through the cracks, but it's in now!

PR #703, opened 2021-02-14
neon/rhadd: optimizations for rhaddq_xxx

aqrit:
* Added or improved SSE2 implementations.
* Better generic fallback formula for unsigned averages.
* Improved WASM path for s8 & s16.

### DIFF
--- a/simde/arm/neon/rhadd.h
+++ b/simde/arm/neon/rhadd.h
@@ -25,6 +25,20 @@
  *   2020      Sean Maher <seanptmaher@gmail.com> (Copyright owned by Google, LLC)
  */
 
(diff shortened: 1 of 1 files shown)


nemequ:
Nice, thanks! I'll take a closer look at this tomorrow, but from a quick skim it looks good to me.

The CI failure wasn't from anything you did, it was unable to download the Intel Software Development Emulator (SDE). Intel probably just changed their site in a way that broke the script to download it :(

I should really re-read Hacker's Delight while keeping an eye out for anything that could apply to SIMDe... I have a feeling there would be several tricks we could use.

nemequ:
Merged, thanks!

Issue #704, opened 2021-02-16
x_mm_abs_ps and similar, lots of problems


jpcima:
Using simde `v0.7.2`, clang 11.0.1, gcc 10.2.0

---
1. `x_mm_abs` intrinsics don't always produce efficient code on SSE

In these cases, the generic path is taken in both functions `simde_x_mm_abs_ps` / `simde_vabsq_f32`.
- Under `clang -O3`, the `x_mm_abs_ps` produces an inefficient chunk of a dozen insns
- Under `clang -O3`, the equivalent `simde_vabsq_f32` manages to reduce to a single insn, surprisingly
- Under `gcc`, the efficient code is obtained in `-O3` only, `-O2` produces a large chunk

A candidate implementation, masking the sign bit, might be
```
    union { uint32_t u32; float f32; } u_;
    u_.u32 = UINT32_C(0x7FFFFFFF);
    return _mm_and_ps(_mm_set1_ps(u_.f32), a);
```
It reduces to the single `ANDPS` instruction. It can replace the `AVX512F_NATIVE` one.

---
 2. Similar remarks apply to `m128d` variant `simde_x_mm_abs_pd`, but as well:

 - NEON and Altivec implementations are looking extremely suspicious. (don't have hw to check)
eg. `r_.neon_f32 = vabsq_f32(a_.neon_f32)`

Applied on 32bit floats, the operation would mask out 4 sign bits rather that 2, which likely results in zeroing 2 innocent bits.

- Similar to above, one may provide a candidate `pd` implementation with use of SSE only
by reinterpreting the register as a `m128` and perform `ANDPS` with an alternance of bitmasks `(0xFFFFFFFF, 0x7FFFFFFF, 0xFFFFFFFF, 0x7FFFFFFF)`.

rosbif:
> A candidate implementation, masking the sign bit, might be
> 
> ```
>     union { uint32_t u32; float f32; } u_;
>     u_.u32 = UINT32_C(0x7FFFFFFF);
>     return _mm_and_ps(_mm_set1_ps(u_.f32), a);
> ```

Rather than passing via a union, which is dodgy if you follow the C standard strictly, maybe someting like this would be better (untested);
`return _mm_and_ps(_mm_castsi128_ps(_mm_set1_epi32(INT32_MAX)), a);`

rosbif:
And for the 64-bit one something like (again untested):
[...]

PR #705, opened 2021-02-16
Fix and improve abs functions on SSE

jpcima:
As described in #704

### DIFF
--- a/simde/arm/neon/abs.h
+++ b/simde/arm/neon/abs.h
@@ -220,9 +220,11 @@ simde_vabsq_f32(simde_float32x4_t a) {
     return vec_abs(a);
   #elif defined(SIMDE_WASM_SIMD128_NATIVE)
     return wasm_f32x4_abs(a);
-  #elif defined(SIMDE_X86_AVX512F_NATIVE) && \
-      (!defined(HEDLEY_GCC_VERSION) || HEDLEY_GCC_VERSION_CHECK(7,0,0))
-    return _mm512_castps512_ps128(_mm512_abs_ps(_mm512_castps128_ps512(a)));
+  #elif defined(SIMDE_X86_SSE_NATIVE)
+    simde_float32 mask_;
+    uint32_t u32_ = UINT32_C(0x7FFFFFFF);
+    simde_memcpy(&mask_, &u32_, sizeof(u32_));
+    return _mm_and_ps(_mm_set1_ps(mask_), a);
   #else
     simde_float32x4_private
       r_,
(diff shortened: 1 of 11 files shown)


jpcima:
Reverted jpcima/simde@9817f8c
because it implemented SSE with some SSE2-only intrinsics
Given that it's platform-dependent code, the union method is not really that inconvenient, or is it?

nemequ:
> Reverted [jpcima/simde@9817f8c](https://github.com/jpcima/simde/commit/9817f8c)
> because it implemented SSE with some SSE2-only intrinsics
> Given that it's platform-dependent code, the union method is not really that inconvenient, or is it?

Nope, but you could do it using SSE intrinsics: `return _mm_and_ps(_mm_set1_ps(SIMDE_MATH_NANF), a)`.  TBH I'm not sure it's worth it, though; SSE2's market penetration on x86 is basically 100% (on the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey) it actually rounds to 100%).

Just out of curiousity, were you having a problem with the AVX-512 implementation?  FP exceptions from garbage in the upper 384-bits, maybe?  I was thinking about suggesting we just use `_mm512_mask_abs_ps` with all zeros for the src parameter, but I just saw that instruction has a latency of 8 and now I'm not sure why it even exists… I'm honestly considering implementing `simde_mm512_abs_ps` using `_mm512_and_ps` in SIMDe.

jpcima:
I'll just skip on the SSE1 versions of the pd variant, but I'm adding a test case that triggers the bug about mantissa bits overwritten with 0.
The reason that I remove AVX, like you observed, it's because of latency 8.
Despite of throughput 0.5, I have intuition that SSE will be more useful in a vast majority of cases.

jpcima:
It's done, good for review.
Regarding the test, it's as said in comment: https://github.com/simd-everywhere/simde/issues/704#issuecomment-780246507
Ideally it's written as exact comparison, but I don't know how to edit simde for this, so I leave it to you.

jpcima:
Update: rebase on master, add unit tests which are able to trigger the mantissa corruption issue
Also added unit test helpers for exact float comparison.
Are these appropriate?

nemequ:
Sorry, I've been without internet access for a while :(

Thanks for tracking this down, and *especially* for finding a way to test this.  Most people would have just tried to get the code working and moved on, I'm really glad you didn't do that :)

Normally an exact test isn't a good idea because we would need a ton of digits for each floating point value in the test vectors, but for abs I think it should be safe since the error should be the same for both.  The implementation seems a bit heavy-handed, though.  Maybe instead of duplicating all the test macros/functions we could just allow people to pass `INT_MAX` for the precision and add special cases to `simde_test_f{32,64}_precision_to_slop` to just `return (slop == INT_MAX) ? SIMDE_FLOAT32_C(0.0) : (...);`, and maybe a special case for 0.0 in `simde_test_equal_f{32,64}`?

jpcima:
> we could just allow people to pass INT_MAX for the precision

The expansion `1e-##precision` expects a literal but `INT_MAX` can't necessarily be expected to be one.
Also some compilers will throw a warning if a too small float literal equals zero.

nemequ:
Hm, I thought I had removed all usage of `1e-##precision` in favor of the *_to_slop functions (in 45fa2c74f55bcbd7c5456da1a1112b822376d99d), but it looks like I missed the ones outside of test.h.  Those should be replaced.

I'm going to be *very* busy today, but I'll try to swap out the `1e-##precision` stuff and get this merged tomorrow.  Thanks again for all the work you've done on this, and sorry it took so long for me to review it.

nemequ:
Merged, thanks again, and sorry about the delay :(

PR #706, opened 2021-02-16
Workaround for GCC with missing NEON intrinsic

jpcima:
Should fix GCC A32 without `vcvtnq_s32_f32` intrinsic.
As seen also in `simde_mm_cvt_ss2si`


jpcima:
I confirm this pair of commits to fix the build on MOD Devices on `armhf` target.
`gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0)`

PR #708, opened 2021-02-21
NEON, SSSE3, AVX512: More Integer Absolute Value Implementations

aqrit:
Add SSE2 implementations and sometimes add WASM_SIMD128, AVX512VL, or ARMv7 NEON implementations.


aqrit:
`wasm_i64x2_abs` got into the proposal 2 weeks ago.
May be some time before it shows up in the wild.

nemequ:
Looks great.  Merged, thanks!

PR #711, opened 2021-03-03
gfni: add ARM, PPC and WASM implementations of *gf2p8mul* intrinsics

rosbif:
Improvements to all `*gf2p8mul*` intrinsics:-
- Add ARM, PPC and WASM implementations.
- ARM implementations now use carry-less multiplications
  and vqtbl1q_u8 or vtbl2_u8 based LUT reduction.
- Ensure fallback to 128-bit GFNI if GFNI-SSE is available.
- Improve most fallbacks.
- Add new fallbacks.


rosbif:
For the moment I have only improved the `*gf2p8mul*` intrinsics.
I still have to clean up my `*gf2p8affine*` intrinsics.

I have only implemented `*gf2p8mul*` based on carry-less multiplication for ARM.
I shall look at other architectures later.

simde/x86/gfni.h
```diff
@@ -19,27 +19,32 @@
  * SOFTWARE.
  *
  * Copyright:
- *   2020      Christopher Moore <moore@free.fr>
  *   2020      Evan Nemerson <evan@nemerson.com>
+ *   2020      Christopher Moore <moore@free.fr>
```
nemequ, line 23: ?

It's not a big deal, but you're pretty clearly the primary author of this file so I would normally expect your name to come first. Are you trying to avoid people e-mailing you or something?

Also, as long as you're tweaking the header, it's 2021 :)

rosbif: OK, I shall modify the Copyright. It seems that I had inadvertently changed the order.
I have coded an IBM z/Arch version.
I saw that you gave me a login to a machine but is there qemu support or a CI z/Arch test?
Do you mind waiting before merging?

nemequ: I seem to remember that I got it working in the docker container a few days ago; IIRC the qemu in stable just segfaults (on anything; even `int main (void) { return 0; }`), but I think testing or maybe unstable was working if I adjusted some flags.  I'll try to take a look again within the next couple days.  FWIW, we should have access to that machine for another couple months; it's a 90-day thing, and it's been on for, i think, a bit under 2 weeks.

Not at all. Just let me know when you want to merge.

rosbif: Thanks. As I wrote in gitter qemu supports vec_gfmsum on s390x :-)
However I have other problems which I explained there.

C&P from gitter:-
I have a working x/Arch version but I have been having some problems with the plumbing. I am compiling with gcc-9 and -march=z14 -mzvector. I haven't found out why but SIMDE_ZARCH_ZVECTOR_13_NATIVE doesn't get defined, so I have been forcing it. Also the simde__m128i_{to,from}_altivec_u8functions are missing so I have been using simde_memcpy. Are there different ones for z/Arch? I am using z14 because several basic operations (e.g. vec_sel, vec_cmpgt, vec_splats) are missing in z13 and I get tons of errors in e.g. sse.h with just z13. I pulled the latest version to see if these have been corrected. Unfortunately not and, worse, gcc now throws an ICE.

PR #714, opened 2021-03-05
Add z/Arch version

rosbif:
The z13 and z14 native defines aren't currently generated.
I have to force them with `-DSIMDE_ZARCH_ZVECTOR_13_NATIVE -DSIMDE_ZARCH_ZVECTOR_14_NATIVE`.
C++ compilation with g++-9 and -10 gives an ICE but gcc-9 and -10 are OK.
This is on Ubuntu 20.04 with `s390x-linux-gnu` and -`march=z14 -mzvector`.

### DIFF
--- a/simde/x86/gfni.h
+++ b/simde/x86/gfni.h
@@ -19,7 +19,7 @@
  * SOFTWARE.
  *
  * Copyright:
- *   2020      Christopher Moore <moore@free.fr>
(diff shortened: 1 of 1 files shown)


rosbif:
After rebasing I could no longer push modifications to my fork.
So I edited the file in my fork.
I didn't realize that it would create a new PR instead of adding to the previous one.
Sorry.
Do you want me to push a new version with a clean history?

PR #718, opened 2021-03-06
x86 arm/neon: Correct z/Arch versions

rosbif:
Intrinsics using f32 need at least z14.
Correct z/Arch versions in:
arm/neon/bsl.h
arm/neon/dup_n.h
x86/gfni.h
x86/sse.h


rosbif:
I pushed another commit mainly to restart the tests which have been "expected" for about 16 hours.
I don't know how to restart them otherwise :-(

nemequ:
Awesome, thanks for doing this ☺.

PR #723, opened 2021-03-08
neon/bcax: adds implementation and tests for BCAX

Glitch18:
Implemented BCAX.
Added tests using 'aarch64-gcc-10' and verified implementation


simde/arm/neon/bcax.h
```diff
@@ -0,0 +1,183 @@
...
+
+HEDLEY_DIAGNOSTIC_PUSH
+SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
+SIMDE_BEGIN_DECLS_
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde_uint8x16_t
+simde_vbcaxq_u8(simde_uint8x16_t a, simde_uint8x16_t b, simde_uint8x16_t c) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE) && defined(__ARM_FEATURE_SHA3)
+    return vbcaxq_u8(a,b,c);
```
nemequ, line 44: Small coding-style nitpick: please put spaces after the commas: `vbcaxq_u8(a, b, c)`.  Obviously this comes down to personal preference in the end, but we want to be consistent.

simde/arm/neon/bcax.h
```diff
@@ -0,0 +1,183 @@
...
+  #else
+    return simde_veorq_u8(a,simde_vandq_u8(b,simde_vmvnq_u8(c)));
```
nemequ, line 46: Obviously this works, but instead of an mnv and an and, it would be better to use [bicq](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vbicq).  This is especially good on x86 which doesn't actually have a bitwise NOT operation; we actually [have to emulate it using andnot](https://github.com/simd-everywhere/simde/blob/5560ca060a055ff3915b0fd442511c73f9507473/simde/arm/neon/mvn.h#L47) (or ternarylogic, but that's only available in AVX-512)!

nemequ (changes requested):
Overall this looks excellent, nice work ☺.  Just a couple of very minor things to clean up.

simde/arm/neon/bcax.h
Glitch18: Yes, Understood. I'll stay consistent with the formatting in my future patches.

simde/arm/neon/bcax.h
Glitch18: That is very interesting! I did not know about BICQ. But seems like it's perfect to be used here.

Glitch18:
Thank you so much! Even if the implementation was mostly straightforward, this was definitely fun to work on! I've made a commit implementing the requested changes.

nemequ:
Looks great, thanks!  Merging.

In case you're interested in something slightly more challenging:

Getting the portable implementations in place (like you did in this patch) is the first priority, but after that we like to add implementations using other instruction sets (for example, implementing these functions using SSE/AVX, WASM SIMD128, AltiVec, z/Arch, etc.).

For example, see [neon/add.h](https://github.com/simd-everywhere/simde/blob/master/simde/arm/neon/add.h).  Note that really only NEON and MMX support 64-bit vectors, so you're probably better of skipping the vadd_* functions and looking mostly at vaddq_* (they start about halfway through the file).  The [Development Resources](https://github.com/simd-everywhere/simde/wiki/Development-Resources) section in the wiki has some links to relevant documentation, and if you're interested in PPC64 or z/Arch I can give you access to a VM.

Glitch18:
That sounds like a lot of fun! I would love to work on other instruction sets too. I'll take a look at the resources you linked, and get back to you. Thanks!

PR #724, opened 2021-03-08
neon/xar: initial implementation

Glitch18:
Implements neon/xar.
Tests were generated using 'aarch64-gcc-10' and implementation was tested using 'gcc-10' and 'aarch64-gcc-10'


simde/arm/neon/xar.h
```diff
@@ -0,0 +1,67 @@
...
+HEDLEY_DIAGNOSTIC_PUSH
+SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
+SIMDE_BEGIN_DECLS_
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde_uint64x2_t
+simde_vxarq_u64(simde_uint64x2_t a, simde_uint64x2_t b, const int d)
+    SIMDE_REQUIRE_CONSTANT_RANGE(d, 0, 63) {
+  simde_uint64x2_t result;
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE) && defined(__ARM_FEATURE_SHA3)
```
nemequ, line 42: There should be a newline before the #if.

simde/arm/neon/xar.h
```diff
@@ -0,0 +1,67 @@
...
+    SIMDE_CONSTIFY_64_(vxarq_u64, result, (HEDLEY_UNREACHABLE(), result), d, a, b);
+    return result;
```
nemequ, line 44: If you put the return statement outside of the ifdef (after the #endif) it can be shared for all code paths.

simde/arm/neon/xar.h
```diff
@@ -0,0 +1,67 @@
...
+  #else
+    simde_uint64x2_private
+      r_,
+      t = simde_uint64x2_to_private(simde_veorq_u64(a,b));
+
+    SIMDE_VECTORIZE
+    for(size_t i=0;i<(sizeof(r_.values) / sizeof(r_.values[0]));i++) {
```
nemequ, line 51: Spacing: `for (size_t i=0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++)`

simde/arm/neon/xar.h
```diff
@@ -0,0 +1,67 @@
...
+      r_.values[i] = (t.values[i] >> d | t.values[i] << (64-d));
```
nemequ, line 52: `64 - d`.  Also, you should add parenthesis to group the operations:

```c
r_.values[i] = ((t.values[i] >> d) | (t.values[i] << (64 - d)));
```

It just makes the code a bit easier to read, which makes it easier to find bugs (which is the same reason I like the spacing around operators).  C is a hard language to write correct code in, so it's very important that the code is formatted in a way which makes it easy to understand the programmer's intent.

test/arm/neon/xar.c
```diff
@@ -0,0 +1,78 @@
...
+    { { UINT64_C( 7249279570232949789), UINT64_C( 6923994607331912604) },
+      { UINT64_C(14206804631696809167), UINT64_C(11142142412447028024) },
+       INT32_C(          19),
+      { UINT64_C( 6060248840644779666), UINT64_C(12039422898904142016) } }
+  };
+
+  for (size_t i = 0 ; i < (sizeof(test_vec) / sizeof(test_vec[0])) ; i++) {
+    simde_uint64x2_t a = simde_vld1q_u64(test_vec[i].a);
+    simde_uint64x2_t b = simde_vld1q_u64(test_vec[i].b);
+    simde_uint64x2_t r = simde_vxarq_u64(a, b, test_vec[i].d);
```
nemequ, line 52: If the third parameter really needs to be an immediate-mode constant you need to use the CONSTIFY macro here.  It's actually a bit surprising that this didn't result in a compile-time error on Arm.  If it *doesn't* need to be constant then you should remove the CONSTIFY macro in the implementation.

nemequ (changes requested):

simde/arm/neon/xar.h
Glitch18: Yes, understood. I'll ensure I use the correct formatting in the future.

test/arm/neon/xar.c
Glitch18: I see now. I removed the macro in `xar.h` and made the corresponding changes in `xar.c`. I ran tests for the updated function and they have passed. Adding a commit with these new changes. Thank you for the feedback!

PR #727, opened 2021-03-10
fix apple silicon gcc-10 requiring -flax-vector-conversions to compile

milot-mirdita:
I finally have access to an apple silicon machine. MMseqs2 compiled without much issues on clang however it needed this small fix to compile on homebrew gcc-10 without having to add `-flax-vector-conversions`.

### DIFF
--- a/simde/x86/sse4.1.h
+++ b/simde/x86/sse4.1.h
@@ -321,7 +321,7 @@ simde_x_mm_blendv_epi64 (simde__m128i a, simde__m128i b, simde__m128i mask) {
       mask_ = simde__m128i_to_private(mask);
 
     #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
-      mask_.u64 = vcltq_s64(mask_.i64, vdupq_n_s64(UINT64_C(0)));
+      mask_.neon_u64 = vcltq_s64(mask_.neon_i64, vdupq_n_s64(UINT64_C(0)));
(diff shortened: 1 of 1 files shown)


nemequ:
Excellent, thanks!

Huh, I can't reproduce this.  Both gcc 10 (with -Wextra) and clang 11 (-Weverything) are completely fine with it when cross-compiling, and they do hit that code…  The fix is correct, though, so I'll definitely merge it.

milot-mirdita:
It's a bit weird: It also doesn't complain in our cross compilation step: https://dev.azure.com/themartinsteinegger/mmseqs2/_build/results?buildId=857&view=logs&j=c319d949-082a-5093-2ff3-cac79c8a3f4e&t=516e54d9-0a19-564e-c1b0-dc69dabddc8f

The arm64 homebrew gcc seems to be a bit stitched together and might be closer to a gcc 11 than 10.
I found this issue that seems to be included in this gcc and seems related to vector type checking:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96377

Maybe it would have started failing with gcc 10.3?

milot-mirdita:
We cross-compile with clang on an amd64 host to arm64 target (https://dev.azure.com/themartinsteinegger/mmseqs2/_build/results?buildId=858&view=logs&j=0cf9826c-0709-569e-ee05-404d4be505a8&t=20e67daf-e478-5044-a01d-f0fc0aaa7e2e) and there it also doesn't complain. 

I tried to build a gcc snapshot but wasn't successful so who knows \*shrug\*

PR #728, opened 2021-03-12
neon/bcax: Adds WASM and SSE2 implementations

Glitch18:
Adds implementations for BCAX using SSE2 and WASM


nemequ:
LGTM, thanks!

PR #729, opened 2021-03-13
NEON: Implements CMLA and FMA.

Glitch18:
Implements CMLA and FMA from NEON. FMA also includes implementations using WASM, FMA, SSE2.

Tests were generated using 'aarch64-gcc-10'. Implementations tested using 'aarch64-gcc-10', 'gcc-10', 'emscripten' and 'intel-all-gcc-10'


simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+ simde_float32x4_t
+ simde_vcmlaq_f32(simde_float32x4_t r, simde_float32x4_t a, simde_float32x4_t b) {
+   #if defined(SIMDE_ARM_NEON_A32V8_NATIVE)
+    return vcmlaq_f32(r, a, b);
+   #else
+    simde_float32x4_private
+      r_ = simde_float32x4_to_private(r),
+      a_ = simde_float32x4_to_private(a),
+      b_ = simde_float32x4_to_private(b);
+
```
nemequ, line 75: This can be better implemented with a shuffle.  Clang's documentation explains the concept: https://clang.llvm.org/docs/LanguageExtensions.html#langext-builtin-shufflevector fairly well.

In SIMDe, we have a `SIMDE_SHUFFLE_VECTOR_` macro which abstracts away the differences between clang's `__builtin_shufflevector` and GCC's `__builtin_shuffle`.  It's not defined for compilers which don't implement the required functionality, so you can test `defined(SIMDE_SHUFFLE_VECTOR_)` to see if you should use it.  The first argument is the width in bits of each element, the second is the size of the entire vector in bytes, then the two input vectors (which are often the same if you only want to shuffle lanes within a single vector), followed by the lane numbers to copy.

In this case, the values for `a` that we want to use are `{ 0, 0, 2, 2 }`, so we can do something like this:

```c
  #if defined(SIMDE_SHUFFLE_VECTOR_)
    a_.values = SIMDE_SHUFFLE_VECTOR_(32, 16, a_.values, a_.values, 0, 0, 2, 2);
    r_.values += b_.values * a_.values;
  #else
    SIMDE_VECTORIZE
    ...
```

Both GCC and clang are very good at generating optimal code for shuffles, so that basically takes care of all the complexity of this function.  After that, all we have is a simple multiply and add.

simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+    SIMDE_VECTORIZE
+    for(size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
+      if(i % 2 == 0)
```
nemequ, line 78: First, you should prefer `& 1` to `% 2`.  Surprisingly, the doesn't *always* do the right thing even for this simple optimization.

In SIMD programming, it's very important to avoid branches.  Remember, SIMD is "single instruction multiple data"; if you have a branch you're running different instructions for different data.  Often this means that the compiler can't automatically vectorize the code.

In this case, you could use something like this for the body of the loop:

```c
r_.values[i] += b_.values[i] * a_.values[i & 2];
```

simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+ simde_float32x4_t
+ simde_vcmlaq_rot90_f32(simde_float32x4_t r, simde_float32x4_t a, simde_float32x4_t b) {
+   #if defined(SIMDE_ARM_NEON_A32V8_NATIVE)
+    return vcmlaq_rot90_f32(r, a, b);
+   #else
+    simde_float32x4_private
+      r_ = simde_float32x4_to_private(r),
+      a_ = simde_float32x4_to_private(a),
+      b_ = simde_float32x4_to_private(b);
+
```
nemequ, line 158: Similarly, you can use a shuffle for these.

It's a bit more complicated to get a single operation because sometimes you want the negated value from `b`, but you can easily take care of that by passing `-b_.values` as one of the vectors, then selecting from the first vector (elements 0 - 3) when you want the original value and the second vector (4 - 7) when you want the negated values:

```c
      a_.values = SIMDE_SHUFFLE_VECTOR_(32, 16, a_.values,  a_.values, 1, 1, 3, 3);
      b_.values = SIMDE_SHUFFLE_VECTOR_(32, 16, b_.values, -b_.values, 5, 0, 7, 2);
      r_.values += a_.values * b_.values;
```

The difference is pretty stark.  It varies by compiler and targeted ISA, but you can play with it here: https://godbolt.org/z/EzxY4z

simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+    SIMDE_VECTORIZE
+    for(size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
```
nemequ, line 160: Another small style nitpick (sorry): there should be a space between `for` and the open-parenthesis.  Same thing with the if statements.

simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+
+    return simde_float64x2_from_private(r_);
+   #endif
+ }
+ #if defined(SIMDE_ARM_NEON_A32V8_ENABLE_NATIVE_ALIASES)
+   #undef vcmlaq_f64
+   #define vcmlaq_f64(r, a, b) simde_vcmlaq_f64(r, a, b)
+ #endif
+
+ /* * rot 90 * */
```
nemequ, line 119: These should actually go in a separate file (`cmla_rot90`).

simde/arm/neon/cmla.h
```diff
@@ -0,0 +1,371 @@
...
+   #else
+    simde_float32x4_private
+      r_ = simde_float32x4_to_private(r),
+      a_ = simde_float32x4_to_private(a),
+      b_ = simde_float32x4_to_private(b);
+
+    SIMDE_VECTORIZE
+    for(size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
+      if(i % 2 == 0)
+        r_.values[i] += -(b_.values[i+1]) * a_.values[i+1];
```
nemequ, line 162: Oh, and `i + 1` instead of `i+1` :/

nemequ (changes requested):
In addition to the issues I mentioned already, it looks like there is a single space at the beginning of every like which shouldn't be there.  Some of the code is also indented incorrectly (the native implementations should be indented two spaces beyond the surrounding preprocessor directives instead of one).

This looks good overall.  Most of the issues with this are pretty trivial, and this is a good opportunity to get comfortable with using shuffles.

simde/arm/neon/cmla.h
Glitch18: This is pretty cool! Thanks, I'll add an implementation using shuffles.

simde/arm/neon/cmla.h
Glitch18: Ah I see. I think I can remove branching code from all 4 variations of the function. Thanks for pointing this out!

Glitch18:
Thank you so much for the review. I've fixed the strange indentation errors. Don't know why there was a single space at the beginning of every line. Thanks for the explanation of shuffles and how to use them! I understood them well and I've added implementations using the same. Those results at Compiler Explorer were very fascinating! 

And please feel free to point out all my formatting errors. I tried being careful of them when I first wrote these files, but I guess some of them still slipped by.

simde/arm/neon/cmla_rot90.h
```diff
@@ -0,0 +1,132 @@
...
+      a_ = simde_float32x2_to_private(a),
+      b_ = simde_float32x2_to_private(b);
+
+    #if defined(SIMDE_SHUFFLE_VECTOR_)
+      a_.values = SIMDE_SHUFFLE_VECTOR_(32, 8, a_.values, a_.values, 1, 1);
+      b_.values = SIMDE_SHUFFLE_VECTOR_(32, 8, -b_.values, b_.values, 1, 2);
+      r_.values += b_.values * a_.values;
+    #else
+      SIMDE_VECTORIZE
+      for (size_t i = 0 ; i < (sizeof(r_.values) / 2 * sizeof(r_.values[0])) ; i++) {
```
nemequ, line 53: This is what is causing the segfault on several compilers.  You're trying to write well past the end of the vector because of how you calculate the upper bound here.  `sizeof(r_.values) == 8`, `sizeof(r_.values[0]) == 4`, so what you have here is `8 / 2 * 4` = 16.  Then, in the loop, you assign to `r_.values[2 * i + 1]` (i.e., `r_.values[2 * 15 + 1]`).

Try temporarily appending ` && 0` to line 47 to disable the shuffle-based implementation, then you should hit this code in modern GCC/clang.

Glitch18: Oh boy, all I was missing was a parentheses. That was embarrassing. I think this commit should fix it now. Thanks for pointing this out.

simde/arm/neon/fma.h
```diff
@@ -0,0 +1,140 @@
...
+  #elif defined(SIMDE_X86_FMA_NATIVE)
+    return _mm_fmadd_pd_(b, c, a)
+  #elif defined(SIMDE_WASM_SIMD128_NATIVE)
+    return wasm_f64x2_add(a, wasm_f64x2_mul(b, c));
+  #else
+    simde_float64x2_private
+      a_ = simde_float64x2_to_private(a),
+      b_ = simde_float64x2_to_private(b),
+      c_ = simde_float64x2_to_private(c);
+
```
nemequ, line 123: I have a suggestion for another implementation here, this one using [GCC-style vector extensions](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html).  Vector extensions are pretty limited; you can only perform basic operations (basically stuff there is an operator for in C), but they pretty much always provide an optimal implementation if all you're doing is sticking to those basics.

Basically, you just need something like:

```c
#if defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
  a_.values += b_.values * c_.values;
#else
  SIMDE_VECTORIZE
  ...
```

Take a look at the sub functions if you want an example.

simde/arm/neon/fma.h
```diff
@@ -0,0 +1,140 @@
...
+  #undef vfma_f64
+  #define vfma_f64(a, b, c) simde_vfma_f64(a, b, c)
+#endif
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde_float64x2_t
+simde_vfmaq_f64(simde_float64x2_t a, simde_float64x2_t b, simde_float64x2_t c) {
+  #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
+    return vfmaq_f64(a, b, c);
+  #elif defined(SIMDE_X86_SSE2_NATIVE)
```
nemequ, line 112: Sorry, I just noticed an issue I missed before.

The SSE2 implementation needs to come *after* the FMA implementation.  FMA is a better fit, so we should prefer to use it if available, and since SSE2 is *always* available when FMA is available, the current order means the FMA version will never be used.

nemequ (changes requested):
Sorry, I was about to merge this until I noticed the SSE2/FMA issue; that should be corrected pre-merge.

I don't mind merging without the vector extension version; it's up to you whether you'd rather do a separate PR for that or just add it to this one, either works for me.

simde/arm/neon/fma.h
Glitch18: Done, Thanks!

PR #730, opened 2021-03-20
avx512: Implement _mm512_mask_or_pd function

simba611:


simba611:
Generated tests using intel-all-gcc10. Checked implementation on gcc10, intel-all-gcc-10, aarch-64-gcc-10.

nemequ:
The problem on emscripten (as well as [MSVC](https://ci.appveyor.com/project/nemequ/simde/builds/38318982)) is that you're assigning directly to a `simde__m512i` as if it were an array.  This only works in most compilers because they support [GCC-style vector extensions](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html), but we can't depend on that.

The reason Emscripten is failing is that, even though it supports vector extensions, it doesn't natively support `_m512i` so `simde__m512i` is a typedef [we create](https://github.com/simd-everywhere/simde/blob/d9b13227c3da210dc93b981552921d4ac32293b9/simde/x86/avx512/types.h#L279), which is a vector of `int_fast32_t`.  It seems that most compilers [implement it](https://github.com/llvm/llvm-project/blob/5cbe2279f723f1cca1a542d95e7d9760e4f52240/clang/lib/Headers/avx512fintrin.h#L31) using `long long`.

Take a look at [`_mm512_mask_xxx_epi64`](https://github.com/simd-everywhere/simde/blob/d9b13227c3da210dc93b981552921d4ac32293b9/test/x86/skel.c#L1584-L1623) in skel.c, you'll notice that instead of an `simde__m512i` in the `test_vec`, we actually have an array of `int64_t` which we then load using `simde_mm512_loadu_epi64()`.  Many older tests actually use `simde__m512i` in the test vector, but they initialize them with the result of a `simde_mm512_set_epi64` not as a normal array.  It looks like what you've done here is created a hybrid of the two approaches in a way which only works on some compilers out of luck.

Use the `simde__m512i_mask_xxx_epi64` function to create your test case, it should do the right thing.  Also, now that we've split up the AVX-512 tests into one file per family (there used to be one giant file for AVX-512F, one giant one for AVX-512BW, etc.) you can leave the code to generate new test vectors in place instead of removing it.  Just change the `#if 0` to an `#if 1` to switch from generating test vectors to verifying them.

PR #731, opened 2021-03-21
AVX512: Implement various mask/maskz functions

simba611:
implemented mm512_mask_or_ps, mm512_maskz_or_ps, mm512_maskz_or_pd, mm512_mask_xor_ps, mm512_maskz_xor_ps, mm512_mask_xor_pd, mm512_maskz_xor_pd. Also implemented mm512_mullo_epi64 with mm512_mask_mullo_epi64 and mm512_maskz_mullo_epi64. 
All tests were generated with intel-all-gcc-10 and implementation tested on emscripten, gcc-10, intel-all-gcc-10, clang-11.


nemequ:
This looks great, thanks!  I'll merge it ASAP.

The `_mm512_mullo_epi64` could be improved; a version using [GCC vector extensions](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html) would be great.  For an example (using add instead of multiply but the idea is the same), see [x86/avx512/](https://github.com/simd-everywhere/simde/blob/b7933e60ff3f368462f342490e2ac823000be76d/simde/x86/avx512/add.h#L293-L300).

Also, you should be careful about matching the element widths when testing.  For the *_ps functions this patch uses 64-bit integers for the test vectors, but those functions operate on single-precision floats, which are 32-bits.  It doesn't actually matter for these functions since bitwise operations like xor and or don't care about the element width, but it's a good habit to get in to. For example, if you tried to do this with add it would result in errors on [big endian](https://en.wikipedia.org/wiki/Endianness) architectures.

simba611:
> This looks great, thanks! I'll merge it ASAP.
> 
> The `_mm512_mullo_epi64` could be improved; a version using [GCC vector extensions](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html) would be great. For an example (using add instead of multiply but the idea is the same), see [x86/avx512/](https://github.com/simd-everywhere/simde/blob/b7933e60ff3f368462f342490e2ac823000be76d/simde/x86/avx512/add.h#L293-L300).
> 
> Also, you should be careful about matching the element widths when testing. For the *_ps functions this patch uses 64-bit integers for the test vectors, but those functions operate on single-precision floats, which are 32-bits. It doesn't actually matter for these functions since bitwise operations like xor and or don't care about the element width, but it's a good habit to get in to. For example, if you tried to do this with add it would result in errors on [big endian](https://en.wikipedia.org/wiki/Endianness) architectures.

I will work on that right away

PR #732, opened 2021-03-23
avx512:implement mm512_insertf32x8,64x2 with mask(z)

simba611:
Implemented mm512_insertf32x8, mm512_mask_insertf32x8, mm512_maskz_insertf32x8, mm512_insertf64x2, mm512_mask_insertf64x2, mm512_maskz_insertf64x2.
All tests were generated using intel-all-gcc-10. Implementation was tested on intel-all-gcc-10, emscripten, gcc-10 , clang-11


nemequ:
Looks good, thanks!

Good job dealing with CI; as you've seen it can be a bit of a pain to get everything to pass on all compilers and architectures, but a better way to look at is that it finds lots of bugs that would otherwise slip through unnoticed ☺

PR #738, opened 2021-03-25
avx512: implement mm256_mask_compress_pd

simba611:
Implemented mm256_mask_compress_pd. 
Tests were generated using intel-all-gcc-10. Implementation was checked on intel-all-gcc-10, emscripten, clang-11, gcc-10.


nemequ (changes requested):
Overall this looks good.  For 128-bit vectors you could use `_mm_shuffle_epi8` / `vtbl1q_s8` / `wasm_v8x16_shuffle` to shave off a few cycles, but that could get a bit gnarly so maybe we should circle back to that later.

Other than that, there are just a few coding style issues to clean up.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+
+SIMDE_FUNCTION_ATTRIBUTES
+simde__m256d
+simde_mm256_mask_compress_pd (simde__m256d src, simde__mmask8 k, simde__m256d a) {
+  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512F_NATIVE)
+    return _mm256_mask_compress_pd(src, k, a);
+  #else
+    simde__m256d_private
+    a_ = simde__m256d_to_private(a),
+    src_ = simde__m256d_to_private(src);
```
nemequ, line 18: These two lines (17 and 18) should be indented to make it clear that these lines are part of the declaration which begins on L16.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+    size_t j = 0;
```
nemequ, line 19: There should be a blank line between the declarations and the next block of code (i.e., between lines 19 and 20) to help make it easy to tell at a glance when the declarations end and the body begins.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+    SIMDE_VECTORIZE
+    for (size_t i = 0; i < (sizeof(a_.f64)/sizeof(a_.f64[0])); i++){
```
nemequ, line 21: There should be spaces on both sides of the semi-colons to make it easier to see where one clause ends and another begins, as well as on both sides of the division operator for similar reasons.  There should also be a space between the close-paren and the open-curly-brace.  So:

```c
for (size_t i = 0 ; i < (sizeof(a_.f64) / sizeof(a_.f64[0])) ; i++) {
```

Notice how you can kind of tell what is going on just by the groupings; you don't even have to read the code.  That's what we're after; code should always be as easy to read as we can make it.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+      if((k >> i) & 1) {
+        a_.f64[j++] =  a_.f64[i];
```
nemequ, line 23: There is an extra space here after the assignment operator (the =).  There should only be one.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+    for (; j < (sizeof(a_.f64)/sizeof(a_.f64[0])); j++){
```
nemequ, line 26: Similar to L21:

```c
for ( ; j < (sizeof(a_.f64) / sizeof(a_.f64[0])) ; j++) {
```

There should also be a blank line between the two loops to help break them up.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+      a_.f64[j] =  src_.f64[j];
```
nemequ, line 27: Similar to L23; there is an extra space here.

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return simde__m256d_from_private(a_);
```
nemequ, line 29: One more empty line here, between lines 28 and 29, just like the blank line between the loops.

nemequ (changes requested):

simde/x86/avx512/compress.h
```diff
@@ -0,0 +1,40 @@
...
+}
+#if defined(SIMDE_X86_AVX512VL_ENABLE_NATIVE_ALIASES) && defined(SIMDE_X86_AVX512F_ENABLE_NATIVE_ALIASES)
+  #undef _mm256_mask_compress_pd
+  #define _mm256_mask_compress_pd(src, k, a) _mm256_mask_compress_pd(src, k, a)
+#endif
+
+SIMDE_END_DECLS_
+HEDLEY_DIAGNOSTIC_POP
+
+#endif /* !defined(SIMDE_X86_AVX512_COPYSIGN_H) */
```
nemequ, line 40: s/COPYSIGN/COMPRESS/

simba611:
Made the changes in formatting.

PR #742, opened 2021-03-28
avx512:implement mm256_mask(z)_compress(storeu)_p*

simba611:
Implemented mm256_mask_compressstoreu_pd, mm256_maskz_compress_pd, mm256_mask_compress_ps, mm256_mask_compressstoreu_ps, mm256_maskz_compress_ps.
All tests were generated using intel-all-gcc-10. Implementation was tested on intel-all-gcc-10, emscripten, gcc-10, clang-11.

### DIFF
--- a/simde/x86/avx512/compress.h
+++ b/simde/x86/avx512/compress.h
@@ -74,17 +74,15 @@ simde_mm256_maskz_compress_pd (simde__mmask8 k, simde__m256d a) {
       a_ = simde__m256d_to_private(a);
     size_t ri = 0;
 
-    for(size_t i = 0 ; k && (i < (sizeof(a_.f64) / sizeof(a_.f64[0]))) ; i++)
-    {
-      if(k & 1){
+    SIMDE_VECTORIZE
+    for (size_t i = 0 ; i < (sizeof(a_.f64) / sizeof(a_.f64[0])) ; i++) {
+      if ((k >> i) & 1) {
         a_.f64[ri++] = a_.f64[i];
       }
-      k = k >> 1;
     }
 
-    SIMDE_VECTORIZE
-    for (size_t j = ri ; j < (sizeof(a_.f64) / sizeof(a_.f64[0])); j++) {
-      a_.f64[j] = SIMDE_FLOAT64_C(0.0);
+    for ( ; ri < (sizeof(a_.f64) / sizeof(a_.f64[0])); ri++) {
+      a_.f64[ri] = SIMDE_FLOAT64_C(0.0);
     }
 
     return simde__m256d_from_private(a_);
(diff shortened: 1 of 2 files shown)


simde/x86/avx512/compress.h
```diff
@@ -37,6 +37,152 @@ simde_mm256_mask_compress_pd (simde__m256d src, simde__mmask8 k, simde__m256d a)
...
+  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512F_NATIVE)
+    return _mm256_maskz_compress_pd(k, a);
+  #else
+    simde__m256d_private
+      a_ = simde__m256d_to_private(a);
+    size_t ri = 0;
+
+    for(size_t i = 0 ; k && (i < (sizeof(a_.f64) / sizeof(a_.f64[0]))) ; i++)
+    {
+      if(k & 1){
```
nemequ, line 79: There should be a space between the `if` and the open-paren, and another between the close-paren and open-curly-bracket (`if (k & 1) {`).

simde/x86/avx512/compress.h
```diff
@@ -37,6 +37,152 @@ simde_mm256_mask_compress_pd (simde__m256d src, simde__mmask8 k, simde__m256d a)
...
+simde__m256d
+simde_mm256_maskz_compress_pd (simde__mmask8 k, simde__m256d a) {
+  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512F_NATIVE)
+    return _mm256_maskz_compress_pd(k, a);
+  #else
+    simde__m256d_private
+      a_ = simde__m256d_to_private(a);
+    size_t ri = 0;
+
+    for(size_t i = 0 ; k && (i < (sizeof(a_.f64) / sizeof(a_.f64[0]))) ; i++)
```
nemequ, line 77: There should be a space between the `for` and the open-paren.

simde/x86/avx512/compress.h
```diff
@@ -37,6 +37,152 @@ simde_mm256_mask_compress_pd (simde__m256d src, simde__mmask8 k, simde__m256d a)
...
+simde__m256
+simde_mm256_maskz_compress_ps (simde__mmask8 k, simde__m256 a) {
+  #if defined(SIMDE_X86_AVX512VL_NATIVE) && defined(SIMDE_X86_AVX512F_NATIVE)
+    return _mm256_maskz_compress_ps(k, a);
+  #else
+    simde__m256_private
+      a_ = simde__m256_to_private(a);
+    size_t ri = 0;
+
+    for(size_t i = 0 ; k && (i < (sizeof(a_.f32) / sizeof(a_.f32[0]))) ; i++)
```
nemequ, line 165: Same thing regarding the spaces around the parenthesis.

More importantly, though, why are you checking that `k` is non-zero here but not in the other implementations?

In this case, the check is largely superflouous; you could switch to only using `k` by masking out the unused bits so you only have one comparison.  Or you can drop the check of `k` completely to make it match the other functions.

In most code my inclination would be to mask `k` so you can terminate the loop early, but based on the use case for these functions I don't think you're going to see much of an improvement from that, and it makes it a bit harder for the CPU to speculate.  I haven't benchmarked it so I can't be sure, but I think it would be better to get rid of the check on `k` and just make this match the other implementations.

nemequ (changes requested):
Most of my objections are based on coding style.  I know this may seem unimportant since they don't alter program behavior, but a *consistent* coding style is very important for keeping the code easy to read.  Code which is easier to read is easier to review, and easier to find bugs in.

Additionally, most of the style decisions aren't arbitrary (whether to add a newline between a close-parenthesis and the opening curly braces is the exception); they are designed to make the code easier read.

Beyond making it easier to identify issues, a clean, consistent coding style is also important to keeping the code maintainable.  Nobody wants to try to read messy code, and most code is read more times than it is written.  Don't take shortcuts on readability.

PR #744, opened 2021-04-02
sse4.1: _mm_blendv_epi8: add sse2 and update wasm_simd128 implementions

aqrit:


aqrit:
`psubusb` avoids the awkward `pandn` clobber.

`_mm_or_si128(_mm_subs_epu8(a, m), _mm_and_si128(m, b))` may be better if the inputs `a` and `b` are not ready.

rosbif:
Le 02/04/2021 à 23:13, aqrit a écrit :
>
> |psubusb| avoids the awkward |pandn| clobber.
>

|Sorry for being thick but please would you mind explaining what you 
mean by "awkward clobber"?
I would have used pandn because I find it more readable and, according 
to the IIG, both have the same latency but pandn has better throughput.
(It took me a couple of minutes to understand how your pxor/psubusb 
version works.)
|

aqrit:
Using `pandn` w/SSE2 would require copying one of input values to scratch register.
We expect the the function to be inlined, so we don't want to change (clobber) the inputs: `mask` or 'b'.

New(ish) processors have "free" move elimination, 
AVX2 instruction coding has "three-operand forms" that don't clobber one of the inputs,
but here we're targeting old processors.

So the extra move required consumes an extra register and adds ~6 bytes of code size.
It also adds 1 cycle of latency to the function because the mask calculation is probably on the critical path.

aqrit:
This is what I imagine it would look like... I haven't run it thru a disassembler.
```
/**
xmm0 // `a`
xmm1 // 'b'
xmm2 // 'mask'
xmm3 // scratch
xmm3 // scratch2
*/

// v1
PXOR     xmm3, xmm3 // anytime
PCMPGTB  xmm3, xmm2 // cycle 1
PXOR     xmm0, xmm1 // cycle 1
PSUBUSB  xmm0, xmm3 // cycle 2
PXOR     xmm0, xmm1 // cycle 3

// v2
PXOR     xmm3, xmm3 // anytime
PCMPGTB  xmm3, xmm2 // cycle 1
PSUBUSB  xmm0, xmm2 // cycle 2
PAND     xmm2, xmm1 // cycle 2
POR      xmm0, xmm2 // cycle 3

// v3
PXOR     xmm3, xmm3 // anytime
PCMPGTB  xmm3, xmm2 // cycle 1
MOVDQA   xmm4, xmm3 // cycle 2 
PAND     xmm3, xmm1 // cycle 2
PANDN    xmm4, xmm0 // cycle 3
POR      xmm3, xmm4 // cycle 4
// movdqa xmm0, xmm3...
```

rosbif:
@aqrit Thank you for your detailed explanation.
I have always used the intrinsics and never really looked at the assembly code.
The IIG gives the false impression that these are three operand instructions even in SSE2 e.g. for _mm_andnot_si128 :-
> Compute the bitwise NOT of 128 bits (representing integer data) in a and then AND with b, and store the result in dst.

It doesn't really make it clear that an extra instruction will be generated if a needs to be preserved.

rosbif:
Thinking further about it wouldn't something like this work:
```
// v4
PXOR     xmm3, xmm3 // anytime
PCMPGTB  xmm3, xmm2 // cycle 1
PAND     xmm1, xmm3 // cycle 2
PANDN    xmm3, xmm0 // cycle 2
POR      xmm3, xmm1 // cycle 3
```

However it leaves the result in xmm3 rather than xmm0 but I am not sure whether this is a problem.
If it is a problem then I think that the register assignments of a and b could be inverted and POR could then leave the result in xmm0 (which in this case was b on entry) like this:

```
/**
xmm0 // `b`
xmm1 // 'a'
xmm2 // 'mask'
xmm3 // scratch
*/

// v5
PXOR     xmm3, xmm3 // anytime
PCMPGTB  xmm3, xmm2 // cycle 1
PAND     xmm0, xmm3 // cycle 2
PANDN    xmm3, xmm1 // cycle 2
POR      xmm0, xmm3 // cycle 3

```

rosbif:
Sorry, please ignore this.
I suppose that to be the equivalent of the blendv, I mustn't modify xmm1 or xmm2 and I must return the result in xmm0.
Sorry for the noise :-(

nemequ:
LGTM, thanks!  Should be merged soon.

And nice explanation; pressure on the registers aside, I always forget to consider code size when I think about this stuff :(

nemequ:
LGTM, thanks!  Should be merged soon.

And nice explanation; pressure on the registers aside, I always forget to consider code size when I think about this stuff :(

PR #750, opened 2021-04-26
neon/cge: Implement remaining functions in CGE family

Glitch18:
Implements:

- vcged_f64
- vcged_s64
- vcged_u64
- vcged_f32

Tests included

### DIFF
--- a/simde/arm/neon/cge.h
+++ b/simde/arm/neon/cge.h
@@ -671,6 +671,62 @@ simde_vcge_u64(simde_uint64x1_t a, simde_uint64x1_t b) {
   #define vcge_u64(a, b) simde_vcge_u64((a), (b))
 #endif
 
+SIMDE_FUNCTION_ATTRIBUTES
+uint64_t
+simde_vcged_f64(simde_float64_t a, simde_float64_t b){
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+    return vcged_f64(a, b);
+  #else
+    return (a >= b) ? UINT64_MAX : 0;
+  #endif
+}
+#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)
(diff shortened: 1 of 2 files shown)


nemequ:
Looks good, thanks ☺.  Usually it's a good idea to use the scalar functions to implement the vector functions (e.g., the loop in `vcgeq_s64` would call `vcged_s64`) to avoid duplicating the logic unnecessarily; everything gets inlined anyways, so there is no performance penalty.  That said, in this case the functions are simple enough that I'm not going to worry about it and I'll just go ahead and merge it as-is.

Glitch18:
Ah I see. Thanks for the explanation! I'll keep that in mind for next time.

nemequ:
It looks like [this fails](https://cloud.drone.io/simd-everywhere/simde/1969/2/2) on clang prior to 12.0 due to the return value of these functions being signed instead of unsigned.  Clang has [fixed the issue](https://github.com/llvm/llvm-project/commit/f93514545cd91b132fe987618488b8c1e5388fb0), but we still need to work around the issue since we have to work with older versions of clang.

Luckily the work-around is pretty trivial; you can just change the native implementations to cast the result to the proper type.  For example, instead of `return vcged_s64(a, b);`, `return HEDLEY_STATIC_CAST(uint64_t, vcged_s64(a, b));`.

Glitch18:
Done. Included those changes in this commit.

nemequ:
`vcges_f32` is cast to `uint64_t`, but the return value is 32 bits.

Glitch18:
Ahh my bad. Didn't see that. Fixing it.

nemequ:
Np. That's a big part of why we have such an aggressive CI setup; stuff like that is *really* easy to miss.

Glitch18:
Yes, I see that now. The CI is definitely very impressive.

PR #752, opened 2021-04-29
neon/vld2_u8: native and portable implementation

ngzhian:
Drive-by fix for neon test generation, it was outputting to stdout, use the macro SIMDE_CODEGEN_FP instead.


simde/arm/neon/ld2.h
```diff
@@ -0,0 +1,66 @@
...
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Copyright:
+ *   2020      Evan Nemerson <evan@nemerson.com>
```
nemequ, line 24: Can you change this to you and/or Google?

simde/arm/neon/ld2.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+#if !defined(SIMDE_ARM_NEON_LD2_H)
```
nemequ, line 27: The failure on ICC is due to a bug in the compiler where it can't handle these types.  I actually thought this was fixed in a recent version, but I just tried enabling the code which triggers the issue and there is still an ICE (I think it may be a slightly different issue, but the report was lost when Intel changed their forums… I'll report the new issue as soon as I have reduced test case).

Anyways, to work around it you can just change this to `#if !defined(SIMDE_ARM_NEON_LD2_H) && !defined(SIMDE_BUG_INTEL_857088)`, and you'll have to ifdef out the tests too.  See ld3 and/or trn for examples.  Sorry, I know it's annoying… if you want me to take care of it instead just let me know, I don't mind.

nemequ (changes requested):
Looks pretty good to me. You beat me to the declare-suites.h issue (it only manifests in cmake, which we only use when testing old distros), so really the big thing is the ICC bug.

PR #753, opened 2021-04-29
wasm: update names of shuffle instructions

ngzhian:
Used a sed script for this

sed -i 's/wasm_v\([0-9x]\+\)_shuffle/wasm_i\1_shuffle/g' \
  $(ag 'wasm_v[0-9x]+_shuffle' -l)

TOT emscripten has deprecation warnings for the old names, so this
should quieten the builds down when we re-enable emscripten testing.


ngzhian:
Oh, i missed this 20682c192f1c7e9fd0d0dc653df74f7cc7f4a83a, should have pulled first.

nemequ:
Sorry, the codebase doesn't usually move fast enough for conflicts like this, just bad luck :(

ngzhian:
Not at all, thanks for fixing this and actively keeping up with Emscripten :)

PR #758, opened 2021-05-05
neon/rsqrts: vrsqrts_f32 and vrsqrtsq_f32 native and portable

ngzhian:


nemequ:
Nice, thanks.  Looks like Microsoft's mirrors are down so CI isn't working, but once it is back I'll merge this (probably some time today).

I did make a few changes.  First, instead of dividing by 2 I switched to multiplying by 0.5.  That's not really a big deal since any decent optimizing compiler will do the same (division is slow), but why risk it?

I also switched from the loop to using existing functions, which meant I could use vmls instead of a separate multiply and subtract.  Since mls has no better availability on Arm than rsqrts it probably doesn't matter much there, but on x86 mls can be implemented using _mm_fnmsub_*, which should be a respectable boost.  Unfortunately I think x86 and arm are the only platforms which allow subtraction *of* the product instead of *from* the product (e.g., `vec_msub` on PPC/s390x won't work).

ngzhian:
Nice, thanks for the improvements to implementation!

PR #763, opened 2021-05-07
avx512: implement mm512_unpacklo_* functions

simba611:
Implemented mm512_unpacklo_epi32, mm512_unpacklo_epi64, mm512_unpacklo_ps, mm512_unpacklo_pd. All tests were generated using intel-all-gcc-10 and implementation was checked on intel-all-gcc-10, emscripten, gcc-10, clang-11.


simde/x86/avx512/unpacklo.h
```diff
@@ -98,6 +98,114 @@ simde_mm512_unpacklo_epi16 (simde__m512i a, simde__m512i b) {
...
+    simde__m512i_private
+      r_,
+      a_ = simde__m512i_to_private(a),
+      b_ = simde__m512i_to_private(b);
+
+    #if defined(SIMDE_SHUFFLE_VECTOR_)
+      r_.i32 = SIMDE_SHUFFLE_VECTOR_(32, 64, a_.i32, b_.i32,
+                                    0, 16, 1, 17, 4, 20, 5, 21,
+                                    8, 24, 9, 25, 12, 28, 13, 29);
+    #else
```
nemequ, line 116: I know you've just replicated what the other functions look like here, but this should really be conditional on `SIMDE_NATURAL_VECTOR_SIZE_LE(256)`, with a "normal" portable implementation (a `SIMDE_VECTORIZE` loop) in an additional else condition.

The reason for that is that if we are on a machine with a natural vector length of 512-bits but don't have AVX-512BW (for example, Fujitsu's A64FX has 512-bit SVE units) we still want to be able to take advantage of the full width of the vector processor.  With what you have here right now, machines like that would fall back on 2 half-vector versions.

simba611: I understand. Will add that. Thanks

nemequ:
Looks great, thanks!

PR #770, opened 2021-05-11
avx512: implement mm512_range_ps/d functions

simba611:
Implemented mm512_range_pd, mm512_range_ps. 
All tests were generated using intel-all-gcc-10 and implementations were tested on intel-all-gcc-10, gcc-10, clang-11, emscripten.

### DIFF
--- a/test/x86/avx512/cmp.c
+++ b/test/x86/avx512/cmp.c
@@ -375,77 +375,272 @@ test_simde_mm512_cmp_ps_mask (SIMDE_MUNIT_TEST_ARGS) {
 static int
 test_simde_mm256_cmp_ps_mask (SIMDE_MUNIT_TEST_ARGS) {
 #if 1
-  static const struct {
-    const simde_float32 a[8];
-    const simde_float32 b[8];
-    const int imm8;
-    const simde__mmask8 r;
-  } test_vec[] = {
-    { { SIMDE_FLOAT32_C(   442.30), SIMDE_FLOAT32_C(   406.52), SIMDE_FLOAT32_C(  -742.16), SIMDE_FLOAT32_C(  -146.21),
-        SIMDE_FLOAT32_C(   589.11), SIMDE_FLOAT32_C(  -396.17), SIMDE_FLOAT32_C(   783.76), SIMDE_FLOAT32_C(   644.46) },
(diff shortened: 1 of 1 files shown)


simde/x86/avx512/range.h
```diff
@@ -0,0 +1,118 @@
...
+simde_mm512_range_pd (simde__m512d a, simde__m512d b, int imm8)
+    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 15) {
+  #if defined(SIMDE_X86_AVX512DQ_NATIVE)
+    simde__m512d r;
+    SIMDE_CONSTIFY_16_(_mm512_range_pd, r, (HEDLEY_UNREACHABLE(), simde_mm512_setzero_pd ()), imm8, a, b);
+    return r;
+  #else
+    simde__m512d r;
+
+    if (imm8 & 1) {
```
nemequ, line 30: This would be a lot easier to follow using a switch.  Not only that, but we want to avoid branching, and compilers often have an easier time optimizing a switch to do that.

```c
switch (imm8) {
  case 1:
    ...
    break;
  case 2:
    ...
    break;
}
```

nemequ (changes requested):
Looks good overall; you did a great job of making use of other functions.

One thing I would suggest is doing the 256 and 128-bit versions at the same time.  That will allow you to add a fallback on those functions to the wider versions.  I know that they aren't available until AVX-512 on Intel, structuring it that way is actually much better when emulating AVX-512 using shorter vectors.  As things currently are, each function you call will extract 256-bit or 128-bit slices of the buffer, perform the requested operation on each one, then combine the results together into a 512-bit vector.  However, if you use the 256/128-bit fallbacks, that extraction and combination only happens once; you extract a slice, call *several* functions on that data, then combine the results.

simba611:
Implemented mm_range_pd, mm_range_ps, mm256_range_pd, mm256_range_ps, mm_cmp_pd_mask, mm_cmp_ps_mask, mm256_cmp_pd_mask, mm256_cmp_ps_mask.
All tests were generated using intel-all-gcc-10 and implementations were tested in intel-all-gcc-10, gcc-10, clang-11, emscripten.
Also updated mm512_range_pd, mm512_range_ps to use switch statement and add support for fallback on mm256_range functions.

PR #773, opened 2021-05-13
neon/vld1q_dup: native and portable implementations

ngzhian:


nemequ:
As I mentioned on chat, this can be simplified by just using `return simde_vdup_n_*(*ptr);` for most of the portable implementations instead of having something like `_mm_set1_epi8(*ptr)`; it will do the same thing (`simde_vdup_n_s8` will call `_mm_set1_epi8`), but that way we can take advantage of all the other dup_n implementations; for example, AltiVec, z/Arch, WASM, and hopefully more in the future.

I also split the code out into a separate file since ld1_dup really doesn't do the same thing as ld1, and went ahead and added WASM versions of these.  The combination of falling back on dup_n and having both WASM and SSE implementations is a bit awkward in the code, but I don't have a better idea ☹…

PR #783, opened 2021-05-19
neon/ext: add __builtin_shufflevector implementation

ngzhian:


ngzhian:
Huh, should I just silence this diagnostic?
```
../arm/neon/ext.c:905:24: error: use of GNU statement expression extension [-Werror,-Wgnu-statement-expression]
```

nemequ:
No need, you can just use `__extension__` instead:

```c
#define FOO(bar) (__extension__ ({ ... }))
```

ngzhian:
Oh, so basically the same as the GCC part above, cool! I didn't find this in https://clang.llvm.org/docs/LanguageExtensions.html or in https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs.
I'll upload a fix, thanks!

nemequ:
> Oh, so basically the same as the GCC part above, cool! I didn't find this in https://clang.llvm.org/docs/LanguageExtensions.html

Yeah, clang mostly only documents clang's extensions, not GNU extensions; they basically just punt on that and tell you to read the GCC manual :/

> or in https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs.

It's not actually specific to statement exprs, so it's not documented there.  It can be used anywhere to indicate that the expression should not emit warnings about being GNU extensions.  For example, if you wanted to use C11 generic selections in C99 mode, normally with -Wpedantic you would get a warning.  For example:

```c
#define foo(x) (_Generic((x), char: foo_char, short: foo_short, int: foo_int)(x))
```

If you use this in C99 mode you may get a warning.

However,

```c
#define foo(x) (__extension__ _Generic((x), char: foo_char, short: foo_short, int: foo_int)(x))
```

Will compile quietly, at least on GCC.  Clang arguably doesn't implement it right, and you'll get a -Wc11-extensions diagnostic (if enabled)… I think clang's thinking is that it's a C11 extension not a GNU extension, and `__extension__` is for GNU extensions, but accepting C11 code in C99 mode is a GNU C extension so IMHO it should work.  Statement exprs, OTOH, are pretty clearly exclusively a GNU extension, though, so even clang is fine with just __extension__.

FWIW, the documentation is a bit terse but it's at https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html if you're interested.

PR #784, opened 2021-05-19
neon: complete implementation of CGEZ family

Glitch18:
Implements:

- vcgezd_f64
- vcgezd_s64
- vcgezs_f32

Tests included. 
The new scalar functions are used to implement the vector functions. Hence avoiding duplication of logic unnecessarily.

### DIFF
--- a/simde/arm/neon/cgez.h
+++ b/simde/arm/neon/cgez.h
@@ -37,6 +37,48 @@ SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
 SIMDE_BEGIN_DECLS_
 
 SIMDE_FUNCTION_ATTRIBUTES
+uint64_t
+simde_vcgezd_f64(simde_float64_t a) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+    return HEDLEY_STATIC_CAST(uint64_t, vcgezd_f64(a));
+  #else
+    return (a >= SIMDE_FLOAT64_C(0.0)) ? UINT64_MAX : 0;
+  #endif
+}
+#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)
+  #undef vcgezd_f64
+  #define vcgezd_f64(a) simde_vcgezd_f64(a)
+#endif
+
+SIMDE_FUNCTION_ATTRIBUTES
+uint64_t
+simde_vcgezd_s64(int64_t a) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+    return HEDLEY_STATIC_CAST(uint64_t, vcgezd_s64(a));
+  #else
+    return (a >= 0) ? UINT64_MAX : 0;
+  #endif
+}
+#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)
+  #undef vcgezd_s64
+  #define vcgezd_s64(a) simde_vcgezd_s64(a)
+#endif
+
+SIMDE_FUNCTION_ATTRIBUTES
+uint32_t
+simde_vcgezs_f32(simde_float32_t a) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+    return HEDLEY_STATIC_CAST(uint32_t, vcgezs_f32(a));
+  #else
+    return (a >= SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
+  #endif
+}
+#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)
+  #undef vcgezs_f32
+  #define vcgezs_f32(a) simde_vcgezs_f32(a)
+#endif
+
+SIMDE_FUNCTION_ATTRIBUTES
 simde_uint32x4_t
 simde_vcgezq_f32(simde_float32x4_t a) {
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
@@ -52,7 +94,7 @@ simde_vcgezq_f32(simde_float32x4_t a) {
     #else
       SIMDE_VECTORIZE
       for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-        r_.values[i] = (a_.values[i] >= SIMDE_FLOAT32_C(0.0)) ? UINT32_MAX : 0;
+        r_.values[i] = simde_vcgezs_f32(a_.values[i]);
       }
     #endif
 
@@ -80,7 +122,7 @@ simde_vcgezq_f64(simde_float64x2_t a) {
     #else
       SIMDE_VECTORIZE
       for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-        r_.values[i] = (a_.values[i] >= SIMDE_FLOAT64_C(0.0)) ? UINT64_MAX : 0;
+        r_.values[i] = simde_vcgezd_f64(a_.values[i]);
       }
     #endif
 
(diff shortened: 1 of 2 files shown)


nemequ:
Nice.  Normally this would be the right way to go, but in this case I think we might be able to do a bit better.

Most SIMD ISAs don't have comparison operations against zero, but they do have instructions to compare two vectors, and that's what we're going to want to do on those targets.  On NEON, that's the `cge` family.

To get a vector of all zeros to compare against, most SIMD ISAs are going to have some sort instruction to duplicate an element to all other lanes.  In NEON, that's the `dup_n` family.

So, for the fallback implementations what we could do is, using `float32x4_t` as an example, `return simde_vcgeq_f32(a, simde_vdupq_n_f32(SIMDE_FLOAT32_C(0.0)));`.  That lets us take advantage of the fast implementations in `cge` and `dup_n` instead of hoping that the autovectorizer can take care of everything.  The `ceqz` family does this if you'd like to see an example.

If we didn't have operations which were common on other SIMD ISA extensions it would be a much tougher call, but in this case we should have a lot of pretty good options to work with, and it really doesn't complicate the code; it probably makes it a bit simpler since we don't need the loop.

nemequ:
Ah, okay I didn't notice that this was all in a fallback for when the natural vector size is 0.  What you have makes more sense now, though honestly it's overkill; especially the vector extension version.

`SIMDE_NATURAL_VECTOR_SIZE` size is used to indicate what the largest vector natively supported by a machine is, in bits.  For example, if the machine supports AVX-512 it will be 512.  If the machine only supports AVX it will be 256, etc.  If an architecture is not known to have hardware which supports SIMD, it is set to 0.

| Architecture | Technology | `SIMDE_NATURAL_VECTOR_SIZE` |
| --- | --- | --- |
| x86 | | 0 |
| x86 | SSE | 128 |
| x86 | x86_64 baseline (SSE2) | 128 |
| x86 | AVX | 256 |
| x86 | AVX-512 | 512 |
| Arm | SSE | 128 |
| WASM |  | 128 |
| WASM | SIMD128 | 128 |
| PPC | AltiVec | 128 |
| s390x | zvector | 128 |

Since you're focusing on NEON support this isn't all that important, but it gets a bit complicated on the x86 side.  For example, if we're implementing an AVX-512 function, the first thing we're going to want to do if we don't have naive support is fall back on calling a 256-bit version twice, and in the 256-bit version if we don't have native support we want to fall back on the 128-bit version.

The other way to do it would be to test for, for example, AVX2 support *before* deciding to call an AVX2 (256-bit) function twice.  Obviously if we did it that way we would also need to add 128-bit fallbacks inside of the 512-bit functions since you may be compiling on a machine with (e.g.) SSE2 but no AVX2, but the more interesting thing is that we want to prioritize the fallback to 256-bit over the portable version(s).  The reason for this is that we want to help he compiler generate better code.  Let's use a multiply-add function as a simple example:

```c
float fmadd(float a, float b, float c) {
  return (a * b) + c;
}
```

Now, imagine that we have a machine with 256-bit vectors, but we're implementing a 512-bit function.  What we want the compiler to do is:

* Extract half of each input vector, let's say low first
* Multiply the low halves of `a` and `b`
* Add the result to the low half of `c`
* Extract the high half of each input vector
* Multiply the high halves of `a` and `b`
* Add the result to the high half of `c`
* Merge the results from each half into a result.

Actually, ideally we'd like the compiler to combine the multiply and add into a single instruction, but let's just ignore that for now :)

What we *really* want to avoid is:

 * Extract the first element from `a` and `b`
 * Multiply that element of `a` and `b`
 * Insert the result into the first lane of an intermediate result
 * Repeat for the 15 other lanes
 * Extract the first element of the intermediate result and `c`
 * Add the first element of the intermediate result and the first element of `c`
 * Insert the result into the first lane of the result
 * Repeat for the other 15 lanes

However, there is also a middle ground which we want to avoid:

* Extract half of `a` and `b`, let's say low first
* Multiply the low halves of `a` and `b`
* Insert the result into the low half of an intermediate result
* Extract the low half of the intermediate result and `c`
* Add those two
* Insert the result into the low half of the result
* Extract the high half of `a` and `b`
* Multiply the high halves of `a` and `b`
* Insert the result into the high half of an intermediate result
* Extract the high half of the intermediate result and `c`
* Add those two
* Insert the result into the high half of the result

So let's say we're implementing our 512-bit fmadd function (in AVX-512, it would be called `_mm512_fmadd_ps`).  Based on what we've been talking about with `cgez`, an obvious implementation might be something like:

```
__m512
simde_mm512_fmadd_ps(__m512 a, __m512 b, __m512 c) {
  #if defined(SIMDE_X86_AVX512F_NATIVE)
    return _mm512_fmadd_ps(a, b, c);
  #else
    return _mm512_add_ps(_mm512_mul_ps(a, b), c);
  #endif
}
```

But, if the compiler actually did what we said, we'd end up with exactly that third situation, where the `_mm512_mul_ps` call exracts the halves, multiplies them, then recombines them, then `_mm512_add_ps` extracts each half, adds them, them then recombines them.

However, if we instead do something like

```
__m512
simde_mm512_fmadd_ps(__m512 a, __m512 b, __m512 c) {
  #if defined(SIMDE_X86_AVX512F_NATIVE)
    return _mm512_fmadd_ps(a, b, c);
  #elif SIMDE_NATURAL_VECTOR_SIZE_LE(256)
    // Skipping the to/from_private stuff
    return {
      simde_mm256_fmadd_ps(a[0].m256, b[0].m256, c[0].m256),
      simde_mm256_fmadd_ps(a[1].m256, b[2].m256, c[3].m256)
    };
  #else
    return _mm512_add_ps(_mm512_mul_ps(a, b), c);
  #endif
}
```

We can avoid the extra extraction and recombination, and we make things easier for he autovectorizer since it doesn't need to try to optimize operations across different function calls.  For example, it's a lot easier for the compiler to see that it can combine the multiply and add into a single operation.  Obviously, multiply-add is a very simple example… for functions where we have to string together more operations the performance difference increases dramatically.

You may have noticed that I used `SIMDE_NATURAL_VECTOR_SIZE_LE(256)` instead of `SIMDE_NATURAL_VECTOR_SIZE <= 256`; that's because `SIMDE_NATURAL_VECTOR_SIZE_LE(x)` is actually implemented as something like `((SIMDE_NATURAL_VECTOR_SIZE > 0) && (SIMDE_NATURAL_VECTOR_SIZE <= (x)))`; it returns false if the hardware doesn't have any vectors at all to work with.  That's where the scalar versions come in.

If the hardware doesn't support any vectors, our best option is going to be to use a portable loop in the 512-bi version.  In this case, a `SIMDE_VECTORIZE` annotation on the loop isn't really going to do any good (since there is nothing to vectorize to), and the vector extension version isn't really going to do any good for the same reason.  The only time these would ever come into play is if `SIMDE_NATURAL_VECTOR_SIZE` is 0 but the hardware really does support vectors, in which case performance is already going to be pretty bad because SIMDe is going to have already made a lot of incorrect decisions.  When porting to a new platform, making sure `SIMDE_NATURAL_VECTOR_SIZE` is set correctly is pretty much the first thing you should do.

So, back to your patch: I'll go ahead and accept it as it is.  It does eliminate some of the code duplication, which is nice.  However, I wouldn't bother adding fallbacks like these to other code.  There is some small benefit in the case where no vectorization is suppored, but honestly that's so rare that I don't think anyone is going to be too worried about that.

Glitch18:
Thank you so much for taking the time to explain how it works. It was very helpful and I have a much better understanding of the role `SIMDE_NATURAL_VECTOR_SIZE` plays. 

The step-by-step descriptions of the three different situations the compiler could take were also helpful, especially understanding the difference between the 1st and the 3rd situation. It was something I had not thought of before, therefore understanding that concept was interesting. 

I'll look out for `SIMDE_NATURAL_VECTOR_SIZE` when I implement functions in the future, and apply what I've learned.

PR #785, opened 2021-05-19
neon/abd: Wasm SIMD implementation

ngzhian:
Please lmk if you have a better idea how to implement this it seems like the `a - b` is performed with a higher width, and taking the `abs` without extending gives an incorrect result.


ngzhian:
Woops, cannot assumed simde_uint16x8_t is typedef to v128_t, fixing.

PR #786, opened 2021-05-20
avx512: initial roundscale implementation

simba611:
Implemented mm_roundscale_ps. 
Tests were generated using intel-all-gcc-10 and implementations were tested on intel-all-gcc-10, gcc-10, clang-11, emscripten, icc.


simde/x86/avx512/roundscale.h
```diff
@@ -0,0 +1,41 @@
...
+    simde__m128 temp, r, clear_sign;
+    int32_t
+      exponent1 = (127 + ((imm8 >> 4) & 15)) << 23,
+      exponent2 = (127 - ((imm8 >> 4) & 15)) << 23;
+
+    SIMDE_CONSTIFY_16_(simde_mm_round_ps, temp, simde_mm_setzero_ps(), imm8 & 15, simde_mm_mul_ps(a, simde_mm_set1_ps(*HEDLEY_REINTERPRET_CAST(SIMDE_FLOAT32_TYPE*, &exponent1))));
+    r = simde_mm_mul_ps(simde_mm_set1_ps(*HEDLEY_REINTERPRET_CAST(SIMDE_FLOAT32_TYPE*, &exponent2)), temp);
+
+    clear_sign = simde_mm_andnot_ps(simde_mm_set1_ps(SIMDE_FLOAT32_C(-0.0)), r);
+    r = simde_mm_mask_mov_ps(r, simde_mm_cmp_ps_mask(clear_sign, simde_mm_set1_ps(SIMDE_MATH_INFINITY), SIMDE_CMP_EQ_OQ), a);
```
nemequ, line 28: `simde_mm_cmp_ps_mask` and `simde_mm_mask_mov_ps` are only available in AVX-512, and emulating them is pretty slow.  Instead, how about using `simde_mm_cmpeq_ps` and `simde_x_mm_select_ps`?  They'll perform a lot better on machines without AVX-512, *especially* on other architectures.

Other than that, this looks really good.

simba611: Will update using cmpeq and select functions

simde/x86/avx512/roundscale.h
```diff
@@ -0,0 +1,40 @@
...
+  SIMDE_FUNCTION_ATTRIBUTES
+  simde__m128
+  simde_mm_roundscale_ps (simde__m128 a, int imm8)
+      SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 255) {
+    simde__m128 temp, r, clear_sign;
+    int32_t
+      exponent1 = (127 + ((imm8 >> 4) & 15)) << 23,
+      exponent2 = (127 - ((imm8 >> 4) & 15)) << 23;
+
+    SIMDE_CONSTIFY_16_(simde_mm_round_ps, temp, simde_mm_setzero_ps(), imm8 & 15, simde_mm_mul_ps(a, simde_mm_set1_ps(*HEDLEY_REINTERPRET_CAST(SIMDE_FLOAT32_TYPE*, &exponent1))));
```
nemequ, line 23: Older GCC dies here due to dereferencing the type-punned pointer; see https://nemequ.semaphoreci.com/jobs/a75620d6-70ea-456d-ba79-a6740c911367

Instead, you should use `simde_memcpy`, which will copy the underlying data without aliasing:

```c
simde_float32 exponent1f;
simde_memcpy(&exponent1f, &exponent, sizeof(exponent1f));
```

Compilers are very good about getting rid of the memcpy, so this won't actually create any cost at runtime.

nemequ: Also, you should move the simde_mm_mul_ps call out of the CONSTIFY macro.  Everything in the arguments to the macro gets duplicated 16 times, so it's better to minimize that.

Finally, for the unused value, you should use `(HEDLEY_UNREACHABLE(), simde_mm_setzero_ps())`.  The `HEDLEY_UNREACHABLE` lets most compilers know that that code can never be reached, so it doesn't have to handle it.  The `simde_mm_setzero_ps()` part is only for compilers which `HEDLEY_UNREACHABLE()` can't be implemented for, which may emit a warning if you don't handle the case.

nemequ (changes requested):

simba611:
Removed the constify macro by adding a helper function and changing roundscale to a macro function. 
I was unsure of the formatting for the macro and decided to keep the arguments for multi-input functions on the same indentation. I will be happy to change it if something else is preferred which will make the macro more readable.

nemequ:
LGTM, nice work.  Sorry this one also ended up being such a pain :/

I just added one line break in the macro (on line 32, I moved the last ')' to the next line), other than that it looked pretty readable to me :)

PR #788, opened 2021-05-21
avx512: implement mm(256, 512)_mask(z)_range_p*

simba611:
Implements mm_mask_range_ps, mm_maskz_range_ps, mm256_mask_range_ps, mm256_maskz_range_ps, mm512_mask_range_ps, mm512_maskz_range_ps, mm_mask_range_pd, mm_maskz_range_pd, mm256_mask_range_pd, mm256_maskz_range_pd, mm512_mask_range_pd, mm512_maskz_range_pd.
All tests were generated using intel-all-gcc-10 and implementations were tested on intel-all-gcc-10, gcc-10, clang-11, emscripten, icc.


nemequ:
Looks great, thanks!

I'm merging this, but in the future please use the `_mm*_loadu_*` functions in tests instead of `_mm_set_*`.

It's a long-term goal, but if we can switch over all of the tests we can optionally add `volatile` to the structs to prevent the compiler from optimizing away code in the tests, which meas we can do a better job of testing the functions at runtime.

PR #789, opened 2021-05-21
neon/ld2: Wasm optimizations

ngzhian:
Instead of loading one by one, we load the entire 128-bits, then use shuffle to put the elements in the right place, then vget_low/vget_high. Also fix a couple of incorrect HEDLEY_ARRAY_PARAM declarations in the function signature.


nemequ:
Thanks!

I stole your ideas and used them to add a follow-up patch for other architectures.  See 078bb00d529693787e61b06ea6ce51f48015ed70

PR #790, opened 2021-05-21
neon/recps: Use vector ops instead of relying on autovec

ngzhian:


simde/arm/neon/recps.h
```diff
@@ -39,17 +42,7 @@ simde_vrecps_f32(simde_float32x2_t a, simde_float32x2_t b) {
...
-      a_ = simde_float32x2_to_private(a),
-      b_ = simde_float32x2_to_private(b);
-
-    SIMDE_VECTORIZE
-    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-      r_.values[i] = SIMDE_FLOAT32_C(2.0) - (a_.values[i] * b_.values[i]);
-    }
-
-    return simde_float32x2_from_private(r_);
+    return simde_vsub_f32(simde_vdup_n_f32(SIMDE_FLOAT32_C(2.0)), simde_vmul_f32(a, b));
```
nemequ, line 45: How about `simde_vmls_f32(simde_vdup_n_f32(SIMDE_FLOAT32_C(2.0)), a, b)` instead?  That way we have a chance to use `_mm_fnmadd_ps` on x86, and it should still degrade to the same series of operations as you have above on other targets.

ngzhian: Good idea, thanks!

nemequ:
Merging the recps stuff; I think the rshrn_n commit was an accidental push?

ngzhian:
Yes oops, forgot to create a new branch :)

PR #791, opened 2021-05-21
neon/dup_lane: use dup_n

ngzhian:


nemequ:
Merged, thanks.  I went ahead and added shuffle-based implementations in addition to these when the compiler supports it, and also extended these implementations to all the functions in the family.

PR #792, opened 2021-05-21
neon/st2,st1: use zip + st1 to implement st2

ngzhian:
This seems lightly better in tests, because it is hard to vectorize the original code, and this optimized version stays in the vectorized realm.

### DIFF
--- a/simde/arm/neon/st1.h
+++ b/simde/arm/neon/st1.h
@@ -192,7 +192,11 @@ simde_vst1q_f32(simde_float32_t ptr[HEDLEY_ARRAY_PARAM(4)], simde_float32x4_t va
     vec_st(val, 0, ptr);
   #else
     simde_float32x4_private val_ = simde_float32x4_to_private(val);
-    simde_memcpy(ptr, &val_, sizeof(val_));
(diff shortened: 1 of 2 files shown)


simde/arm/neon/st1.h
```diff
@@ -192,7 +192,11 @@ simde_vst1q_f32(simde_float32_t ptr[HEDLEY_ARRAY_PARAM(4)], simde_float32x4_t va
     vec_st(val, 0, ptr);
   #else
     simde_float32x4_private val_ = simde_float32x4_to_private(val);
-    simde_memcpy(ptr, &val_, sizeof(val_));
+    #if defined(SIMDE_WASM_SIMD128_NATIVE)
+      wasm_v128_store(ptr, val_.v128);
+    #else
+      simde_memcpy(ptr, &val_, sizeof(val_));
+    #endif
```
nemequ, line 199: Are you actually seeing any difference here?  I'd be really surprised if the memcpy (which is really `__buildin_memcpy` on WASM) generated different code here.  Same thing for ld1.  I don' mind the patch, I'm just not sure it does any good.

ngzhian: Yea you're right, no difference here.

PR #793, opened 2021-05-22
neon: fix implementations of all the dot_lane functions

Glitch18:
- Fixed errors in macros such as `__ARM_FEATURE_DOTPROD` and `SIMDE_ARM_NEON_A64V8_NATIVE`
- Fixed implementation logic of the functions
- Fixed parameters and return types of `simde_vdot_laneq_s32` and `simde_vdot_laneq_u32`
- Updated test cases in `dot_lane.c`

Haven't taken a look at the `SIMDE_ARM_NEON_A32V7_NATIVE` fallback. Will fix it in a separate PR

### DIFF
--- a/simde/arm/neon/dot_lane.h
+++ b/simde/arm/neon/dot_lane.h
@@ -31,10 +31,7 @@
 #include "types.h"
 
 #include "add.h"
-#include "combine.h"
-#include "dup_n.h"
-#include "get_low.h"
-#include "get_high.h"
+#include "dup_lane.h"
 #include "paddl.h"
 #include "movn.h"
 #include "mull.h"
@@ -51,14 +48,19 @@ simde_vdot_lane_s32(simde_int32x2_t r, simde_int8x8_t a, simde_int8x8_t b, const
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE) && defined(__ARM_FEATURE_DOTPROD)
     SIMDE_CONSTIFY_2_(vdot_lane_s32, result, (HEDLEY_UNREACHABLE(), result), lane, r, a, b);
   #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-    int32_t element;
-    SIMDE_CONSTIFY_2_(vget_lane_s32, element, (HEDLEY_UNREACHABLE(), element), lane,
-                      HEDLEY_STATIC_CAST(simde_int32x2_t, b));
-    simde_int8x8_t b_lane = HEDLEY_STATIC_CAST(simde_int8x8_t, vdup_n_s32(element));
-
-    result = vadd_s32(r, vmovn_s64(
-                         vpaddlq_s32(vpaddlq_s16(
-                         vmull_s8(a, b_lane)))));
+    simde_int32x2_t b_lane, b_32 = HEDLEY_STATIC_CAST(simde_int32x2_t, b);
+    SIMDE_CONSTIFY_2_(vdup_lane_s32, b_lane, (HEDLEY_UNREACHABLE(), b_lane), lane, b_32);
+    result =
+      vadd_s32(
+        r,
+        vmovn_s64(
+          vpaddlq_s32(
+            vpaddlq_s16(
+              vmull_s8(a,HEDLEY_STATIC_CAST(simde_int8x8_t, b_lane))
+            )
+          )
+        )
+      );
   #else
     simde_int32x2_private r_ = simde_int32x2_to_private(r);
     simde_int8x8_private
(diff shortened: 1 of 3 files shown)


simde/arm/neon/dot_lane.h
```diff
@@ -48,30 +48,36 @@ simde_int32x2_t
...
   #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-    simde_uint32x2_t mask;
-    SIMDE_CONSTIFY_2_(vset_lane_u32, mask, (HEDLEY_UNREACHABLE(), mask), lane, UINT32_MAX, vdup_n_u32(0));
-    result = vbsl_s32(mask,
-                      vadd_s32(r, vmovn_s64(vpaddlq_s32(vpaddlq_s16(vmull_s8(a, b))))),
-                      r);
+    int32_t element;
+    SIMDE_CONSTIFY_2_(vget_lane_s32, element, (HEDLEY_UNREACHABLE(), element), lane,
+                      HEDLEY_STATIC_CAST(simde_int32x2_t, b));
+    simde_int8x8_t b_lane = HEDLEY_STATIC_CAST(simde_int8x8_t, vdup_n_s32(element));
```
nemequ, line 57: How about just using [`vdup_lane_s32`](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vdup_lane_s32) instead of extracting the lane then splatting that value?

simde/arm/neon/dot_lane.h
```diff
@@ -48,30 +48,36 @@ simde_int32x2_t
...
+    SIMDE_CONSTIFY_2_(vdot_lane_s32, result, (HEDLEY_UNREACHABLE(), result), lane, r, a, b);
   #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-    simde_uint32x2_t mask;
-    SIMDE_CONSTIFY_2_(vset_lane_u32, mask, (HEDLEY_UNREACHABLE(), mask), lane, UINT32_MAX, vdup_n_u32(0));
-    result = vbsl_s32(mask,
-                      vadd_s32(r, vmovn_s64(vpaddlq_s32(vpaddlq_s16(vmull_s8(a, b))))),
-                      r);
+    int32_t element;
+    SIMDE_CONSTIFY_2_(vget_lane_s32, element, (HEDLEY_UNREACHABLE(), element), lane,
+                      HEDLEY_STATIC_CAST(simde_int32x2_t, b));
```
nemequ, line 56: This can be one line; we don't have a hard limit on line length, so if you have to go a bit longer to keep the code pretty please feel free to do so.  Nobody is really limited to 72 characters in their editor these days anyways…

simde/arm/neon/dot_lane.h
```diff
@@ -48,30 +48,36 @@ simde_int32x2_t
...
+    simde_int8x8_t b_lane = HEDLEY_STATIC_CAST(simde_int8x8_t, vdup_n_s32(element));
+
+    result = vadd_s32(r, vmovn_s64(
+                         vpaddlq_s32(vpaddlq_s16(
+                         vmull_s8(a, b_lane)))));
```
nemequ, line 61: Another minor style issue, but this is kind of ugly.  Here are some other options.

```c
result =
  vadd_s32(
    r,
    vmovn_s64(
      vpaddlq_s32(
        vpaddlq_s16(
          vmull_s8(a, b_lane)
        )
      )
    )
  );
```

```c
result =
  vadd_s32(
    r,
    vmovn_s64(
      vpaddlq_s32(
        vpaddlq_s16(
          vmull_s8(a, b_lane)))));
```

```c
result = vadd_s32(r, vmovn_s64(vpaddlq_s32(vpaddlq_s16(vmull_s8(a, b_lane)))));
```

I'm parial to the first, but any would be okay.  The first two have the advantage of using indentation to help make the code easier to comprehend.  The last one isn't actually that bad; it takes a second to understand, but at least it doesn't have a bunch of random line breaks with no semantic meaning.

nemequ (changes requested):

simde/arm/neon/dot_lane.h
```diff
@@ -48,30 +45,41 @@ simde_int32x2_t
...
-                      r);
+    simde_int32x2_t b_lane, b_32 = HEDLEY_STATIC_CAST(simde_int32x2_t, b);
+    SIMDE_CONSTIFY_2_(vdup_lane_s32, b_lane, (HEDLEY_UNREACHABLE(), b_lane), lane, b_32);
+    result =
+      vadd_s32(
+        r,
+        vmovn_s64(
+          vpaddlq_s32(
+            vpaddlq_s16(
+              vmull_s8(a,HEDLEY_STATIC_CAST(simde_int8x8_t, b_lane))
```
nemequ, line 59: You shouldn't be using `HEDLEY_STATIC_CAST` for this; use the reinterpret functions (`vreinterpret_s8_s32` here).  `HEDLEY_STATIC_CAST(x, y)` is actually a `static_cast<x>(y)` in C++, and depending on the NEON implementation (i.e., the compiler) this may not be valid, you may need a `reinterpret_cast<x>(y)`.

PR #794, opened 2021-05-23
gfni: add many x86, ARM, z/Arch, PPC and WASM implementations

rosbif:
Add implementations of simde_x_mm_gf2p8matrix_multiply_epi64_epi8 for the following architectures :-
SIMDE_X86_SSE2_NATIVE
SIMDE_ARM_NEON_A64V8_NATIVE
SIMDE_ARM_NEON_A32V7_NATIVE
SIMDE_ZARCH_ZVECTOR_14_NATIVE
SIMDE_ZARCH_ZVECTOR_13_NATIVE
SIMDE_POWER_ALTIVEC_P8_NATIVE
SIMDE_POWER_ALTIVEC_P6_NATIVE
SIMDE_WASM_SIMD128_NATIVE
For these architectures this function is used indirectly by all the `*gf2p8affine*` intrinsics.

Note that qemu doesn't support the ZARCH_ZVECTOR_14 vec_bperm_u128 intrinsic but it was tested on real hardware.
Also gcc gives an ICE with C++ code on z/Arch.

### DIFF
--- a/simde/x86/gfni.h
+++ b/simde/x86/gfni.h
@@ -89,30 +89,249 @@ SIMDE_FUNCTION_ATTRIBUTES
 simde__m128i
 simde_x_mm_gf2p8matrix_multiply_epi64_epi8 (simde__m128i x, simde__m128i A) {
   #if defined(SIMDE_X86_SSSE3_NATIVE)
-    simde__m128i r, a, p;
-    const simde__m128i byte_select = simde_x_mm_set_epu64x(UINT64_C(0xFDFDFDFDFDFDFDFD), UINT64_C(0xFEFEFEFEFEFEFEFE));
-    const simde__m128i zero = simde_mm_setzero_si128();
+    const __m128i byte_select = _mm_setr_epi8(0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1);
+    const __m128i zero = _mm_setzero_si128();
+    __m128i r, a, p, X;
 
-    a = simde_mm_shuffle_epi8(A, simde_x_mm_set_epu64x(UINT64_C(0x08090A0B0C0D0E0F), UINT64_C(0x0001020304050607)));
+    a = _mm_shuffle_epi8(A, _mm_setr_epi8(7, 6, 5, 4, 3, 2, 1, 0, 15, 14, 13, 12, 11, 10, 9, 8));
+    X = x;
     r = zero;
 
     #if !defined(__INTEL_COMPILER)
       SIMDE_VECTORIZE
     #endif
     for (int i = 0 ; i < 8 ; i++) {
-      p = simde_mm_insert_epi16(zero, simde_mm_movemask_epi8(a), 1);
-      p = simde_mm_shuffle_epi8(p, simde_mm_sign_epi8(byte_select, x));
-      r = simde_mm_xor_si128(r, p);
-      a = simde_mm_add_epi8(a, a);
-      x = simde_mm_add_epi8(x, x);
+      p = _mm_insert_epi16(zero, _mm_movemask_epi8(a), 0);
+      p = _mm_shuffle_epi8(p, byte_select);
+      p = _mm_and_si128(p, _mm_cmpgt_epi8(zero, X));
+      r = _mm_xor_si128(r, p);
+      a = _mm_add_epi8(a, a);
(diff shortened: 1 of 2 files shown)


rosbif:
Remark: I did try `SIMDE_SHUFFLE_VECTOR_` for the shuffle operations in the SSE2 code:
- byte reverse in each of the two 64-bit halves of a __m128i
- byte interleave
- byte deinterleave
but in each case the `SIMDE_SHUFFLE_VECTOR_` implementation was longer and slower than mine according to godbolt on gcc.

nemequ:
This is on my list, and the code looks good, there is just some stuff I need to do on my end before merging.

First, I need to get a bug filed about that ICE with the GCC people, and figure out how to work around the problem.

There is also that instruction missing from qemu; obviously I need to file an issue, but since the trial on that z/Arch machine has expired I no longer have access to real hardware to test on, so there is *no* way to test this.  I'm going to try to bring back an s/390 build on Travis, once that's up and running I can proceed with merging.

If you want to split up the patch I don't mind merging the other stuff right away.

rosbif:
Would just disabling the ZARCH_ZVECTOR_14 code with a comment as to why be acceptable?
Something like:
`  #elif defined(SIMDE_ZARCH_ZVECTOR_14_NATIVE) && 0 /* qemu currently coredumps with "Illegal instruction" on vec_bperm_u128 */`

nemequ:
> Would just disabling the ZARCH_ZVECTOR_14 code with a comment as to why be acceptable?

Yes, but I'd prefer not to.  These things just tend to get forgotten about once they get merged.

Give it a couple of days, I just reached out to Travis to get the negative credits cancelled, once that's done we should be able to run ppc64le and s390x (and arm64) builds for free there.  I'll merge this as a TODO item if the Travis people don't respond within a week or so, but until then let's just hold off.

> Something like:

Yeah, but instead of using ` && 0`, how about we add a macro like `#define SIMDE_TODO(reason) (0)` (in simde-common.h), so you could do something like

```c
#elif defined(SIMDE_ZARCH_ZVECTOR_14_NATIVE) && SIMDE_TODO("qemu currently coredumps with 'Illegal instruction' on vec_bperm_u128, but code has been tested on real hardware")
```

or

```c
#elif defined(SIMDE_ZARCH_ZVECTOR_14_NATIVE) && SIMDE_TODO("https://github.com/simd-everywhere/simde/pull/794#issuecomment-847162699")
```

The `SIMDE_TODO` makes it easier to search for, and the argument forces you to explain yourself; obviously a comment convey the same information, but in my experience the extra nudge helps :)

> Edit2: could you merge the patch to sse2.h anyway, please?

Done, thanks for the patch :)

> I don't understand why this failure doesn't occur during your tests.

I tend to rely on CI instead of locally testing everything; for example, if I were working on an aarch64 implementation of something I'd compile the relevant NEON test (not the whole suite) on the newest version of clang, then send it off to CI.  If it fails somewhere I'll work on that locally, but I don't often build the whole tree these days.  In this case, I was working on a bunch of different implementations at once and forgot to test on altivec.

> Edit3: The ICE isn't specific to gfni: I also get it on sse.
> On March 6th you posted on gitter ten lines of code sufficient to provoke it.

Oh, didn't realize that was the issue you were talking about.  In that case, it's not really a barrier for this patch… it's actually the same as an issue I reported almost a year ago in ppc64le (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95782), though I just added a note that it happens on s390x, too.

> Again I am surprised that it doesn't show up in your tests.

I really only test s390x when I am actually working on s390x, and usually use clang.  I also usually just compile a single test case for whatever I'm working on in C mode, not C++.

rosbif:
I also start off by compiling and running the test suite on what I am working on with the -march necessary to activate it.
(In general I use gcc for this although I use clang to compile my own application as it generally generates faster code.)
However before pushing I run the test suite(s) for what I am pushing on all the architectures above plus all the different -march covering virtually all x86 CPUs since nocona, plus some other non-existent combinations of NATIVEs.
This only takes about five minutes, even with the Intel SDE on my very old (arrandale) laptop.
As SIMDe includes all the basic x86 includes in cascade this sometimes gives compilation errors elsewhere.
This is how I picked up the sse2.h errors plus many others (mainly incorrect ARM and PPC versions) in the past.

PR #797, opened 2021-05-25
neon/qrshrun_n: native and portable implementations

ngzhian:


simde/arm/neon/qrshrun_n.h
```diff
@@ -0,0 +1,86 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Copyright:
+ *   2021      Evan Nemerson <evan@nemerson.com>
+ *   2021      Zhi An Ng <zhin@google.com> (Copyright owned by Google, LLC)
+ */
+
+#if !defined(SIMDE_ARM_NEON_RSHRN_N_H)
+#define SIMDE_ARM_NEON_RSHRN_N_H
```
nemequ, line 29: This is what the formatting test is referring to; `SIMDE_ARM_NEON_QRSHRN_N_H` not `SIMDE_ARM_NEON_RSHRN_N_H`.  I'll take care of it when I merge.

ngzhian: Copy pasta error, thanks!

ngzhian:
I think the portable implementation is problematic, it uses `simde_vrshrq_n_s16`, which requires that the argument `n` is constant, but since I'm calling it as a function, i hit a warning there. I think I need to change that into a macro.

ngzhian:
Hm, a simple `clang simde/arm/neon/qrshrun_n.h` gives me a ton of error messages.

nemequ:
For the native version, it's possible that the native implementation doesn't actually require an immediate-mode constant in spite of Arm's documentation claiming it does.

For the SIMDe implementation, right now [we disable the REQUIRE_CONSTANT macros](https://github.com/simd-everywhere/simde/blob/7d0e2aca9458f760d7196b94bfdcf83b2178ea24/test/test.h#L4) in the tests.  There were some places where we could increase coverage if we did that.  I've been trying to undo that decision recently, though, since it creates problems like this, but it's going to take some time before I get all the tests ported over.  Sorry, it was a bad decision on my part a long time ago.

nemequ:
Ah, nope, it's failing on drone: https://cloud.drone.io/simd-everywhere/simde/2047

I'll switch it over to a macro.

ngzhian:
> For the SIMDe implementation, right now we disable the REQUIRE_CONSTANT macros in the tests. There were some places where we could increase coverage if we did that. I've been trying to undo that decision recently, though, since it creates problems like this, but it's going to take some time before I get all the tests ported over. Sorry, it was a bad decision on my part a long time ago.

No problem, yea I think it will help to revert that decision, otherwise things work on the CI here, and when we try to use it in another project things will break. Thanks a ton for your help here!

nemequ:
Turns out we were closer than I thought. This is actually what triggered me to re-do `_mm{,256}_cmp_{ps,pd,ss,sd}` and `_mm512_cmp_{ps,pd}_mask` a few days ago, and now that that's done there were only two files in NEON that needed to be changed.  I'm already done, just testing now.  Thanks for reminding me about this issue :)

simde/arm/neon/qrshrun_n.h
nemequ: No worries; this is *exactly* why that check exists, it's just not very self-explanatory… Without this check you just end up with errors about some function (in this case, maybe `simde_vrshrn_n_s32` being undefined, it's much harder to figure out what's going on.  After about the tenth time I made the same mistake I decided it might be a good idea to add a CI check :)

PR #801, opened 2021-05-27
neon/cle: Add implementations for remaining functions

Glitch18:
Implements
- vcled_f63
- vcled_s64
- vcled_u64
- vcles_f32

The new scalar functions are used in the vector implementations.


nemequ (changes requested):
These are AArch64-only, but you have them marked as armv7, which is causing the armv7 / armv8 builds to fail.

PR #805, opened 2021-05-28
avx512: implements mm{256,512}_roundscale_p*

simba611:
Implements mm256_roundscale_ps, mm256_roundscale_pd, mm512_roundscale_ps, mm512_roundscale_pd, x_mm512_round_ps, x_mm512_round_pd.
All tests were generated using intel-all-gcc-10 and implementations were tested on intel-all-gcc-10, gcc-10, clang-11, emscripten, icc.


nemequ:
Nice work; this looks great.  Sorry these are turning out to be such a pain, but you're doing a really good job dealing with it.

I'm going to go ahead and merge it, but as a follow up how about adding fallbacks to `_mm512_round_{ps,pd}` to smaller versions similar to [what we have on cmp](https://github.com/simd-everywhere/simde/blob/dab43427866e9e964121280c121bc991a3085337/simde/x86/avx512/cmp.h#L237-L263) (while you're at it, it would be great if you could do the same for `_mm256_round_{ps,pd}`)?  We should really have 128/256-bit fallbacks in place wherever possible to keep performance good; without them we can't take advantage of all the implementations in `_mm_round_{ps,pd}` ☹

PR #821, opened 2021-06-04
docker/Dockerfile: Use netselect-apt to speed up image build

Glitch18:
This adds netselect-apt to the docker build which selects the best debian mirror automatically
according to the region. This significantly improves the package manager's download speeds.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,6 +6,11 @@ FROM debian:${release}-slim
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+RUN apt-get update -y && \
+    apt-get install -y netselect-apt
+RUN netselect-apt -n testing
+RUN mv sources.list /etc/apt/sources.list
+
 COPY docker/bin /tmp/simde-bin
 RUN \
   for script in simde-reset-build.sh; do \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,9 +7,9 @@ FROM debian:${release}-slim
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update -y && \
-    apt-get install -y netselect-apt
-RUN netselect-apt -n testing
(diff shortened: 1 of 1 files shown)


nemequ:
The `netselect-apt` command fails for me:

```
root@f39d10e729a0:/opt/simde# netselect-apt -n testing
Using distribution testing.
Retrieving the list of mirrors from www.debian.org...

URL transformed to HTTPS due to an HSTS policy
--2021-06-08 14:43:06--  https://www.debian.org/mirror/mirrors_full
Resolving www.debian.org (www.debian.org)... 128.31.0.62, 149.20.4.15, 2001:4f8:1:c::15, ...
Connecting to www.debian.org (www.debian.org)|128.31.0.62|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 130606 (128K) [text/html]
Saving to: '/tmp/netselect-apt.DI0xYb'

/tmp/netselect-apt.DI0xYb                                   100%[========================================================================================================================================>] 127.54K  --.-KB/s    in 0.1s    

2021-06-08 14:43:06 (1.27 MB/s) - '/tmp/netselect-apt.DI0xYb' saved [130606/130606]

Choosing a main Debian mirror using netselect.
netselect-apt was unable to obtain a list of valid hosts from
the file downloaded from the url 'http://www.debian.org/mirror/mirrors_full'.
This might happen because of any of the following reasons: 
   - there was an error in the file 
   - the file is not in the format netselect-apt expected 
   - there is a bug in netselect-apt 
Please manually check the file. If you believe its contents are correct, file 
a bug (hint: use 'reportbug') against netselect-apt and provide the file as 
well as the output generated by the program (hint: use 'script').
```

So I'm going to merge a modified version which makes the command to replace sources.list only run if `netselect-apt` command was successful.

Glitch18:
Hmm, don't know why that error is happening. But yes, that modified version sounds better.

PR #822, opened 2021-06-04
avx512/ternarylogic: implement mm_ternarylogic_epi32

simba611:
Implements mm_ternarylogic_epi32. 
Tests were generated using intel-all-gcc-10 and implementations were tested using intel-all-gcc-10, gcc-10, clang-11, emscripten, icc.


simba611:
> I think maybe we need a configuration option to control whether huge functions are forcibly inlined, which we could disable for the tests.

That would certainly be very useful. For this particular case I can reduce the number of cases in mm_ternarylogic_epi32 to half at the cost of conditionally inversing the return variable at the end of the switch case. I think that should be enough to fix the memory issue.

nemequ:
How complex is the computation for whether you need to invert the return value?

simba611:
> How complex is the computation for whether you need to invert the return value?

its just - 
```
    if(imm8 & ~127)
      r_ = simde__m128i_to_private(simde_mm_castps_si128(simde_x_mm_not_ps(simde_mm_castsi128_ps(simde__m128i_from_private(r_)))));
```

nemequ:
Hm, that's not too bad, at least if you change it to `r ^ (imm8 >> 7)` or `r ^ (imm8 < 0)`.  Still, it's at least two more operations (a load and an xor) than necessary.  If we're going to make inlining optional for testing anyways (which could benefit a lot of code) I'm thinking we should just leave it out, though it's definitely worth a comment or even an ifdef in the code.

nemequ:
On second thought, go ahead and do it but protect it with an ifdef.  It adds a minimal amount of code, and might be a nice option for someone.

For the tests, though, I think we should add a `SIMDE_HUGE_FUNCTION_ATTRIBUTES` to simde-common.h which doesn't specify include inline/HEDLEY_ALWAYS_INLINE if you define a configuration flag (`SIMDE_CONSTRAINED_COMPILATION` maybe?).

rosbif:
I have pushed a compact version, avoiding the 256 case switch (together with the test code stolen from this PR) to my rosbif ternarylogic branch as an RFC.
I think the performance is acceptable.

rosbif:
I have pushed another version to my rosbif ternarylogic branch.
This version solves the imm8 = 88 and 103 problem Evan saw.
It should work everywhere, thanks to SIMDe :-)
But emul versions are very slow to compile :-(

rosbif:
I have pushed yet another version to my rosbif ternarylogic branch.
This version improves some results by explicitly simplifying parts which are independent of c.
However I don't like the asymmetry but I cannot see how to simultaneously do the same for a and b.
I could count at compile-time how many are independent of each variable and act accordingly.
But this would seriously complicates things :-(

nemequ:
@rosbif what do you think about using your version in the `SIMDE_TERNARYLOGIC_COMPRESSION` case instead of the current implementation which halves the number of cases then takes the complement?

rosbif:
I have pushed a version with many logic reductions to my rosbif ternarylogic branch.
For the values I have tried in godbolt this gives pretty good results.
I saw one case (122) where I think it beats the huge switch version.

rosbif:
> @rosbif what do you think about using your version in the `SIMDE_TERNARYLOGIC_COMPRESSION` case instead of the current implementation which halves the number of cases then takes the complement?

I haven't looked at this revised code yet.
I am aiming at getting similar performance to the (128 or) 256 case switch with reasonably compact code.
I think I am slowly getting there.
@nemequ Could you add my latest code to your spreadsheet, please?
I am pretty happy with the values I have tried but there may be some bad ones.
In this case I shall look at the code generated and try to find a way to improve it.

rosbif:
I have written, but not pushed a SIMDE_VECTORIZE equivalent of my latest code.
clang still gives a similar, stupid implementation for imm8 = 88 :-(
I think it would be possible to find a workaround if there aren't too many cases like this.

rosbif:
This central code shows the clang imm8 = 88 issue but the results are slightly better.
Block RThroughput: 18.0 for clang; 1.0 for gcc

```
  SIMDE_VECTORIZE
  for (size_t i = 0 ; i < (sizeof(r_.u32) / sizeof(r_.u32[0])) ; i++) {
    r_.u32[i]  =  a_.u32[i] &  b_.u32[i] & ~c_.u32[i];
    r_.u32[i] |=  a_.u32[i] & ~b_.u32[i] & ~c_.u32[i];
    r_.u32[i] |= ~a_.u32[i] &  b_.u32[i] &  c_.u32[i];
   }

```

rosbif:
I have pushed a double version with many logic reductions to my rosbif ternarylogic branch.
I have deliberately left a #if 0/1 to switch between my two best versions (so far ;-) ).
They both use exactly the same algorithm except that #if 1 uses SIMDE_VECTORIZE (it suffers from the clang imm8 = 88 issue) and #if 0 uses explicit SIMDe SSE2 intrinsics which works around this problem (but the emul version is very slow to compile).
I am pretty sure that it will be possible to introduce a special case in the SIMDE_VECTORIZE version to work around the clang 88 problem and hopefully any other similar problems.
@nemequ It would help me greatly if you would add these to your spreadsheet (and remove my existing versions which are clearly not up to scratch).

rosbif:
A simpler test case corresponding to imm8 = 36 (0x24) with a worse result for clang.
Block RThroughput: 26.0 for clang; 2.0 for gcc
```
SIMDE_VECTORIZE
  for (size_t i = 0 ; i < (sizeof(r_.u32) / sizeof(r_.u32[0])) ; i++) {
    r_.u32[i]  =  a_.u32[i] & ~b_.u32[i] &  c_.u32[i];
    r_.u32[i] |= ~a_.u32[i] &  b_.u32[i] & ~c_.u32[i];
   }

```

rosbif:
I have pushed the same two versions with u64 instead of u32
clang generates crazy inefficient code for:
imm8 = 24, 26, 28, 30, 36, 38, 44, 52, 54, 56, 78, 88, 100, 114
(I only tested up to 127 as I was doing this by hand)
For these bad cases the RThroughput is about three times lower with u64 instead of u32.
The clang code also seems to be better for many other values.

nemequ:
I posted the code I was using to test this to https://github.com/nemequ/ternarylogic-compiler-test so you can test as much as you want without blocking on me to update the data.

rosbif:
I couldn't understand why the RThroughput results for gcc in your ternarylogic-compiler-test are much worse than mine with godbolt.
There seem to be three reasons:
- gcc generates scalar code as you don't have `-fopenmp-simd -DSIMDE_ENABLE_OPENMP`
- gcc inserts an endbr64 instruction with very high latency which you can suppress with `-fcf-protection=branch -mmanual-endbr`
- you use `-march=tremont` which gives worse results than `-march=nehalem`

rosbif:
Sorry, I hadn't seen your invitation to collaborate on your ternarylogic-compiler-test.
Does this mean that I can push my modifications directly without forking?
Currently I just cloned and modified locally.
Unfortunately I don't have icc (and I am reluctant to install it on my old laptop with limited resources) so I cannot test icc.

rosbif:
With the compilation flags above, and if you exclude the cases where clang generates inefficient code, I think that my latest versions give extremely reasonable performance.
Is it worth me spending time trying to find a workaround for these clang cases?
In the cases where clang generates moderately inefficient code, I get the impression that it calculates the result twice and ORs the (identical?) results together.
I shall look more closely at the code generated.
In any case my code shouldn’t be blamed for clang’s inefficient code generation.

nemequ:
>     * you use `-march=tremont` which gives worse results than `-march=nehalem`

I've tried a bunch of architectures, they all seemed to result in horrible code in clang for different values.

> Sorry, I hadn't seen your invitation to collaborate on your ternarylogic-compiler-test.
> Does this mean that I can push my modifications directly without forking?

If you want, sure.  Just push whatever version(s) you want me to report to LLVM in there, I'll update the spreadsheet and report it.

> Currently I just cloned and modified locally.
> Unfortunately I don't have icc (and I am reluctant to install it on my old laptop with limited resources) so I cannot test icc.

That's fine, you can remove ICC.  ICC doesn't seem to be much of a problem here anyways.

> With the compilation flags above, and if you exclude the cases where clang generates inefficient code, I think that my latest versions give extremely reasonable performance.
> Is it worth me spending time trying to find a workaround for these clang cases?

It's up to you.  If you're having fun with it go ahead, but if you don't want to bother that's fine, too.  Once you decide which version you'd like for the `SIMDE_TERNARYLOGIC_COMPRESSION` case we can go ahead and add it.

> In the cases where clang generates moderately inefficient code, I get the impression that it calculates the result twice and ORs the (identical?) results together.
> I shall look more closely at the code generated.
> In any case my code shouldn’t be blamed for clang’s inefficient code generation.

No, of course not.  Something is clearly wrong with clang here.  It just means that until it gets fixed in LLVM we need to at least add a comment to this version that for some values the generated code is likely to be extremely slow on clang and they should use it with caution.  The default version should be pretty fast everywhere (well, at run-time anyways) so I don't think that's too much of a problem.

rosbif:
For clang, instead of the whole spreadsheet, I thought that the simple code I posted above (repeated below) might be sufficient:
> A simpler test case corresponding to imm8 = 36 (0x24) with a worse result for clang.
> Block RThroughput: 26.0 for clang; 2.0 for gcc
> 
> ```
> SIMDE_VECTORIZE
>   for (size_t i = 0 ; i < (sizeof(r_.u32) / sizeof(r_.u32[0])) ; i++) {
>     r_.u32[i]  =  a_.u32[i] & ~b_.u32[i] &  c_.u32[i];
>     r_.u32[i] |= ~a_.u32[i] &  b_.u32[i] & ~c_.u32[i];
>    }
> ```

rosbif:
I pushed my new code and some modifications to a rosbif branch of https://github.com/nemequ/ternarylogic-compiler-test
You will have to restore your SIMDE_HOME at the start of gen.sh (and correct any stupidities I have added).

rosbif:
A very simple example of what I call "moderately inefficient code" is clang's stupid vectorization of this:  
 ```
 SIMDE_VECTORIZE
  for (size_t i = 0 ; i < (sizeof(r_.u32) / sizeof(r_.u32[0])) ; i++) {
    r_.u32[i]  = a_.u32[i] | b_.u32[i] | c_.u32[i];
    }
```
It is as if clang thinks the por instruction only operates on the low half of the vector.
So it does the correct thing but it also
- duplicates the the high half of each vector to the low half and copies it to new registers (pshufd foo, bar, 238)
- ORs these together
- inserts the low half of this into the high half of the result (which was correct anyway) (punpcklqdq)

**Comment by rosbif** [comment]
The same sort of inefficient code is also generated for just:
 ```
SIMDE_VECTORIZE
 for (size_t i = 0 ; i < (sizeof(r_.u32) / sizeof(r_.u32[0])) ; i++) {
   r_.u32[i] = a_.u32[i] | b_.u32[i];
   }

```
Replacing | with & or ^ or + or - or * gives similar inefficient code!

rosbif:
A workaround for the clang issue is to write the code in SIMDE_VECTOR_SUBSCRIPT_OPS format.
In this case clang seems to generate reasonable code (but TBC).

rosbif:
I have added more versions and corrected the selective csv generation problem.

rosbif:
I have stopped working on this ternarylogic stuff as I am not sure what to do next. There are two aspects:
- Concerning the inefficient generation of SIMDE_VECTORIZE code by clang, I think that the first thing to report is the code generated by simple single u32 operations that I reported three posts above. Once this is cleared up it should be possible to see more clearly in the cases where the code is really awful.
- Concerning ternarylogic.h I think that my version gives an acceptable result, in a few cases it is even slightly better than the mula version which is nevertheless clearly the fastest overall. If you consider it for inclusion I could probably improve it for the worst cases left after correction of the clang inefficiencies. Otherwise it is not worth my spending any more time.

nemequ:
@rosbif, can you clarify *which* version of your code you like best? 2000f20ee2b1bce6e931098c38db84e1a0e3a3b0?  Whichever one you like I think we should use for the `SIMDE_TERNARYLOGIC_COMPRESSION` version for now.

I'll start filing some bug reports this week based on your results.  Obviously clang is the priority here, but I have a feeling the GCC folks will want to take a look to; this is a great test case to use for improving optimizations.

Meanwhile, @simba611, once @rosbif confirms which version he'd like to use can you update your patch to use that and finish up the other ternarylogic functions?

simba611:
Yes, I will do that

rosbif:
> @rosbif, can you clarify _which_ version of your code you like best? [2000f20](https://github.com/simd-everywhere/simde/commit/2000f20ee2b1bce6e931098c38db84e1a0e3a3b0)? Whichever one you like I think we should use for the `SIMDE_TERNARYLOGIC_COMPRESSION` version for now.
> 

It is really a question of "horses for courses": the best version depends on your criteria: compact code, avoiding clang inefficiencies, compilation time, support by all compilers, speed.
I put five versions in tl-rosbif.c in the rosbif branch of nemequ's ternarylogic-compiler-test selected by ROSBIF_VERSION:
- 1: This version has simple, compact code. The initial version was even simpler without the negate logic which was added to improve worst-case performance (not counting clang inefficiencies).
- 2: This version uses the same high-level algorithm but was written using SIMDe SSE2 functions mainly to try to force clang to avoid generating really stupid code. It should work everywhere thanks to SIMDe, but emul compilation is very slow .
- 3: This version contains logic reductions for cases where parts (or all) of the result are independent of one, or more of the input vectors. All the following version also use the same high-level algorithm which could probably be improved to reduce worst-case performance. It also uses SIMDe SSE2 functions and the same remarks apply.
- 4: This is a SIMDE_VECTORIZE version of 3. The initial version of this used u32 but this was changed to u64 as clang generates better (but still bad) code with u64 for many values.
- 5: This is a SIMDE_VECTOR_SUBSCRIPT_OPS version of 3 for which clang generates better code than with SIMDE_VECTORIZE. However I think that compiler support for this option is worse.

Versions 3, 4 and 5 are also in ternarylogic.h in my rosbif fork of simde, together with the u32 version of 4.

If speed is the main criterion I think that 5 is best, falling back on 4 where there is no SIMDE_VECTOR_SUBSCRIPT_OPS support. In this case the code can be taken from ternarylogic.h in my rosbif fork of simde by simply removing alternative code excluded by the pre-processor.

> I'll start filing some bug reports this week based on your results. Obviously clang is the priority here, but I have a feeling the GCC folks will want to take a look to; this is a great test case to use for improving optimizations.
> 

Thank you

rosbif:
Thank you @simba611.
I haven't tested the result but the code looks good.
However the header with copyright information seems to be missing.

simba611:
> However the header with copyright information seems to be missing.
Hope this is fine.

PR #823, opened 2021-06-04
neon/qshluq_n: native and portable implementations

Glitch18:
Implements:
 - vqshlu_n_s*
 - vqshluq_n_s*

Also fixes incorrect tests for `shll_n`


nemequ (changes requested):
This looks good overall, but can merge the implementations a bit?  There are a lot of places where you declare variables three times (once for the vector extension version, once for the wasm version, and once for the portable version), then also call the `*_from_private` function three times.  If you merge these there should be less repetition in the code, and it should be a bit easier to read.

Glitch18:
Yes. I see that now. Will be careful about that next time onwards. I'll make the changes and push a commit.

nemequ:
Merged as 77af9f12e93eacd5cf107faaf7c244d46c5c167f, thanks!

PR #826, opened 2021-06-08
neon/tbl: Wasm optimization

ngzhian:


nemequ:
Argh, I just did these, too.  I should have looked at the PRs *before* I wrote the code instead of after.

I'm going to use mine for this one; it generates the same code, but avoids aliasing.  I'll take a look at the others right now.

PR #827, opened 2021-06-08
neon/st3q_u8: Wasm optimization

ngzhian:
Autovectorizer is unable to do anything here, so implement it using a
bunch of shuffles.


nemequ:
Nice ☺

PR #833, opened 2021-06-13
neon/st3: Add shuffle vector implementations

Glitch18:
Adds SIMDE_SHUFFLE_VECTOR implementations for `vst3`


Glitch18:
That's strange. I had confirmed the commit was passing all CI tests at my fork. It looks like the emscripten build's error is unrelated to the code.

nemequ:
Yep, it was a failure to install emscripten.  I restarted the build.

I think we're at a bit over 50 builds in CI, and most (all?) require installing packages from the network, so an occasional spurious failure is *fairly* common.

PR #834, opened 2021-06-14
neon: Complete Implementations of all incomplete function families

Glitch18:
Includes the functions from the following famillies :

 - sub
 - shr_n
 - sra_n
 - fma_lane
 - dup_lane
 - neg
 - padd
 - pmax
 - pmin
 - qdmulh
 - rndn
 - qrdmulh
 - rshr_n
 - qrshrun_n
 - rshl


Glitch18:
I added a custom function in the `rshr_n` family. I had not used the correct naming scheme for custom functions. I'm adding the corrections as a separate commit instead of doing a force push. 

Changes `simde_vrshrs_n_s32` to `simde_x_rshrs_n_s32`.
Removes native macros.

nemequ (changes requested):
The qrdmulh_lane stuff throws tons of `-Wvector-conversion` diagnostics on clang prior to version 10 (you can reproduce using the stable docker container; `./simde-dev.sh stable` to build and run it), or look at the results from drone.io.

Unfortunately which gets pulled into the caller's code we can't just wrap everything with a `SIMDE_DIAGNOSTIC_DISABLE_VECTOR_CONVERSION_`.  The most straightforward workaround would probably be to user statement exprs on clang to push the warning stack then use `SIMDE_DIAGNOSTIC_DISABLE_VECTOR_CONVERSION_`, call the native function, then pop the warning stack.

PR #837, opened 2021-06-17
avx512/scalef: initial scalef implementation

simba611:
Implements mm_scalef_ps, x_mm_flushsubnormal_ps.
The tests are generated live for the flushsubnormal function. I can switch it to load a (SIMDE_MATH_FLT_MIN / 2.0f) containing vector and test against FLOAT32_C(0.0) in the output of the function.


simde/simde-math.h
```diff
@@ -322,6 +322,16 @@ SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
   #endif
 #endif
 
+#if !defined(simde_math_issubnormalf)
+  #if SIMDE_MATH_BUILTIN_LIBM(fpclassify)
+    #define simde_math_issubnormalf(v) __builtin_fpclassify(0, 0, 0, 1, 0, v)
+  #elif defined(fpclassify)
+    #define simde_math_issubnormalf(v) (fpclassify(v) == FP_SUBNORMAL)
+  #elif defined(SIMDE_IEEE754_STORAGE)
+    #define simde_math_issubnormalf(v) (((simde_float32_as_uint32(v) & UINT32_C(0x7F800000)) == UINT32_C(0)) && ((simde_float32_as_uint32(v) & UINT32_C(0x007FFFFF)) != UINT32_C(0)));
```
nemequ, line 331: I think the trailing semicolon here is what is causing the error on MSVC.

simba611: thanks, fixed it

PR #838, opened 2021-06-17
avx512/ror: implements ror family

simba611:
Implements mm{256, 512}{_mask(z)}_ror_epi32/64.


simde/x86/avx512/ror.h
```diff
@@ -0,0 +1,276 @@
...
+    simde__m128i retval;
+
+    if(imm8 & 31)
+      retval = result;
+    else
+      retval = a;
+
+    return retval;
+  }
+  #define simde_mm_ror_epi32(a, imm8) simde_mm_ror_epi32_internal_(a, simde_mm_or_si128(simde_mm_srli_epi32(a, (imm8 & 31)), simde_mm_slli_epi32(a, (32 - (imm8 & 31)))), imm8)
```
nemequ, line 31: This won't work.  You're using `a` multiple times in the macro, which can easily cause horrible problems.  An easy to understand example would be something like

```c
r = simde_mm_ror_epi32((value = simde_mm_add_epi32(value, other_value)), 4);
```

However, it can be a lot more subtle.  For example, what if instead of `simde_mm_add_epi32` they pass the result of a function which does some expensive computation?  Or if you have a pass-by-reference parameter it's really easy to mess something up, especially in C++ where the function call looks exactly the same.

It's okay to do this sort of thing with the `imm8` parameter because they are required to be an integer constant expression, but for "normal" parameters it's a big problem.  And, since that's an issue on the *caller* side, there isn't really anything we can do to prevent it in SIMDe.

For most compilers you can work around the issue using a [statement expression](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html); we even have a macro (`SIMDE_STATEMENT_EXPR_`) which you can check to see if statement expressions are supported.  That would look something like:

```c
#if defined(SIMDE_STATEMENT_EXPR_)
  #define simde_mm_ror_epi32(a, imm8) \
    SIMDE_STATEMENT_EXPR_(({ \
      simde_mm_ror_epi32_a_ = a; \
      simde_mm_ror_epi32_internal_( \
        simde_mm_ror_epi32_a_, \
        simde_mm_or_si128( \
          simde_mm_srli_epi32(simde_mm_ror_epi32_a_, \
            (imm8 & 31) \
          ), \
          simde_mm_slli_epi32( \
            simde_mm_ror_epi32_a_, \
            (32 - (imm8 & 31)) \
          ) \
        ), \
      imm8) \
    }))
#endif
```

However, this doesn't work on MSVC.

This might mean that you can't use `simde_mm_srli_epi32` and `simde_mm_slli_epi32`, but that's probably okay in this case.  This is a pretty simple function which I would expect most compilers to be able to autovectorize well, especially those which support GCC-style vector extensions (where you can just do `(a_.u32 >> count) | (a_.u32 << count)`, though of course that *also* doesn't include MSVC).

simde/x86/avx512/ror.h
```diff
@@ -0,0 +1,276 @@
...
+  SIMDE_FUNCTION_ATTRIBUTES
+  simde__m128i
+  simde_mm_ror_epi32_internal_(simde__m128i a, simde__m128i result, int imm8)
+      SIMDE_REQUIRE_CONSTANT_RANGE (imm8, 0, 255) {
+    simde__m128i retval;
+
+    if(imm8 & 31)
+      retval = result;
+    else
+      retval = a;
```
nemequ, line 27: For cases like this it's generally a good idea to get into the habit of using the ternary operator.  In this case, the whole function could just be

```c
return (imm8 & 31) ? result : a;
```

Not only does that make the code a bit easier to read IMHO (as long as you understand the ternary operator, which I think is a reasonable expectation for people reading the SIMDe source code), it also eliminates the if / else.  When writing performance-sensitive code (especially when we want the compiler to vectorize) if/else is a bit of a code smell; in this case I'm sure any compiler will handle it exactly the same, but if you were doing something a bit more complicated like:

```c
if (imm8 & 31)
  foo = result;
else
  bar = result;
```

That will cause a branch in the compiled code, and branches are *slow*.  The ternary operator is really nice because you can't really do this sort of general-purpose branching; all you can do is "based on condition, the result is either X or Y", which is generally handled much better in hardware.  On the scalar side, you have the `cmov` instruction on x86, and in SIMD you have various bitselect/blend functions (see `wasm_v128_bitselect` in WASM, it has a bunch of implementations on other architectures).

nemequ (changes requested):

simde/x86/avx512/ror.h
nemequ: You should also be wrapping variables used in the macro in parenthesis.  For a discussion of why this is important, see https://stackoverflow.com/questions/7186504/c-macros-and-use-of-arguments-in-parentheses, but it comes down to the same issue; you're not passing the result of the expression, you're passing the entire expression (which may have side-effects).

simba611: Thanks, I will fix this and keep that in mind

simde/x86/avx512/ror.h
simba611: I agree, that is what I tried originally but ran into that bug with icc we talked about.

nemequ: That's the issue with ICC? I thought the icc bug was the `SIMDE_VECTORIZE_` on the loop.

That's rather disconcerting, actually.  That should definitely go in an ifdef, then, otherwise someone is likely to come along and "fix" it to be just `return (imm8 & 31) ? result : a;`  It sounds like this is going to have to go away anyways because of the macro issue above, though…

PR #839, opened 2021-06-19
avx512/conflict: implements _mm_conflict_epi32

simba611:
Implements mm_conflict_epi32.
All tests were generated using intel-all-gcc-10.


nemequ:
Oh, that's a known issue with ICC; you can't use OpenMP SIMD annotations on a loop of SIMD instructions.

To be fair, I can't see that ever working anyways.

simde/x86/avx512/conflict.h
```diff
@@ -0,0 +1,39 @@
...
+  #else
+    simde__m128i_private
+      r_ = simde__m128i_to_private(simde_mm_setzero_si128()),
+      a_ = simde__m128i_to_private(a);
+
+    #if !defined(__INTEL_COMPILER)
+      SIMDE_VECTORIZE
+    #endif
+    for (size_t i = 1 ; i < (sizeof(r_.i32) / sizeof(r_.i32[0])) ; i++) {
+      r_.i32[i] = HEDLEY_STATIC_CAST(int32_t, simde_mm_movepi32_mask(simde_mm_cmpeq_epi32(simde_mm_set1_epi32(a_.i32[i]), a))) & ((1 << i) - 1);
```
nemequ, line 25: Be careful with functions like `simde_mm_movepi32_mask` which return or use a predicate type (`__mmask*`); they are usually quite slow outside of AVX-512.

Luckily this particular function can be emulated pretty well using `_mm_movemask_ps` on x86, and AArch64 isn't *too* bad, so in this case it's okay.  I've just noticed you doing similar things for other functions, too, and there are some pretty big performance cliffs here so I wanted to warn you :)

PR #841, opened 2021-06-21
wasm: add simde_wasm_i64x2_ne

coderzh:
add this missing function: simde_wasm_i64x2_ne

### DIFF
--- a/simde/wasm/simd128.h
+++ b/simde/wasm/simd128.h
@@ -1402,6 +1402,35 @@ simde_wasm_i32x4_ne (simde_v128_t a, simde_v128_t b) {
 
 SIMDE_FUNCTION_ATTRIBUTES
 simde_v128_t
+simde_wasm_i64x2_ne (simde_v128_t a, simde_v128_t b) {
+  #if defined(SIMDE_WASM_SIMD128_NATIVE)
+    return wasm_i64x2_ne(a, b);
+  #else
+    simde_v128_private
+      a_ = simde_v128_to_private(a),
+      b_ = simde_v128_to_private(b),
+      r_;
+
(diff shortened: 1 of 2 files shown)


simde/wasm/simd128.h
```diff
@@ -1400,6 +1400,35 @@ simde_wasm_i32x4_ne (simde_v128_t a, simde_v128_t b) {
...
+simde_wasm_i64x2_ne (simde_v128_t a, simde_v128_t b) {
+  #if defined(SIMDE_WASM_SIMD128_NATIVE)
+    return wasm_i64x2_ne(a, b);
+  #else
+    simde_v128_private
+      a_ = simde_v128_to_private(a),
+      b_ = simde_v128_to_private(b),
+      r_;
+
+    #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```
nemequ, line 1414: This actually needs to be `SIMDE_ARM_NEON_A64V8_NATIVE`; `vceqq_s64` is only available on AArch64.

nemequ (changes requested):
Thanks!

This looks good overall, I'll definitely be happy to merge once the Arm test is fixed ☺

simde/wasm/simd128.h
nemequ: Actually, I just realized `vmvnq_u64` (somewhat inexplicably) doesn't exist at all on arm, so really you'll need something like

```c
      r_.neon_u32 = vmvnq_u32(vreinterpretq_u32_u64(vceqq_s64(a_.neon_i64, b_.neon_i64)));
```

coderzh: Yes, I make this mistake. Now the new commit fixed.

nemequ:
Merged, thanks!

PR #844, opened 2021-06-23
avx512/multishift: initial implementation

simba611:
Implements mm_multishift_epi64_epi8.


nemequ:
LGTM, go ahead and the others.

PR #850, opened 2021-06-29
avx512/dbsad: initial implementation

simba611:
Implements mm_dbsad_epu8.


simde/x86/avx512/dbsad.h
```diff
@@ -0,0 +1,54 @@
...
+SIMDE_BEGIN_DECLS_
+
+#if defined(SIMDE_X86_AVX512BW_NATIVE) && defined(SIMDE_X86_AVX512VL_NATIVE)
+  #define simde_mm_dbsad_epu8(a, b, imm8) _mm_dbsad_epu8((a), (b), (imm8))
+#else
+  SIMDE_FUNCTION_ATTRIBUTES
+  simde__m128i
+  simde_mm_dbsad_epu8_internal_ (simde__m128i a, simde__m128i b, int imm8)
+      SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 255) {
+    HEDLEY_STATIC_CAST(void, imm8);
```
nemequ, line 17: Since you only use it in the preprocessing step, why not just get rid of the imm8 parameter in the function?

simba611: Done, sorry about that.

nemequ:
Okay, LGTM, go ahead and add the others.

PR #852, opened 2021-07-01
avx512/dpbusd: initial implementation

simba611:
Implements mm_dpbusd_epi32.


nemequ (changes requested):
This looks good, but I think a `SIMDE_SHUFFLE_VECTOR_` + `SIMDE_CONVERT_VECTOR_` based implementation would be a good idea.  Basically, shuffle to a vector of 8-bit values, convert to 32-bit values, then multiply.

Compilers aren't always good about recognizing stuff like this (especially the shuffle part), so it's good to help them when we can.  It's also worth noting that a lot of this depends on the backend, so just because x86 works well you can't assume others will.

simde/x86/avx512/dpbusd.h
```diff
@@ -0,0 +1,55 @@
...
+      src_ = simde__m128i_to_private(src),
+      a_ = simde__m128i_to_private(a),
+      b_ = simde__m128i_to_private(b);
+    #if defined(SIMDE_SHUFFLE_VECTOR_) && defined(SIMDE_CONVERT_VECTOR_)
+      simde__m512i_private
+        r1_,
+        r2_;
+
+      a_.u8 = SIMDE_SHUFFLE_VECTOR_(8, 16, a_.u8, a_.u8, 0, 4, 8, 12, 1, 5, 9, 13, 2, 6, 10, 14, 3, 7, 11, 15);
+      b_.i8 = SIMDE_SHUFFLE_VECTOR_(8, 16, b_.i8, b_.i8, 16, 20, 24, 28, 17, 21, 25, 29, 18, 22, 26, 30, 19, 23, 27, 31);
```
nemequ, line 26: One (very) minor issue: you should add a space before the single-digit parameters on line 25 so the arguments line up.  It just helps make the code a little more readable since you can see which values correspond without having to count the arguments.

simde/x86/avx512/dpbusd.h
```diff
@@ -0,0 +1,55 @@
...
+
+      SIMDE_CONVERT_VECTOR_(r1_.u32, a_.u8);
+      SIMDE_CONVERT_VECTOR_(r2_.i32, b_.i8);
+
+      src_.i32 =
+        src_.i32 +
```
nemequ, line 32: I think we discussed (on our call on Saturday) using `src_i32 += ...` instead of `src_.i32 = src_.i32 + ...` and you said you had tried but it didn't work somewhere; did you remember where that was?  There should probably be a comment (at least on the first implementation) explaining the issue since IMHO it's a bit of a code smell.

nemequ (changes requested):
Looks good overall; I added a couple of very minor nitpicks, once those are addressed you can go ahead and implement the rest of the functions.

simde/x86/avx512/dpbusd.h
simba611: Made the changes. I must have been doing something else wrong when I tried += before. It's working now.

PR #854, opened 2021-07-08
avx512/popcnt: initial implementation

simba611:
Implements mm_popcnt_epi16.


simde/x86/avx512/popcnt.h
```diff
@@ -0,0 +1,49 @@
...
+  #if defined(SIMDE_X86_AVX512BITALG_NATIVE) && defined(SIMDE_X86_AVX512VL_NATIVE)
+    return _mm_popcnt_epi16(a);
+  #else
+    simde__m128i_private
+      r_,
+      a_ = simde__m128i_to_private(a);
+
+    #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
+      a_.u16 -= ((a_.u16 >> 1) & 21845);
+      a_.u16 = ((a_.u16 & 13107) + ((a_.u16 >> 2) & 13107));
```
nemequ, line 22: I wouldn't even bother going back and adding a patch for this now, but in the future, similarly to your other PR where I suggested lining up the arguments, lining up the assignments is probably a good idea here:

```c
a_.u16 -= ...
a_.u16  = ...
```

Unfortunately the operations don't really line up, not much you can do there.

simde/x86/avx512/popcnt.h
```diff
@@ -0,0 +1,49 @@
...
+simde_mm_popcnt_epi16 (simde__m128i a) {
+  #if defined(SIMDE_X86_AVX512BITALG_NATIVE) && defined(SIMDE_X86_AVX512VL_NATIVE)
+    return _mm_popcnt_epi16(a);
+  #else
+    simde__m128i_private
+      r_,
+      a_ = simde__m128i_to_private(a);
+
+    #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
+      a_.u16 -= ((a_.u16 >> 1) & 21845);
```
nemequ, line 21: The constants actually make a lot more sense in hexadecimal notation; 21845 = 0x5555, 3855 = 0x0f0f, 257 = 0x0101.  If you change them to hex I'm not saying the code will be *easy* to understand, but it should be a little less difficult…

nemequ (changes requested):
It would probably be a good idea to fall back on the 8-bit version of this function with pairwise addition under at least some conditions (maybe checking the architecture).  NEON and WASM both have 8-bit versions; see https://github.com/simd-everywhere/simde/blob/master/simde/arm/neon/cnt.h and https://github.com/simd-everywhere/simde/blob/11d97ba487339390d39239a9b3973a63e0e7f9e1/simde/wasm/simd128.h#L7063-L7126 

They also have a bunch of good fallback implementations, though they could definitely use some clean-up; remember when we talked about porting the assembly from the WASM issue tracker, and how I first did a straight port then cleaned it up a bit (for wasm_f32x4_floor)? This is the same thing, except I didn't bother with the clean-up part :/

rosbif:
In simde_wasm_i8x16_popcnt aren't the AVX2 and SSSE3 versions effectively identical?
The AVX2 version avoids the double _mm_set_epi8 so it is probably better, although the compiler will probably avoid it anyway.
As AVX2 implies the presence of SSSE3, the AVX2 version could be labelled SSSE3 and the SSSE3 version removed.

nemequ:
> In simde_wasm_i8x16_popcnt aren't the AVX2 and SSSE3 versions effectively identical?
> The AVX2 version avoids the double _mm_set_epi8 so it is probably better, although the compiler will probably avoid it anyway.

Yeah, I think you're right.  I was probably basing the two versions in the linked implementation off of the issue WASM tracker, and the AVX2 version would be different because it uses VEX encoding (and the destination can often be a separate register so you need fewer temporary variables), but the compiler will automatically take care of that (if available the compiler will generally prefer VEX or EVEX encoding even if we only use SSE intrinsics).

> As AVX2 implies the presence of SSSE3, the AVX2 version could be labelled SSSE3 and the SSSE3 version removed.

Yep, sounds good.

nemequ:
Are you interested in cleaning up the implementations a bit? If so, maybe we can work though it a bit together on Saturday?

rosbif:
> Are you interested in cleaning up the implementations a bit? If so, maybe we can work though it a bit together on Saturday?

Are you talking about those in this PR or all the implementations in wasm/simd128.h?
I will certainly look at them but I don't have much free time ATM and my wife has other ideas at how to spend it :-(

> The constants actually make a lot more sense in hexadecimal notation; 21845 = 0x5555, 3855 = 0x0f0f, 257 = 0x0101. If you change them to hex I'm not saying the code will be _easy_ to understand, but it should be a little less difficult…

Agreed 100%

rosbif:
Rather than having multiple copies of complex or moderately complex code (like popcount) wouldn't it be a good idea for SIMDe to have generic versions e.g. simde_generic_popcount_8x16 which can be called from AVX512, WASM, ARM, maybe PPC later ... ?
This could also even be beneficial for other simple, common operations (and, or, xor, andnot, add, sub, shifts, ...).
Ensuring a new host architecture is covered everywhere would be simplified as would be adding a new SIMDe architecture e.g. PPC.

simba611:
> Are you interested in cleaning up the implementations a bit? If so, maybe we can work though it a bit together on Saturday?

Sure

nemequ:
> > Are you interested in cleaning up the implementations a bit? If so, maybe we can work though it a bit together on Saturday?
> 
> Are you talking about those in this PR or all the implementations in wasm/simd128.h?
> I will certainly look at them but I don't have much free time ATM and my wife has other ideas about how to spend it :-(

LOL, oops, I'm sorry; honestly it didn't register to me that you were the one commenting, I thought it was @simba611.

> Rather than having multiple copies of complex or moderately complex code (like popcount) wouldn't it be a good idea for SIMDe to have generic versions e.g. simde_generic_popcount_8x16 which can be called from AVX512, WASM, ARM, maybe PPC later ... ?
> This could also even be beneficial for other simple, common operations (and, or, xor, andnot, add, sub, shifts, ...).
> Ensuring a new host architecture is covered everywhere would be simplified as would be adding a new SIMDe architecture e.g. PPC.

Yeah, I've thought about this, too.  I even thought about making WASM the abstraction layer, but I don't actually think it's very well suited for that.  I think we would have to add a new "generic" (or whatever) backend, with types that would then be added to the existing `*_private` unions.

There are a few drawbacks, and after seeing how many differences there are across APIs I'm not sure how big of a win it would really be.  There are tons of functions like conversions from float->int where the rules for OOB + NaN are different, min/max where NaN is different, etc., and most functions where we could use an abstraction layer aren't as complicated as popcnt so it's not as big a deal.  Plus we would likely end up pulling in some unused code a lot of times (*e.g.*, popcnt isn't available on x86 until AVX-512BITALG).  I think it would still be a net gain for us, just by less than you might initially think (or at least less than *I* initially thought) which is why I didn't bother.

So for me it's a pretty low-priority, long-term item.  There is a good chance I'll never get around to it, since things like SVE (and MSA, MVE, RISC-V V, AltiVec, *etc.*) are much higher priorities and there is still a *ton* to do there.

PR #858, opened 2021-07-15
avx512/dpbusd: implement remaining functions

simba611:
Implements _mm{256,512}{_mask(z)}_dpbusd_epi32.
All tests were generated using intel-all-gcc-10.


simde/x86/avx512/dpbusd.h
```diff
@@ -22,8 +23,24 @@ simde_mm_dpbusd_epi32(simde__m128i src, simde__m128i a, simde__m128i b) {
...
+        );
+      b_.i8 =
+        SIMDE_SHUFFLE_VECTOR_(
+          8, 16,
+          b_.i8, b_.i8,
+           0,  4,  8, 12,
+           1,  5,  9, 13,
+           2,  6, 10, 14,
+           3,  7, 11, 15
+        );
```
nemequ, line 43: Nice! I like how you've grouped the items here so it's easy to see the pattern.

simde/x86/avx512/dpbusd.h
```diff
@@ -48,6 +65,208 @@ simde_mm_dpbusd_epi32(simde__m128i src, simde__m128i a, simde__m128i b) {
...
+           1,  5,  9, 13, 17, 21, 25, 29,
+           2,  6, 10, 14, 18, 22, 26, 30,
+           3,  7, 11, 15, 19, 23, 27, 31
+        );
+
+      SIMDE_CONVERT_VECTOR_(r1_.u32, a_.u8);
+      SIMDE_CONVERT_VECTOR_(r2_.i32, b_.i8);
+
+      src_.i32 +=
+        (HEDLEY_REINTERPRET_CAST(__typeof__(a_.i32), r1_.m256i_private[0].u32) * r2_.m256i_private[0].i32) +
```
nemequ, line 136: This is actually undefined behavior. Contrary to popular belief, type-punning through a union isn't allowed.  Unions allow you to choose which type you access something as, but once you choose a type you have to stick to it.  Since you've written to the object as a vector of unsigned 8-bit integers, you can't read it as an array of `simde__m256i_private`.

nemequ (changes requested):

PR #863, opened 2021-07-17
avx512/popcnt: implement remaining functions

simba611:
Implements mm{256,512}{_mask(z)}_popcnt_epi8/16/32/64.

### DIFF
--- a/simde/x86/avx512/popcnt.h
+++ b/simde/x86/avx512/popcnt.h
@@ -2,6 +2,7 @@
 #define SIMDE_X86_AVX512_POPCNT_H
 
 #include "types.h"
+#include "mov.h"
 
 HEDLEY_DIAGNOSTIC_PUSH
 SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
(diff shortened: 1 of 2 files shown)


simde/x86/avx512/popcnt.h
```diff
@@ -310,6 +473,953 @@ simde_mm_popcnt_epi64 (simde__m128i a) {
...
+  #else
+    simde__m256i_private
+      r_,
+      a_ = simde__m256i_to_private(a);
+
+    #if SIMDE_NATURAL_VECTOR_SIZE_LE(128)
+      for (size_t i = 0 ; i < (sizeof(r_.m128i) / sizeof(r_.m128i[0])) ; i++) {
+        r_.m128i[i] = simde_mm_popcnt_epi8(a_.m128i[i]);
+      }
+    #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```
nemequ, line 518: The Arm and WASM cases here aren't necessary; actually, they won't even be hit since `SIMDE_NATURAL_VECTOR_SIZE_LE(128)` will be true in both cases.  Even if you reordered them so they came first, though, it's just doing the exact same thing as the `SIMDE_NATURAL_VECTOR_SIZE_LE(128)`, so we may as well not duplicate the code.

nemequ (changes requested):
Basically looks good.  Once you remove the unnecessary Arm/WASM/Altivec cases I'll merge it.

PR #884, opened 2021-07-29
if `_M_ARM64EC`, use ARM intrinsics

tommyvct:
https://github.com/simd-everywhere/simde/pull/877#issuecomment-889525106

### DIFF
--- a/simde/simde-arch.h
+++ b/simde/simde-arch.h
@@ -70,7 +70,9 @@
 /* AMD64 / x86_64
    <https://en.wikipedia.org/wiki/X86-64> */
 #if defined(__amd64__) || defined(__amd64) || defined(__x86_64__) || defined(__x86_64) || defined(_M_X64) || defined(_M_AMD64)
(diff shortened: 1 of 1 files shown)


simde/simde-arch.h
```diff
@@ -67,12 +67,6 @@
 #  define SIMDE_ARCH_AVR __AVR_ARCH__
 #endif
 
-/* AMD64 / x86_64
-   <https://en.wikipedia.org/wiki/X86-64> */
-#if defined(__amd64__) || defined(__amd64) || defined(__x86_64__) || defined(__x86_64) || defined(_M_X64) || defined(_M_AMD64)
-#  define SIMDE_ARCH_AMD64 1000
-#endif
-
```
nemequ, line 75: This seems unnecessary; why not just leave it untouched?  AFAICT all you should need are the two lines which add ` || defined(_M_ARM64EC)` to get the compiler to use the native versions there…

PR #894, opened 2021-08-04
avx512/fixupimm: initial implementation

simba611:
Implements mm_fixupimm_ps.

### DIFF
--- a/simde/simde-math.h
+++ b/simde/simde-math.h
@@ -410,10 +410,27 @@ simde_math_fpclassifyf(float v) {
   #else
     return
       simde_math_isnormalf(v) ? SIMDE_MATH_FP_NORMAL    :
-      (v == 0.0f)           ? SIMDE_MATH_FP_ZERO      :
+      (v == 0.0f)             ? SIMDE_MATH_FP_ZERO      :
       simde_math_isnanf(v)    ? SIMDE_MATH_FP_NAN       :
       simde_math_isinff(v)    ? SIMDE_MATH_FP_INFINITE  :
                                 SIMDE_MATH_FP_SUBNORMAL;
+  #endif
+}
+
+static HEDLEY_INLINE
+int
+simde_math_fpclassify(double v) {
+  #if SIMDE_MATH_BUILTIN_LIBM(fpclassify)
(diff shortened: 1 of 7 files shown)


simde/simde-math.h
```diff
@@ -235,6 +235,20 @@ SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
...
+    #define SIMDE_MATH_FLT_MAX __FLT_MAX__
+  #elif defined(__cplusplus)
+    #include <cfloat>
+    #define SIMDE_MATH_FLT_MAX FLT_MAX
+  #else
+    #include <float.h>
+    #define SIMDE_MATH_FLT_MAX FLT_MAX
+  #endif
+#endif
+
```
nemequ, line 251: I like that you're doing this instead of just using FLT_MAX/DBL_MAX directly, but I think something like this would be cleaner:

```c
#if !defined(SIMDE_MATH_FLT_MAX)
  #if defined(__FLT_MAX__)
    #define SIMDE_MATH_FLT_MAX __FLT_MAX__
  #else
    #if !defined(FLT_MAX)
      #if defined(__cplusplus)
        #include <cfloat>
      #else
        #include <float.h>
      #endif
    #endif
    #define SIMDE_MATH_FLT_MAX FLT_MAX
  #endif
#endif
```

It may be slightly longer, but it keeps the FLT_MAX case together.

Maybe it's just me, though?  I know some people like to limit nested logic as much as possible…

simde/x86/avx512/fixupimm.h
```diff
@@ -0,0 +1,113 @@
...
+#define SIMDE_X86_AVX512_FIXUPIMM_H
+
+#include "types.h"
+#include "flushsubnormal.h"
+
+HEDLEY_DIAGNOSTIC_PUSH
+SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
+SIMDE_BEGIN_DECLS_
+
+#if defined(SIMDE_X86_AVX512F_NATIVE) && defined(SIMDE_X86_AVX512VL_NATIVE)
```
nemequ, line 11: I notice you've gotten into the habit of defining the function conditionally lately, but that's not what we usually do in SIMDe.  We unconditionally define the function, then define a macro to override the function.  This is mostly because you can do things with a function that you can't with a macro; for example, use them reliably in a CONSTIFY macro.

When the function is nonsense without the supporting macro, like it is with some of those functions where you create an internal function, then wrap it with a macro which handles the immediate-mode parameter, you should only define the function conditionally.  In cases like this, though, please define the function unconditionally.

nemequ (changes requested):

PR #895, opened 2021-08-05
neon/ld1q: u8_x2, u8_x3, u8_x4

ngzhian:
Create a new test file ld1.c to test these new additions that load to
multiple registers.


ngzhian:
@shanumantesc fyi

ngzhian:
@nemequ ready for review, ptal, not sure while some CIs are stuck, but the Github Actions are green.

PR #899, opened 2021-08-12
Fix warnings with -fno-lax-vector-conversions

### LINKED ISSUE #898: Compile failures with `-fno-lax-vector-conversions`
Command: `clang simde/arm/neon/qsub.h  -Werror -O0 -Wno-c++11-narrowing -I.`

Seems like there is an issue here:
https://github.com/simd-everywhere/simde/blob/master/simde/arm/neon/qsub.h#L331

```
simde/arm/neon/qsub.h:331:17: error: cannot convert between vector type '__attribute__((__vector_size__(2 * sizeof(int)))) int' (vector of 2 'int' values) and vector type '__attribute__((__vector_size__(2 * sizeof(uint32_t)))) uint32_t' (vector of 2 'uint32_t' values) as implicit conversion would cause truncation
      r_.values &= (r_.values <= a_.values);
                ^
simde/arm/neon/qsub.h:360:17: error: cannot convert between vector type '__attribute__((__vector_size__(1 * sizeof(long long)))) long long' (vector of 1 'long long' value) and vector type '__attribute__((__vector_size__(1 * sizeof(uint64_t)))) uint64_t' (vector of 1 'uint64_t' value) as implicit conversion would cause truncation
      r_.values &= (r_.values <= a_.values);
                ^
simde/arm/neon/qsub.h:666:17: error: cannot convert between vector type '__attribute__((__vector_size__(2 * sizeof(long long)))) long long' (vector of 2 'long long' values) and vector type '__attribute__((__vector_size__(2 * sizeof(uint64_t)))) uint64_t' (vector of 2 'uint64_t' values) as implicit conversion would cause truncation
      r_.values &= (r_.values <= a_.values);
```

Do we need a reinterpret cast here?

ngzhian:
Fix #898.

### DIFF
--- a/simde/arm/neon/cvt.h
+++ b/simde/arm/neon/cvt.h
@@ -630,7 +630,7 @@ simde_vcvtq_u64_f64(simde_float64x2_t a) {
       static const double min_representable SIMDE_VECTOR(16) = { SIMDE_FLOAT64_C(0.0), };
       r_.values &= HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), a_.values > min_representable);
 
-      r_.values &= (a_.values == a_.values);
+      r_.values &= HEDLEY_REINTERPRET_CAST(__typeof__(r_.values), (a_.values == a_.values));
     #else
       SIMDE_VECTORIZE
       for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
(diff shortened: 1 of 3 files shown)


nemequ:
Nice, thanks!

I think it would also be a good idea to add this flag to the clang builds on CI; TBH I thought `-Weverything` covered it, but I guess not ☹.  I'll try to add a patch for that soon.

ngzhian:
I am surprised -Weverything doesn't cover it, but maybe this flag doesn't count as a diagnostic under W (since it begins with `-f`)? Another reason might be https://bugs.llvm.org/show_bug.cgi?id=17164#c15 this cannot be turned on since it breaks clang, so maybe that's why it's not included in -Weverything.

ngzhian:
Btw, tried adding "'-fno-lax-vector-conversions', '-Wno-c++11-narrowing'" to emscripten.cross, lots of errors :(

PR #907, opened 2021-08-30
Fix defines of vreinterpret to work with templated callers

ngzhian:


ngzhian:
Interesting failure on emscripten that I cannot reproduce locally.

ngzhian:
Upstream bug: https://lists.llvm.org/pipermail/llvm-dev/2021-August/152411.html
It is already reverted, probably have to wait for the revert to roll into emsdk.

ngzhian:
@nemequ this is ready.

PR #909, opened 2021-09-14
added checks so `windows.h` is only included when it is actually need…

boris-kuz:
…ed (to use `MemoryBarrier`).

If you follow the preprocessor define checks, you'll see that MemoryBarrier is only used in very specific circumstances, and not just when `_WIN32` is defined. These additional checks make sure that Windows.h is only included in these circumstances.

### DIFF
--- a/simde/x86/sse.h
+++ b/simde/x86/sse.h
@@ -32,7 +32,7 @@
 
 #include "mmx.h"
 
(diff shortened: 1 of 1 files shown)


nemequ:
Merged, thanks!  Sorry about the delay :(

PR #929, opened 2021-10-13
neon/mla_lane: initial implementation using mla+dup

ngzhian:


mr-c:
Thank you @ngzhian !

MSVC doesn't like it when we combine CONSTIFY and macro defined functions, so I added a workaround and queued this up for merge in 0b593ce

PR #934, opened 2021-12-12
Add missing `static const` in simde-math.h. NFC

sbc100:
The float versions below already have `static const` so I assume these
should be too.


mr-c (approved):

mr-c:
Thank you @sbc100 ; this was merged as https://github.com/simd-everywhere/simde/commit/6bd6562198594b7eb719c657b457362cb05ebcb0

PR #946, opened 2022-03-18
Fixed FMA detection macro on msvc

### LINKED ISSUE #945: FMA detection not correct on msvc
Hello,

According to msvc's [documentation](https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-170), no `__FMA__` flag is defined when compiling with msvc, as its availbility is already implied in `__AVX2__`. Current detection macro only checks `__FMA__` thus native FMA instruction will never be used with msvc.

dhbloo:
Fixes #945

### DIFF
--- a/simde/simde-arch.h
+++ b/simde/simde-arch.h
@@ -273,6 +273,9 @@
 #  endif
 #  if defined(__AVX2__)
 #    define SIMDE_ARCH_X86_AVX2 1
+#    if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


mr-c (approved):

mr-c:
Thank you @dhbloo ; this was merged as https://github.com/simd-everywhere/simde/commit/286ba3dab7117a3e92e69405f916046dc6f67b8e

PR #960, opened 2022-07-07
arch: Add LoongArch LASX/LSX support

XiWeiGu:


mr-c:
Thank you @XiWeiGu ; can you suggest how we can test LoongArch using Continuous Integration?

XiWeiGu:
> Thank you @XiWeiGu ; can you suggest how we can test LoongArch using Continuous Integration?

Thank you for your reply. As far as I know there is no qemu or cross-toolchain support not at the time.  I only tested on the native LoongArch machine.

mr-c (approved):
Okay, I'll take this. I hope that a CI solution is added soon, so there aren't regression.

XiWeiGu:
Thank you for your reply.  Continuous Integration can be achieved with the development of LoongArch ecology. I will submit the patch to increase CI in the next few days.

mr-c:
This was merged as d0cc0abacbc4addd210d94ecdd5fa303e0b0e8ca & 31367e2b3263701987e3718a4179994bfd828eee !

jiegec:
> Okay, I'll take this. I hope that a CI solution is added soon, so there aren't regression.

QEMU support for LSX/LASX has already been landed and released, we will just wait for the release of GCC 14, after which we can easily test them on CI.

PR #965, opened 2022-08-29
Properly map __mm functions to __simde_mm

psaab:
The defines for native aliases never mapped to the simde implementation


mr-c (approved):

mr-c:
Thank you @psaab ; this was merged as https://github.com/simd-everywhere/simde/commit/96c963fce0c017982a7a5d670f0f18607393ea11

Issue #905, opened 2021-08-18
simde_mm_slli_epi32 behaviour different between intel and arm


FigBug:
On Intel, 2nd parameter may be a variable, on arm it must be a constant.

mr-c:
Thanks @baconpaul ; do the tests still pass with that change? If so, could you send a PR with additional test data that demonstrates the correctness?

baconpaul:
OK running brew install meson right now :)

baconpaul:
OK a PR change in which seems to fix the problem.

aqrit:
> __m128i _mm_slli_epi32 (__m128i a, int **imm8**)

`pslld` requires an 8-bit immediate.

If you use it incorrectly...
some compiler (but not all) will convert it to a `_mm_sll_epi32(a, _mm_cvtsi32_si128(b))` (or similar).

aqrit:
I don't think simde could provide a portable "fix" for this without introducing performance regressions.

baconpaul:
Oh thank you! So the error is mine not simde. I appreciate it. I'll close my PR.

baconpaul:
Also yes clearly the library shouldn't fix a problem which is a usage problem.

PR #970, opened 2022-10-20
Allow non-immediate values in _mm_slli_epi api

baconpaul:
_mm_slli_epi32 and their ilk allow non-constant values, for instance like `int shift=2; _mm_slli_epi32(a, shift);`. The SIMDE translation of these apis to `vshlq_n_s32` and so on required shift to be immediate mode (functionally constexpr) meaning valid SSE2 code would not compile.

This diff changes the API to `vshlq_s32` with the associated broadcast of the shift register, adds tests, removes the constant constraint on the API. All tests pass on my macos 12.4 / xcode 14 M1 macbook.

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -5862,7 +5862,8 @@ simde_mm_sra_epi32 (simde__m128i a, simde__m128i count) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m128i
 simde_mm_slli_epi16 (simde__m128i a, const int imm8)
-    SIMDE_REQUIRE_CONSTANT_RANGE(imm8, 0, 255)  {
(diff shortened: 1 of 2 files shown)


baconpaul:
As discussed in issue #905 this would solve the problem reported there.

I'd be amazed if my first time contribution to your library met your standards, so please offer any changes you want or just rewrite this completely.

In a c++20 world you would do an if constexpr on argument shift so you could still project the immediate mode version, but with the macro and inline strategy here I don't think it is possible.

Thanks for looking!

baconpaul:
Hi!

I see the formatting check failed; is there some way I can run this check on my mac so I can make sure my submission works?

And any thoughts on that x86 failure? It looks unrelated to my change but I'm not sure how to address it?

Thanks

PR #973, opened 2022-11-15
wasm: f32x4 and f64x2 nearest roundeven

wrv:
According to the [WASM SIMD spec](https://github.com/WebAssembly/simd/blob/main/proposals/simd/SIMD.md#round-to-nearest-integer-ties-to-even), the instructions `f32x4.nearest` and `f64x2.nearest` should round to the nearest even number when the value is equally near. 

This pull request changes `simde_math_round(f)` to  `simde_math_roundeven(f)` to satisfy that requirement, and provides tests to check the even condition.


mr-c:
Thank you @wrv ; this was merged as https://github.com/simd-everywhere/simde/commit/dc75f4c378cd6ae21ed0b85ef4d8c49eb930b7d4

PR #993, opened 2023-02-13
Suppress min/max macro definitions from windows.h

quyykk:
`windows.h` for backwards compatibility defines the `min` and `max` macros, which is a problem because it conflicts with C++ functions that happen to be named min/max as well. Especially since this is a header file and included by other projects (at least once such project, vvenc, sees compilation errors due to this transitive include).

So this PR suppress the definitions of these macros.


mr-c (approved):
Thank you @quyykk !

mr-c:
We are having CI issues due to a recent bug in emscripten, so merging will take a while, but it will happen.

PR #1004, opened 2023-04-21
avx, avx2: fix maskload illegal mem access

### LINKED ISSUE #998: `_mm_maskload_ps` non-native implementation can lead to illegal memory access
The culprit being here, where the whole memory is accessed first, before applying the mask:
https://github.com/simd-everywhere/simde/blob/c900d5e195c4692e0f6b1c3ad8e18e577cf55e11/simde/x86/avx.h#L4011

from the corresponding [x86 docs](https://www.felixcloutier.com/x86/vmaskmov):
> ...
> The mask bit for each data element is the most significant bit of that element in the first source operand. If a mask is 1, the corresponding data element is copied from the second source operand to the destination operand. If the mask is 0, the corresponding data element is set to zero in the load form of these instructions, and unmodified in the store form.
> ...
> Faults occur only due to mask-bit required memory accesses that caused the faults. Faults will not occur due to referencing any memory location if the corresponding mask bit for that memory location is 0. For example, no faults will be detected if the mask bits are all zero.
> ...

<details>

<summary>Code to reproduce</summary>

`main.c`:
```C
#include <simde/x86/avx.h>
#include <stdio.h>
#include <sys/mman.h>


int main(void)
{

    float *ptr = mmap(NULL, 8 * sizeof(float), PROT_NONE , MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    const __m256i mask = _mm256_set_epi32(0, 0, 0, 0, 0, 0, 0, 0);
    __m256 test = _mm256_maskload_ps(ptr, mask);
    printf("test: %f\n", test[1]);
}
```

compile on arm64 silicon (this executable will fail with `Bus Error`):
```
clang -o repro -Isimde -DSIMDE_ENABLE_NATIVE_ALIASES main.c
```

compile on intel (this executable will run fine):
```
clang -o repro -Isimde -march=native main.c
```

</details>

big thanks to @svenpeter42 for helping a lot with debugging this and actually pinning down the issue :)

k-dominik:
makes sure memory is only accessed when mask value is negative, prevents preemptive access that might lead to segfaults.

~~There are also the `maskload` variants in `avx2.h` - should I adapt them accordingly in this PR or should this be done separately?~~

fixes #998

### DIFF
--- a/simde/x86/avx.h
+++ b/simde/x86/avx.h
@@ -3948,19 +3948,23 @@ simde_mm_maskload_pd (const simde_float64 mem_addr[HEDLEY_ARRAY_PARAM(2)], simde
       return _mm_maskload_pd(mem_addr, mask);
     #endif
   #else
-    simde__m128d_private
-      mem_ = simde__m128d_to_private(simde_mm_loadu_pd(mem_addr)),
-      r_;
-    simde__m128i_private mask_ = simde__m128i_to_private(mask);
+    simde__m128d_private r_;
+    simde__m128i_private
+      mask_ = simde__m128i_to_private(mask),
+      mask_shr_;
 
     #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-      r_.neon_i64 = vandq_s64(mem_.neon_i64, vshrq_n_s64(mask_.neon_i64, 63));
+      mask_shr_.neon_i64 = vshrq_n_s64(mask_.neon_i64, 63);
     #else
       SIMDE_VECTORIZE
-      for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
-        r_.i64[i] = mem_.i64[i] & (mask_.i64[i] >> 63);
+      for (size_t i = 0 ; i < (sizeof(mask_.i64) / sizeof(mask_.i64[0])) ; i++) {
+        mask_shr_.i64[i] = mask_.i64[i] >> 63;
       }
     #endif
+      SIMDE_VECTORIZE
+      for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
+        r_.f64[i] = mask_shr_.i64[i] ? mem_addr[i] : 0.0;
+      }
 
     return simde__m128d_from_private(r_);
   #endif
@@ -3983,10 +3987,9 @@ simde_mm256_maskload_pd (const simde_float64 mem_addr[HEDLEY_ARRAY_PARAM(4)], si
     simde__m256d_private r_;
     simde__m256i_private mask_ = simde__m256i_to_private(mask);
 
-    r_ = simde__m256d_to_private(simde_mm256_loadu_pd(mem_addr));
     SIMDE_VECTORIZE
     for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
-      r_.i64[i] &= mask_.i64[i] >> 63;
+      r_.f64[i] = (mask_.i64[i] >> 63) ? mem_addr[i] : 0.0;
     }
 
     return simde__m256d_from_private(r_);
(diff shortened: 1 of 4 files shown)


simde/x86/avx.h
```diff
@@ -3948,19 +3948,23 @@ simde_mm_maskload_pd (const simde_float64 mem_addr[HEDLEY_ARRAY_PARAM(2)], simde
...
-      for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
-        r_.i64[i] = mem_.i64[i] & (mask_.i64[i] >> 63);
+      for (size_t i = 0 ; i < (sizeof(mask_.i64) / sizeof(mask_.i64[0])) ; i++) {
+        mask_shr_.i64[i] = mask_.i64[i] >> 63;
       }
     #endif
+      SIMDE_VECTORIZE
+      for (size_t i = 0 ; i < (sizeof(r_.f64) / sizeof(r_.f64[0])) ; i++) {
+        r_.f64[i] = mask_shr_.i64[i] ? mem_addr[i] : 0.0;
+      }
```
k-dominik, line 3967: I'm actually not sure if the extra path for `SIMDE_ARM_NEON_A32V7_NATIVE` brings much here, since it's not combined anymore with the vectorized bitwise and (`vandq_s64`) and the loop will be done anyways. Does anyone by any chance have an idea or should I look at the assembly?

mr-c: I don't have any insight, sorry. So I would recommend checking the assembly with godbolt.org or similar

simde/x86/avx2.h
```diff
@@ -2854,19 +2859,24 @@ simde_mm_maskload_epi64 (const int64_t mem_addr[HEDLEY_ARRAY_PARAM(2)], simde__m
     return _mm_maskload_epi64(HEDLEY_REINTERPRET_CAST(const long long *, mem_addr), mask);
   #else
     simde__m128i_private
-      mem_ = simde__m128i_to_private(simde_x_mm_loadu_epi64((mem_addr))),
```
mr-c, line 2857: https://github.com/simd-everywhere/simde/actions/runs/4783302859/jobs/8505014399?pr=1004#step:9:25

```
../simde/x86/avx512/../avx2.h:2867:31: error: ‘mem_’ undeclared (first use in this function)
 2867 |       r_.neon_i64 = vandq_s64(mem_.neon_i64, vshrq_n_s64(mask_.neon_i64, 63));
      |                               ^~~~
```

k-dominik: oh right :) thank you!

test/x86/avx.c
```diff
@@ -9785,6 +9791,25 @@ test_simde_mm_maskload_pd (SIMDE_MUNIT_TEST_ARGS) {
   return 0;
 }
 
+#if !defined(HEDLEY_MSVC_VERSION)
+static int
+test_simde_mm_maskload_pd_no_illegal_memory_access (SIMDE_MUNIT_TEST_ARGS) {
```
mr-c, line 9796: New test functions need to be added to the list at the end of the file. Make sure to also surround those entries with `!defined(HEDLEY_MSVC_VERSION)`

https://github.com/simd-everywhere/simde/blob/28246b76b11331f9f235923175accffe90f47546/test/x86/avx.c#L16241

test/x86/avx.c
```diff
@@ -9785,6 +9791,25 @@ test_simde_mm_maskload_pd (SIMDE_MUNIT_TEST_ARGS) {
...
+  // ref: https://github.com/simd-everywhere/simde/issues/998
+  // make sure maskload never accesses memory for masked out regions
+  // will segfault in case memory is accessed
+  #if defined(_GNU_SOURCE)
+    simde_float64 *ptr = HEDLEY_STATIC_CAST(simde_float64 *, mmap(NULL, 2 * sizeof(simde_float64), PROT_NONE , MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  #else
+    simde_float64 *ptr = HEDLEY_STATIC_CAST(simde_float64 *, mmap(NULL, 2 * sizeof(simde_float64), PROT_NONE , MAP_PRIVATE, -1, 0));
+  #endif
+  const simde__m128i mask = simde_mm_set_epi64(INT64_C(0), INT64_C(0));
```
mr-c, line 9805:
```suggestion
  const simde__m128i mask = simde_mm_set_epi64x(INT64_C(0), INT64_C(0));
```

mr-c:
I have a fix for the native-aliases test, one moment

k-dominik:
> I have a fix for the native-aliases test, one moment

So making the pointers const would not have been enough? (because with those it seems to work in the other tests)

mr-c (approved):

k-dominik:
in any case, thanks for bearing with me over this one :)

mr-c:
merged via https://github.com/simd-everywhere/simde/commit/39f723e4e2ccded72decb233b7df5d9537ee2c4f !

PR #1009, opened 2023-05-02
Test some Wasm instructions more strictly

keithw:
Hi, thank you for SIMDe! I'm working on wasm2c and trying to get the full Wasm testsuite passing when using SIMDe as backend for the Wasm simd128 operations. We're seeing a small number of differences from the Wasm-specified behavior (https://github.com/WebAssembly/wabt/pull/2021#issuecomment-1396386135), most involving either negative zero or NaNs.

As a first step, here is a PR that cranks up the strictness of the Wasm tests for demote, floor, max, min, ~nearest,~ pmax, pmin, and trunc. Currently these are tested to within the nearest 0.1; this PR would change that to an exact comparison (which I think is acceptable even in a fast-math situation as these are basically all selectors or integer truncation?). This also changes the overall test.h to do an exact bitwise comparison when slop=0, rather than `==`.

The intention is to write tests that capture the Wasm behavior related to negative zero and implement these instructions in a conforming way. (E.g., Wasm requires that min(-0, 0) → -0 , and checking that requires a bitwise comparison or at least `signbit()`.)

(Update: removed `nearest` from this list because it already has a test that involves negative zero. This test passes on GCC using `__builtin_roundeven` but fails on clang with SIMDe's C implementation of `simde_math_roundeven`.)

### DIFF
--- a/test/test.h
+++ b/test/test.h
@@ -628,7 +628,7 @@ simde_test_equal_f32(simde_float32 a, simde_float32 b, simde_float32 slop) {
   } else if (simde_math_isinf(a)) {
     return !((a < b) || (a > b));
   } else if (slop == SIMDE_FLOAT32_C(0.0)) {
-    return a == b;
+    return !simde_memcmp(&a, &b, sizeof(simde_float32));
   } else {
     simde_float32 lo = a - slop;
     if (HEDLEY_UNLIKELY(lo == a))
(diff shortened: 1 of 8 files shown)


test/test.h
```diff
@@ -9,6 +9,7 @@
 #include <math.h>
 #include <inttypes.h>
 #include <stdarg.h>
+#include <string.h>
```
mr-c, line 12: ?

keithw: Just fixed (by using simde_memcmp instead of memcmp directly).

mr-c (approved):

keithw:
Likewise -- thank you! We'll be really happy when we can check the box for SIMD support in wasm2c.

PR #1010, opened 2023-05-02
Pass WebAssembly testsuite

keithw:
~This should probably be sequenced after #1099 (test Wasm instructions more strictly).~

Here is a PR that makes SIMDe pass the WebAssembly SIMD tests, run via the wasm2c test runner.

The major changes are:
- floating-point min/max: ~reduce to just C versions (removing native asm lowerings)~ adjust C version and adapt V8/Cranelift's SSE2 lowerings (following @mr-c's example) to explicitly check for -0 situations. (Removes other asm lowerings of f32/f64 min/max.)
- floating-point abs: modify the C version to check for -0 (the Intel asm versions seem fine)
- floor/ceil/trunc/sqrt: quiet signaling NaNs in the C versions. The Intel asm versions appear to be fine except the SSE2 implementation of f64x2_trunc, which doesn't seem to handle -0 correctly. The much simpler SSE4.1 version seems fine.
- Switch 'nearest' from using `roundeven` to `nearbyint`, and also quiet signaling NaNs. Make the test (which already had a negative-zero case) enforce bit-exact equivalence.
- reduce q15_mulr_sat to just the C version on Intel. The SSSE3 and SSE2 versions both seem to fail a Wasm test (https://github.com/WebAssembly/spec/blob/main/test/core/simd/simd_i16x8_q15mulr_sat_s.wast#L64-L66). The C version does fine.


mr-c:
> * V8 and Cranelift have conforming lowerings of these Wasm instructions that could be copied here, but it doesn't look trivial.

Got a link?

keithw:
> > * V8 and Cranelift have conforming lowerings of these Wasm instructions that could be copied here, but it doesn't look trivial.
> 
> Got a link?

Sure, here's Cranelift: https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/isa/x64/lower.isle#L2331

And here's V8: https://github.com/v8/v8/blob/main/src/codegen/shared-ia32-x64/macro-assembler-shared-ia32-x64.cc#L202

Some helpful discussion I found: https://github.com/WebAssembly/simd/issues/186

mr-c (changes requested):
I'm happy to see this, though the loss of the optimized implementations is a bit sad.

Can you try to bring some of the optimized implementations up to date?

simde/wasm/simd128.h
```diff
@@ -3115,20 +3127,10 @@ simde_wasm_f32x4_abs (simde_v128_t a) {
       r_.neon_f32 = vabsq_f32(a_.neon_f32);
     #elif defined(SIMDE_POWER_ALTIVEC_P6_NATIVE)
       r_.altivec_f32 = vec_abs(a_.altivec_f32);
-    #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
-      int32_t SIMDE_VECTOR(16) m = HEDLEY_REINTERPRET_CAST(__typeof__(m), a_.f32 < SIMDE_FLOAT32_C(0.0));
```
mr-c, line 3119: Can this be restored, using `simde_math_signbit`?

keithw:
> @keithw ~Once the appveyor CI has finished (which I expect to pass)~ heavy_check_mark can you fix the tabs? https://github.com/simd-everywhere/simde/actions/runs/4860415394/jobs/8664216907?pr=1010#step:5:1

:+1: Just force-pushed.

simde/wasm/simd128.h
keithw: I'm afraid I couldn't think of an obvious way -- signbit takes a scalar so it can't just be substituted in place of the `a_.f32 < 0.0` test. We could run signbit on each lane individually, but at that point I don't think it's going to be better than the (hopefully vectorized) for loop below it. :-/ SIMDe does have SSE2/Neon/Altivec versions of this one at least.

keithw:
> Okay, I was able to do one conversion following these. It appears to pass the tests. Hopefully this is enough of a hint to help you re-implement the other optimizations.

Thank you, that's very helpful! Just force-pushed a version for testing that includes this SSE2 for f32x4min and a similar version for f64x2min. I'll try to come back later today or tomorrow with a working SSE2 for f32/f64 max, which looks pretty doable (but if you or somebody else gets to it first, I won't be offended...).

I think it would be a bigger lift if you also want to preserve the Neon/Altivec versions of f32/f64 min and max at this point, just because I really don't have the expertise or ability to test these locally. :-/

mr-c (approved):
Thanks! I'll queue this up for merging later. Additional optimizations in future PRs are very welcome!

mr-c:
I rebased this branch for final testing over in https://github.com/simd-everywhere/simde/tree/staging

simde/simde-math.h
```diff
@@ -706,6 +706,16 @@ simde_math_fpclassify(double v) {
   #endif
 #endif
 
+#if !defined(simde_math_signbit)
+  #if SIMDE_MATH_BUILTIN_LIBM(signbit)
+    #define simde_math_signbit(x) __builtin_signbit(x)
```
mr-c, line 711: In old versions of clang & GCC, the `__builtin_signbit` function is only for doubles.

clang-8, no error:  https://nemequ.semaphoreci.com/jobs/d4bfe35a-c9d1-4be5-8278-8afa5dfcf0f3
clang-6.0 error: https://nemequ.semaphoreci.com/jobs/4df6cec3-b65b-4d31-b48b-58a436667ecf
clang-4 error: https://nemequ.semaphoreci.com/jobs/f94a6879-6b34-457c-a82a-ca1f1a7c8ac0

How would you like to handle this? Maybe and `#if def` to only add a cast to `double` for these old compilers?

https://android.googlesource.com/platform/prebuilts/clang/darwin-x86/sdk/3.5/+/refs/heads/studio-1.3-release/include/clang/Basic/Builtins.def#374
For Clang, appears this was fixed in version 7.0 https://github.com/llvm/llvm-project/commit/dd0e2b01ae58f3b5c9cf8adfd016c9d239458955

Looks like GCC fixed this sometime after GCC 4.4.0 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36757 ; so I don't think we need GCC specific handling here.

mr-c: I'm testing https://github.com/simd-everywhere/simde/commit/c22e7c5a4d3daff8a02b3dff3c0ccebce8641ddd now

PR #1011, opened 2023-05-05
Update clang version detection for 14..16 and add link

jan-wassenberg:
See https://github.com/google/highway/pull/952


mr-c:
Merged as https://github.com/simd-everywhere/simde/commit/4957a9e32c253d37c96b5484c960280d80c990ab; thanks!

PR #1023, opened 2023-05-21
[NEON] Add qtbl/qtbx polyfills for A32V7

easyaspi314:
Similar to _mm_shuffle_epi8, qtbl/qtbx now split up to multiple shuffles on A32V7. With multi-vector shuffles, two tables are created, the indexes are subtracted, and either TBX'd on the result of the other table or ORR'd together (TBX is mandatory for qtbx, but ORR is chosen when the lookups would be theoretically dual issued).

The code seems like a lot but it really boils down to in the worst case
```
vmov.i8 tmp, #32
vsub.i8 idx_hi, idx, tmp
vtbx.8 lo(a), {t0-t3}, lo(idx)
vtbx.8 hi(a), {t0-t3}, hi(idx)
vtbx.8 lo(a), {t4-t7}, lo(idx_hi)
vtbx.8 hi(a), {t4-t7}, hi(idx_hi)
```

**Note that there are compile errors with A32V7 in other files** (is there even a test vector?) but the qtbl/qtbx tests pass when the errors are fixed.

### DIFF
--- a/simde/arm/neon/qtbl.h
+++ b/simde/arm/neon/qtbl.h
@@ -41,7 +41,8 @@ simde_vqtbl1_u8(simde_uint8x16_t t, simde_uint8x8_t idx) {
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
     return vqtbl1_u8(t, idx);
   #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-    uint8x8x2_t split = {{ vget_low_u8(t), vget_high_u8(t) }};
+    uint8x8x2_t split;
+    simde_memcpy(&split, &t, sizeof(split));
     return vtbl2_u8(split, idx);
   #else
     simde_uint8x16_private t_ = simde_uint8x16_to_private(t);
@@ -90,10 +91,8 @@ simde_vqtbl2_u8(simde_uint8x16x2_t t, simde_uint8x8_t idx) {
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
     return vqtbl2_u8(t, idx);
   #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
-    uint8x8x4_t split = {{
-      vget_low_u8(t.val[0]), vget_high_u8(t.val[0]),
-      vget_low_u8(t.val[1]), vget_high_u8(t.val[1])
-    }};
+    uint8x8x4_t split;
+    simde_memcpy(&split, &t, sizeof(split));
     return vtbl4_u8(split, idx);
   #else
     simde_uint8x16_private t_[2] = { simde_uint8x16_to_private(t.val[0]), simde_uint8x16_to_private(t.val[1]) };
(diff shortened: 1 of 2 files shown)


easyaspi314:
for the reference, existing A32V7 bugs I immediately noticed:
 - `cvtn` is not a thing on v7-a
 - `ld1_f64_xN` is not a thing on v7-a
 - Clang assumes 256 byte alignment on x4 (even though `alignof(uint64x2x4_t) == 16`) and bus errors (not detectable on qemu)

simde/arm/neon/qtbl.h
```diff
@@ -187,6 +211,20 @@ simde_uint8x8_t
 simde_vqtbl4_u8(simde_uint8x16x4_t t, simde_uint8x8_t idx) {
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
     return vqtbl4_u8(t, idx);
+  #elif defined(SIMDE_ARM_NEON_A64V8_NATIVE)
```
mr-c, line 214:
```suggestion
  #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```

simde/arm/neon/qtbl.h
```diff
@@ -135,6 +146,19 @@ simde_uint8x8_t
 simde_vqtbl3_u8(simde_uint8x16x3_t t, simde_uint8x8_t idx) {
   #if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
     return vqtbl3_u8(t, idx);
+  #elif defined(SIMDE_ARM_NEON_A64V8_NATIVE)
```
mr-c, line 149: I think you meant

```suggestion
  #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```

mr-c (changes requested):
Hello @easyaspi314 and thanks for your pull request. I left a few code comments on things that jumped out at me.

Do you want me to add an ARMv7 qemu CI test to this PR?

Did you compare the results of your implementations with the fallback code? Which produces fewer instructions? Perhaps you have the results of timing tests as well.

easyaspi314:
Thanks for the heads up, silly copy paste error of mine.

No compiler is smart enough to naturally generate an instruction as complex as tbl from a scalar fallback loop, and all the splitting is free given that register allocation isn't stupid, so there would naturally be no contest. However I can throw together a quick benchmark on some AArch64 devices in 32-bit mode.

mr-c:
@easyaspi314 even a Compiler Explorer dump showing a reduced number of instructions is probably a good enough sanity check for me. I'm not an expert at all, I'm just the cheerleader / co-maintainer :-)

easyaspi314:
Well I was wrong about the register allocation (GCC is stupid) but even still it is no contest. I can set up a godbolt later today.

mr-c:
> Well I was wrong about the register allocation (GCC is stupid, clang is not) but even still it is no contest compared to scalarization. I can set up a godbolt later today.

Thanks! By then https://github.com/simd-everywhere/simde/commit/2ed0d0e70c472ebd7342f353ba59d758ea8a0e67 will be merged into the trunk, so we can confirm the test results

FYI: https://github.com/emscripten-core/emscripten/pull/19304 may be of interest to you (I can refresh it when you're done with your optimizations, etc..)

easyaspi314:
I had some free time so I pushed the memcpy changes.

Later today I can work on banging out the other A32V7 bugs but these changes don't directly affect emscripten (although it seems like it is missing a simd128 fallback anyways)

mr-c:
@easyaspi314 Are you ready for me to queue this up for merging? Or did you want to double check with Compiler Explorer first?

easyaspi314:
https://godbolt.org/z/bfqaPajMe

Here is a godbolt for vqtbx4q_u8, the most complex function. As you can see there is no contest.

The loads in the loop are solely because calling convention forces the `uint8x16x4_t` to the stack.

mr-c (approved):
Queued for merging as https://github.com/simd-everywhere/simde/commit/a2fef9ebe4b9e380d45a03e3c4703ea5e0f1cc71 ; thanks!

PR #1026, opened 2023-05-21
neon/ld1_xN: Remove f64xN loads for A32, don't trust clang A32 for load multiple

easyaspi314:
- Clang incorrectly sets the alignment bit on 32-bit for the multiple load instructions. This results in a bus error on hardware. This is actually a bug in the compiler itself that would occur without the polyfill. 
 - Don't try to use f64 loads


easyaspi314:
With this all NEON tests pass with Clang ARMv7-a on hardware (Cortex-X1 32-on-64)

mr-c (approved):

mr-c:
Thanks! I've queued this for merging as https://github.com/simd-everywhere/simde/commit/927f141a7bcfea79863604cfaaa6790b2649e59d

PR #1030, opened 2023-05-31
NEON: properly implement _high intrinsics

easyaspi314:
High intrinsics merely have an implicit vget_high or vcombine as a helper for most of the widen or narrow instructions since 64-bit can't address the upper halves of registers anymore. There is no need to complicate them further.


simde/arm/neon/addw_high.h
```diff
@@ -40,19 +40,8 @@ simde_int16x8_t
...
-    simde_int16x8_private r_;
-    simde_int16x8_private a_ = simde_int16x8_to_private(a);
-    simde_int8x16_private b_ = simde_int8x16_to_private(b);
-
-    SIMDE_VECTORIZE
-    for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-      r_.values[i] = a_.values[i] + b_.values[i + ((sizeof(b_.values) / sizeof(b_.values[0])) / 2)];
-    }
-
-    return simde_int16x8_from_private(r_);
```
mr-c, line 55: Hmm.. So you think that there is no architecture/compiler combo that would produce better code from this vectorize loop than the fallback of `simde_vaddw_s8(a, simde_vget_high_s8(b))` ?

easyaspi314: I am mostly going for ease of implementation on this PR.

If the compiler is reasonably intelligent it would be able to detect the redundant assignment/shuffle and eliminate it. However I haven't tested codegen.

easyaspi314: Hold up, the story changes with uint16_t... GCC vomits.

mr-c: With which version does GCC vomit when compiling the `uint16_t` functions: the vectorized or the downscaled version?

easyaspi314: It actually seems to be the opposite problem. The autovec codegen is actually bad on vaddw_u16.

mr-c: > It actually seems to be the opposite problem. The autovec codegen is actually bad on vaddw_u16. GCC couldn't autovec the one-shot one.

So you're seeing better code from this PR for GCC?

easyaspi314: No. Rather it is vaddw_u16 having mediocre codegen and reusing it passes those codegen issues to vaddw_high_u16.

mr-c: Okay. Is this PR ready, or do you want to make other changes?

mr-c:
@easyaspi314 hey-o, does this PR need more work or should I rebase and merge?

PR #1044, opened 2023-06-11
Packit: Add a feature to skip specific tests in the specific CPU cases.

junaruga:
Added the i386 (machine cpu: i686) case to the Packit CI. The CI only runs 1 test (GCC O2 flag test) in the i386 case, because the case takes longer running time than other CPU cases, and fails in the GCC default test.

## Shorter running time

I tested by running Packit CI 2 times in my forked repository. Interestingly the running times are shorter than my test yesterday. Now we can manage the executed tests for each CPU case.

* https://copr.fedorainfracloud.org/coprs/packit/junaruga-simde-wip-ci-packit-skip-cpu/build/6061404/
* https://copr.fedorainfracloud.org/coprs/packit/junaruga-simde-wip-ci-packit-skip-cpu/build/6061407/

## Test

I checked that only one test was executed in the i686 case, and 2 tests were executed as before in other CPU cases.

x86_64

```
== Results ==
Exit status: 0
ok gcc without flags
ok gcc with -O2 flag
skipped gcc with RPM build flags
skipped clang without flags
skipped clang with -O2 flag
skipped clang with RPM build flags
```

i686

The "gcc without flags" test is skipped.

```
== Results ==
Exit status: 0
skipped gcc without flags
ok gcc with -O2 flag
skipped gcc with RPM build flags
skipped clang without flags
skipped clang with -O2 flag
skipped clang with RPM build flags
```

aarch64

```
== Results ==
Exit status: 0
ok gcc without flags
ok gcc with -O2 flag
skipped gcc with RPM build flags
skipped clang without flags
skipped clang with -O2 flag
skipped clang with RPM build flags
```

ppc64le

```
== Results ==
Exit status: 0
ok gcc without flags
ok gcc with -O2 flag
skipped gcc with RPM build flags
skipped clang without flags
skipped clang with -O2 flag
skipped clang with RPM build flags
```

s390x

```
== Results ==
Exit status: 0
ok gcc without flags
ok gcc with -O2 flag
skipped gcc with RPM build flags
skipped clang without flags
skipped clang with -O2 flag
skipped clang with RPM build flags
```

## Bash code analysis tools

As a note, I am checking the bash script code with bash code analysis tools by the `sh -n`, [bashate](https://github.com/openstack/bashate) and [shellcheck](https://www.shellcheck.net/).

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -19,9 +19,7 @@ jobs:
     # "Motivation and context" for details.
     targets:
       - fedora-rawhide-x86_64
-      # Comment out the i386 (i686) case as it takes much longer time.
-      # TODO: Implement a flexible way to run only specific tests.
-      # - fedora-rawhide-i386
+      - fedora-rawhide-i386
       - fedora-rawhide-aarch64
       - fedora-rawhide-ppc64le
       - fedora-rawhide-s390x
--- a/.packit/ci.sh
+++ b/.packit/ci.sh
@@ -3,6 +3,7 @@
 set -euxo pipefail
 
 # Define variables
+HOST_CPU="$(uname -m)"
 JOBS="$(nproc)"
 
 # Functions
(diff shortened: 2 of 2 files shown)


mr-c:
Also, is SIMDe used in Fedora package building with clang at all? If not, then we can exclude clang from the testing matrix for Fedora

junaruga:
> Would it be possible to still run the excluded cases, but then "fail" if they unexpectedly pass?
>
> Otherwise we'll never know to include them later, if they get fixed.

I think if you still want to run the excluded cases, we need to change the implementation

TEST_FOO:
* empty or undefined: execute the test
* "ignore": ignore running the test (same with the current `SKIP_FOO=true`).
* "pending": execute test. But it the test fails, return "pass" (exit status zero). if the test passes unexpectedly, return "fail" (exit status non-zero).

After the implementation, the setting can be like this. I can implement it.

```
TEST_ALL_GCC_DEFAULT=
TEST_ALL_GCC_O2=
TEST_ALL_GCC_RPM="ignore"
TEST_ALL_CLANG_DEFAULT="ignore"
TEST_ALL_CLANG_O2="ignore"
TEST_ALL_CLANG_RPM="ignore"

TEST_ALL_GCC_DEFAULT="pending"
TEST_i686_GCC_DEFAULT="ignore"
```

```

mr-c (approved):

PR #1055, opened 2023-08-18
fix SIMDE_ARCH_X86_SSE4_2 define

cbielow:
when using `/arch=AVX` or similar then SSE4.2 does not get activated (on MSVC at least, but should be similar on g++/clang)

This PR fixes this.

### DIFF
--- a/simde/simde-arch.h
+++ b/simde/simde-arch.h
@@ -269,7 +269,7 @@
 #    if !defined(SIMDE_ARCH_X86_SSE4_1)
 #      define SIMDE_ARCH_X86_SSE4_1 1
 #    endif
-#    if !defined(SIMDE_ARCH_X86_SSE4_1)
+#    if !defined(SIMDE_ARCH_X86_SSE4_2)
 #      define SIMDE_ARCH_X86_SSE4_2 1
 #    endif
 #  endif


mr-c (approved):

mr-c:
@junaruga seems to be an issue with the fedora builds; I'm going to make them not required for now

`/builddir/build/SOURCES/ci.sh: line 15: meson: command not found`

junaruga:
Thanks for the report. I found the CI log.
https://copr.fedorainfracloud.org/coprs/packit/simd-everywhere-simde-1055/build/6315258/

Seeing the x86_64 log file.
https://download.copr.fedorainfracloud.org/results/packit/simd-everywhere-simde-1055/fedora-rawhide-x86_64/06315258-simde/builder-live.log.gz

Perhaps, Fedora Rawhide's build environment might change the PATH setting.

```
+ pip3 install meson==0.55.0
...
Installing collected packages: meson
  WARNING: The script meson is installed in '/builddir/.local/bin' which is not on PATH.
...
+ meson setup build/gcc
/builddir/build/SOURCES/ci.sh: line 15: meson: command not found
...
```

The solution may be adding the `PATH=$PATH:$HOME/.local/bin` in the [ci.sh](https://github.com/simd-everywhere/simde/blob/master/.packit/ci.sh). I will check it later.

Another temporary workaround is to change the `fedora-rawhide-*` to `fedora-latest-*` to use the latest stable version.

https://github.com/simd-everywhere/simde/blob/5e4b3089dbaa54f9b9f46336bcb132bfcb72de4d/.packit.yml#L20C5-L25

```
    targets:
      - fedora-latest-x86_64
      - fedora-latest-i386
      - fedora-latest-aarch64
      - fedora-latest-ppc64le
      - fedora-latest-s390x
```

Issue #1064, opened 2023-09-24
Wrong constant range in simde_vshll_n_XXX intrinsics


M-HT:
In `simde_vshll_n_XXX` intrinsics the constant range is defined as 1-7 for 8-bit vectors, 1-15 for 16-bit vectors and 1-31 for 32-bit vectors (For example [here](https://github.com/simd-everywhere/simde/blob/4d55fc25bcfddc0e2f0c744448557a070dd310c9/simde/arm/neon/shll_n.h#L49C11-L49C11) for `simde_vshll_n_s8`).

These ranges are not correct because:
* In 32-bit arm, the  instruction `vshll` has two encodings. One encoding for shifts 1-7 for 8-bit vectors, 1-15 for 16-bit vectors and 1-31 for 32-bit vectors. And another encoding for shift 8 for 8-bit vectors, 16 for 16-bit vectors a 32 for 32-bit vectors.
* In 64-bit arm, there are instructions `sshll` and `ushll` which have shifts 0-7 for 8-bit vectors, 0-15 for 16-bit vectors and 0-31 for 32-bit vectors, and there's instruction `shll` which has shift 8 for 8-bit vectors, 16 for 16-bit vectors a 32 for 32-bit vectors.

Therefore the ranges in `simde_vshll_n_XXX` intrinsics should be extended at least to 1-8 for 8-bit vectors, 1-16 for 16-bit vectors a 1-32 for 32-bit vectors.

I'm not sure how zero shifts should be handled.

rosbif:
I generally refer to the following document when writing NEON code: 
https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[Neon]&q=vshll

From this document:
- It seems clear that shifts of 0 should be allowed.
- However I can see no mention of shifts of the element size in bits being allowed.
- I don't see any difference between A32 and A64.

Maybe you could change the minima from 1 to 0 and add test cases for these zero length shifts.

Remark:
My name appears in the copyright of shll.h but I am pretty sure that I didn't write it.
It probably started life as a copy of shl.h which I did write.

M-HT:
About zero shifts on A32 - my documentation didn't mention it, but according to [this](https://developer.arm.com/documentation/100076/0200/a32-t32-instruction-set-reference/advanced-simd-instructions--32-bit-/vshll--by-immediate-?lang=en), zero shift is permitted, but resulting instruction is `vmovl`.

[...]

PR #1068, opened 2023-09-26
Extend constant range in simde_vshll_n_XXX intrinsics

M-HT:
This extends constant range in simde_vshll_n_XXX intrinsics according to #1064, including tests for new values.

### DIFF
--- a/simde/arm/neon/shll_n.h
+++ b/simde/arm/neon/shll_n.h
@@ -46,7 +46,7 @@ SIMDE_BEGIN_DECLS_
 SIMDE_FUNCTION_ATTRIBUTES
 simde_int16x8_t
 simde_vshll_n_s8 (const simde_int8x8_t a, const int n)
-    SIMDE_REQUIRE_CONSTANT_RANGE(n, 1, 7) {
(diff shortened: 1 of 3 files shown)


mr-c:
https://github.com/simd-everywhere/simde/actions/runs/6317206038/job/17153486684#step:9:2257

M-HT:
It looks like gcc bug, but when I try vshll_n_XXX intrinsics with zero shift in [Compiler Explorer](https://godbolt.org/), it compiles correctly with ARM GCC 12.3.0 (other versions too), so I'm not sure why it doesn't compile here.

M-HT:
My mistake, Compiler Explorer doesn't compile the resulting code, that's why there's no error.

M-HT:
I filed a [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111609) with gcc.

mr-c:
Thank you @M-HT . Now that we have a bug number, can you add a workaround that disables that test?

See https://github.com/simd-everywhere/simde/blob/6858b928d19842fdc1a48dbd0fb1b57447da86b0/simde/simde-common.h#L915C1-L915C1

mr-c (approved):

PR #1081, opened 2023-10-17
NEON: more fp16 using intrinsics supported by architecture v7 (skip version)

yyctw:
Hi all, this is Eric from Andes Technology Corporation.
This PR is based on the [previous PR](https://github.com/simd-everywhere/simde/pull/1075) with certain functions removed to avoid triggering compiler bugs.


simde/arm/neon/fma.h
```diff
@@ -35,6 +35,20 @@ HEDLEY_DIAGNOSTIC_PUSH
 SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
 SIMDE_BEGIN_DECLS_
 
+SIMDE_FUNCTION_ATTRIBUTES
+simde_float16_t
+simde_vfmah_f16(simde_float16_t a, simde_float16_t b, simde_float16_t c) {
+  #if defined(SIMDE_ARM_NEON_A32V8_NATIVE) && (defined(__ARM_FEATURE_FMA) && __ARM_FEATURE_FMA) && defined(SIMDE_ARM_NEON_FP16)
```
mr-c, line 41:
```suggestion
  #if defined(SIMDE_ARM_NEON_A32V8_NATIVE) && defined(SIMDE_ARCH_ARM_FMA) && defined(SIMDE_ARM_NEON_FP16)
```
And everywhere else, thanks!

simde/arm/neon/cale.h
```diff
@@ -0,0 +1,165 @@
...
+#include "cage.h"
+
+HEDLEY_DIAGNOSTIC_PUSH
+SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
+SIMDE_BEGIN_DECLS_
+
+SIMDE_FUNCTION_ATTRIBUTES
+uint16_t
+simde_vcaleh_f16(simde_float16_t a, simde_float16_t b) {
+  #if defined(SIMDE_ARM_NEON_A64V8_NATIVE) && defined(SIMDE_ARM_NEON_FP16)
```
mr-c, line 39: does `SIMDE_FLOAT16_API_FLOAT16` (using `_Float16` as per the C standard) work with either GCC or clang on arm?

mr-c: Still, https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vcaleh_f16 only listed `A64` as a requirement; is the `defined(SIMDE_ARM_NEON_FP16)` check really needed?

simde/arm/neon/get_lane.h
```diff
@@ -276,7 +276,7 @@ simde_vgetq_lane_f16(simde_float16x8_t v, const int lane)
   simde_float16_t r;
 
   #if defined(SIMDE_ARM_NEON_A32V7_NATIVE) && defined(SIMDE_ARM_NEON_FP16)
-    SIMDE_CONSTIFY_8_(vget_lane_f16, r, (HEDLEY_UNREACHABLE(), SIMDE_FLOAT16_VALUE(0.0)), lane, v);
+    SIMDE_CONSTIFY_8_(vgetq_lane_f16, r, (HEDLEY_UNREACHABLE(), SIMDE_FLOAT16_VALUE(0.0)), lane, v);
```
mr-c, line 279: Odd, the tests already covered `lane` values above `3`; why didn't that cause a problem already?

@yyctw Can you take a look at https://github.com/simd-everywhere/simde/blob/d08d67cf1d9b80fcde34a0ca846f572b590edd45/test/arm/neon/get_lane.c#L505-L518 and investigate why this wasn't caught earlier? Feels like this code path is not being run by the tests..

simde/arm/neon/ld1_x2.h
```diff
@@ -40,6 +41,29 @@ SIMDE_BEGIN_DECLS_
...
+    return vld1_f16_x2(ptr);
+  #else
+    simde_float16x4_private a_[2];
+    for (size_t i = 0; i < 8; i++) {
+      a_[i / 4].values[i % 4] = ptr[i];
+    }
+    simde_float16x4x2_t s_ = { { simde_float16x4_from_private(a_[0]),
+                                 simde_float16x4_from_private(a_[1]) } };
+    return s_;
+  #endif
```
mr-c, line 60: On the x86 side we would have something like a `simde_float16x4x2_private` `typedef union` with `simde_float16x4` and `simde_float16x4_private` fields.

simde/arm/neon/get_lane.h
yyctw: Based on my previous conclusion [[ref](https://github.com/simd-everywhere/simde/pull/1075#discussion_r1358220373)], even if the `HEDLEY_UNREACHABLE()` is executed during the testing process (when `lane > 3`), it will not cause the test to fail.

simde/arm/neon/cale.h
yyctw: According to Wikipedia [[ref](https://en.wikipedia.org/wiki/AArch64#ARMv8.2-A)], the ARMv8-A series architecture supports the A64 instruction set. However, half-precision floating-point support was introduced in ARMv8.2-A. This means that ARMv8.1-A does not support the data type `__fp16`. Instead, ARMv8.1-A defines `SIMDE_ARM_NEON_A64V8_NATIVE`. Therefore, I believe it is necessary to check for the existence of `defined(SIMDE_ARM_NEON_FP16)` to determine whether the architecture supports `fp16`.

PR #1114, opened 2023-12-06
Implement reverse-lane ABI for wasm

### LINKED ISSUE #1113: wasm simd128: "big-endian" ABI support?
Both wasm2c and wasmtime are known for supporting storing v128 lanes in reverse order on big-endian platforms. It would be nice if simde supported it too, as it would simplify wasm2c a lot without adding too much complexity in simde. Since a v128 acts as a slice of wasm memory, it should be aware of how wasm memory is represented on the host. You may want to take a look at webassembly/wabt#2340 for more details.

(We do note that forcing this big-endian representation on little-endian hosts has its uses, even as it goes against the wasm spec. In particular it makes running some regression tests (e.g. in pull requests) much faster than running the full test suite in qemu.)

SoniEx2:
This is probably slower than it needs to be, but we tried to keep it clean more than anything.

Reverse-lane ABI is used in wabt, and even in wasmtime sometimes. Not really aware of any other wasm runtimes that support s390x/big-endian platforms...

(This is also probably wrong somehow, what with the whole "new contributor" thing. Suggestions/feedback/guidance would be appreciated.)

Closes #1113

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -355,6 +355,12 @@ jobs:
           arch_deb: s390x
           distro: ubuntu-22.04
         - version: 12
+          cross: s390x
+          extra: -wasm_reverse_lane_order
+          arch_gnu: s390x
+          arch_deb: s390x
+          distro: ubuntu-22.04
+        - version: 12
           cross: power9
           arch_gnu: powerpc64le
           arch_deb: ppc64el
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,20 @@ jobs:
       apt:
         packages: ninja-build python3-pip gcc g++
 
+  - name: z/Arch WASM_REVERSE_LANE_ORDER
+    if: branch != master OR type == pull_request
+    arch: s390x
+    compiler: gcc
+    env:
+    - CC="gcc"
+    - CXX="g++"
+    - CFLAGS="-march=native -mzvector -Wextra -Werror -DSIMDE_WASM_REVERSE_LANE_ORDER"
+    - CXXFLAGS="-march=native -mzvector -Wextra -Werror -DSIMDE_WASM_REVERSE_LANE_ORDER"
+    - JOBS="$(nproc)"
+    addons:
+      apt:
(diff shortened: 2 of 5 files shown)


mr-c:
Thank you @SoniEx2 for this; how shall we test this in our CI? Do we have to use WABT, or ..?

SoniEx2:
How is simde testing set up? The intended use-case of this feature is to run unmodified wasm (simd) tests on s390x platform, including (simd) endian tests. Certainly, using wasm2c would work, but may be too heavyweight...? (And arguably the cyclic dependency is less than ideal...)

mr-c:
> How is simde testing set up? The intended use-case of this feature is to run unmodified wasm (simd) tests on s390x platform, including (simd) endian tests. Certainly, using wasm2c would work, but may be too heavyweight...? (And arguably the cyclic dependency is less than ideal...)

In https://github.com/simd-everywhere/simde/pull/1114/commits/9212c4a271cd9319759e0054a4380c470f650c67 I added s390x GCC builds using `-DSIMDE_WASM_REVERSE_LANE_ORDER`; should that be enough?

mr-c:
Error log from s390x qemu + GCC 12: https://github.com/simd-everywhere/simde/actions/runs/7129811835/job/19414754673?pr=1114#step:10:1012

mr-c:
Here is some of the test code:

https://github.com/simd-everywhere/simde/blob/471a34285aa6909d5b9b9ff3dcebfa6acf3bce47/test/wasm/simd128/bitmask.c#L62-L64

SoniEx2:
ah, indeed!

since wabt stores the entire wasm memory in reverse byte order (so wasm byte address 0 is at mem[size-1]), it aligns nicely with reverse lane order. but the simde tests assume direct byte order.

SoniEx2:
the solution we would like to see would be for the tests to use `simde_wasm_*_make`, but we no longer feel like these changes would be a particularly good fit for simde. we will just do what we need to do in wabt, tho if anyone wants to revisit this whole thing at some point, they are welcome to do so.

thanks anyway.

PR #1118, opened 2023-12-14
avx512 abs: refine GCC compiler checks for `_mm512{,_mask}_abs_pd`

thomas-schlichter:
Fix check for gcc bug in abs.h .

The gcc bug is described here:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87467

It is fixed in versions 7.4.0 and 8.3.0 and above.

### DIFF
--- a/simde/x86/avx512/abs.h
+++ b/simde/x86/avx512/abs.h
@@ -524,7 +524,7 @@ simde_mm512_mask_abs_ps(simde__m512 src, simde__mmask16 k, simde__m512 v2) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m512d
 simde_mm512_abs_pd(simde__m512d v2) {
-  #if defined(SIMDE_X86_AVX512F_NATIVE) && (!defined(HEDLEY_GCC_VERSION) || HEDLEY_GCC_VERSION_CHECK(8,3,0))
+  #if defined(SIMDE_X86_AVX512F_NATIVE) && !defined(SIMDE_BUG_GCC_87467)
     return _mm512_abs_pd(v2);
   #elif defined(SIMDE_X86_AVX512F_NATIVE) && (!defined(HEDLEY_GCC_VERSION) || HEDLEY_GCC_VERSION_CHECK(7,0,0))
     /* gcc bug: https://gcc.gnu.org/legacy-ml/gcc-patches/2018-01/msg01962.html */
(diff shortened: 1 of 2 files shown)


simde/x86/avx512/abs.h
```diff
@@ -524,7 +524,7 @@ simde_mm512_mask_abs_ps(simde__m512 src, simde__mmask16 k, simde__m512 v2) {
 SIMDE_FUNCTION_ATTRIBUTES
 simde__m512d
 simde_mm512_abs_pd(simde__m512d v2) {
-  #if defined(SIMDE_X86_AVX512F_NATIVE) && (!defined(HEDLEY_GCC_VERSION) || HEDLEY_GCC_VERSION_CHECK(8,3,0))
+  #if defined(SIMDE_X86_AVX512F_NATIVE) && (!defined(HEDLEY_GCC_VERSION) || (HEDLEY_GCC_VERSION_CHECK(7,4,0) && !HEDLEY_GCC_VERSION_CHECK(8,0,0)) || HEDLEY_GCC_VERSION_CHECK(8,3,0))
```
mr-c, line 527: Can you move this logic to `simde/simde-common.h` by defining a `SIMDE_BUG_GCC_87467` there?

See https://github.com/simd-everywhere/simde/blob/fdbd6b2616e8ec2509c3ad7087c2e4d36c0a5f5d/simde/simde-common.h#L973-L976 and onwards

mr-c (approved):

simde/x86/avx512/abs.h
thomas-schlichter: Ah, I see. I will do so. Thanks!

mr-c, line 527:
```suggestion
  #if defined(SIMDE_X86_AVX512F_NATIVE) && !defined(SIMDE_BUG_GCC_87467)
```
And same for the other function

mr-c:
Thank you @thomas-schlichter ; if you know of a project using SIMDe that can be listed publicly, it would be nice to add that to https://github.com/simd-everywhere/simde/wiki/Projects-Using-SIMDe

thomas-schlichter:
Yes, OpenAirInterface (https://gitlab.eurecom.fr/oai/openairinterface5g/) is using it.  It is quite a big project and we are just one of many contributors, so I don't know if I would have the right to add it there myself... ;-)

mr-c:
> Yes, OpenAirInterface (https://gitlab.eurecom.fr/oai/openairinterface5g/) is using it. It is quite a big project and we are just one of many contributors, so I don't know if I would have the right to add it there myself... ;-)

That is really cool! I've gone and added it to the list myself.

PR #1121, opened 2024-01-08
apply arm64 windows workaround only on older version msvc

Changqing-JING:
The arm_neo.h bug of msvc is fixed. So that the workaround is not needed on new msvc

### DIFF
--- a/simde/arm/neon/sqadd.h
+++ b/simde/arm/neon/sqadd.h
@@ -32,7 +32,8 @@
 
 // Workaround on ARM64 windows due to windows SDK bug
 // https://developercommunity.visualstudio.com/t/In-arm64_neonh-vsqaddb_u8-vsqaddh_u16/10271747?sort=newest
-#if (defined _MSC_VER) && (defined SIMDE_ARM_NEON_A64V8_NATIVE)
+#if (defined _MSC_VER) && (defined SIMDE_ARM_NEON_A64V8_NATIVE) && (_MSC_VER < 1938)
+#pragma message ("Due to msvc bug, current version of msvc is supported by workaround. Recommend to update msvc")
 #undef vsqaddb_u8
 #define vsqaddb_u8(src1, src2) neon_usqadds8(__uint8ToN8_v(src1), __int8ToN8_v(src2)).n8_u8[0]
 #undef vsqaddh_u16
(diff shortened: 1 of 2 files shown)


simde/arm/neon/sqadd.h
```diff
@@ -30,17 +30,9 @@
...
-#define vsqaddb_u8(src1, src2) neon_usqadds8(__uint8ToN8_v(src1), __int8ToN8_v(src2)).n8_u8[0]
-#undef vsqaddh_u16
-#define vsqaddh_u16(src1, src2) neon_usqadds16(__uint16ToN16_v(src1), __int16ToN16_v(src2)).n16_u16[0]
-#undef vsqadds_u32
-#define vsqadds_u32(src1, src2) _CopyUInt32FromFloat(neon_usqadds32(_CopyFloatFromUInt32(src1), _CopyFloatFromInt32(src2)))
-#undef vsqaddd_u64
-#define vsqaddd_u64(src1, src2) neon_usqadds64(__uint64ToN64_v(src1), __int64ToN64_v(src2)).n64_u64[0]
+#if (defined _MSC_VER) && (_MSC_VER < 1938)
+#pragma message ("_MSC_VER value is: " _CRT_STRINGIZE(_MSC_VER))
+#error "due to the bug of MSVC, this current version of visual studio is not supported. Please update. See https://developercommunity.visualstudio.com/t/In-arm64_neonh-vsqaddb_u8-vsqaddh_u16/10271747?sort=newest"
```
mr-c, line 35: Thanks for this PR. Can we keep the workaround for old versions of MSVC and issue a warning?

mr-c (changes requested):

simde/arm/neon/sqadd.h
Changqing-JING: Sure, changed as your idea

Changqing-JING:
@mr-c why can emscripten test fail on this PR? This PR should not be related to emscripten

mr-c:
> @mr-c why can emscripten test fail on this PR? This PR should not be related to emscripten

The emscripten tests use the latest "tip-of-tree" build of emcc/clang to catch issues early. So it wasn't your change, but their change. Next we need to see if the stable emsdk version also fails, and if not, to switch to that from the "tot" version. Simultaneously we should file an issue with the emscripten people

mr-c:
This might be the cause, so hopefully we can try again when the fix is released

https://github.com/emscripten-core/emscripten/issues/21026

mr-c (approved):

PR #1123, opened 2024-01-08
correction of simde_mm256_sign_epi16().

Proudsalsa:
Hi,

There is an error in the implementation of simde_mm256_sign_epi16(). The implementation does not handle the "==0" case of the avx2 intrinsic. I didn't understand the entire test workflow but i would guess that the error is not detected because the test vector in test_simde_mm256_sign_epi16() doesn't contain any 0s.

Feel free to edit the pull request/code.

This is the Pseudo code of the _mm256_sign_epi16 intrinsic from the Intel Intrinsics guide:

```
FOR j := 0 to 15
	i := j*16
	IF b[i+15:i] < 0
		dst[i+15:i] := -(a[i+15:i])
	ELSE IF b[i+15:i] == 0
		dst[i+15:i] := 0
	ELSE
		dst[i+15:i] := a[i+15:i]
	FI
ENDFOR
dst[MAX:256] := 0
```

### DIFF
--- a/simde/x86/avx2.h
+++ b/simde/x86/avx2.h
@@ -4116,7 +4116,7 @@ simde_mm256_sign_epi16 (simde__m256i a, simde__m256i b) {
 
     SIMDE_VECTORIZE
     for (size_t i = 0 ; i < (sizeof(r_.i16) / sizeof(r_.i16[0])) ; i++) {
-      r_.i16[i] = (b_.i16[i] < INT32_C(0)) ? -a_.i16[i] : a_.i16[i];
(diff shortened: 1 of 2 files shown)


mr-c:
Thank you @Proudsalsa ; can you add additional test cases so we don't make these mistakes again?

Proudsalsa:
i added the testcases by adding an additional element to the test vectors. I hope this is fine.

I ran the tests as described in the CONTRIBUTING.md and checked if they detect an error when run with the old implementation as well.

mr-c (approved):
Thank you @Proudsalsa !

PR #1131, opened 2024-02-17
Update simde-detect-clang.h for clang 13 detection

Coeur:
https://releases.llvm.org/13.0.0/tools/clang/docs/DiagnosticsReference.html#waix-compat

seems to have been incorrectly written in 7bc774fc7087748b8abf2bd6586d2d87df72fe8b

verified working as expected for clang 13 with godbolt: https://godbolt.org/z/cr1YhKv6b


mr-c (approved):
Thank you!

PR #1132, opened 2024-02-17
Update simde-detect-clang.h for clang 17 detection

Coeur:
Verified with godbolt:

- when using clang 17, it's 170000: https://godbolt.org/z/MTzb8va48
- when using clang 16, it's 160000: https://godbolt.org/z/8avxvq7rj


mr-c (approved):
Thank you @Coeur !

PR #1152, opened 2024-03-21
update list of fully implemented instruction sets

mr-c:


mr-c:
Good catch, I'll split out CRC32 separately since our SSE4.2 support is incomplete. 

And https://github.com/simd-everywhere/implementation-status/blob/main/x86.md#vpclmulqdq is implemented, so I'll add it to the list

Happy to have new issues for AVX512-VAES, SSE4a, FMA4 and others, sure. Please double check https://github.com/simd-everywhere/implementation-status/blob/main/x86.md in case it is already done, but not documented.

If you or your collaborators are interested in implementing some of these, we can have a video call where I can walk you through adding a new intrinsic implementation.

mr-c:
> * SSE Implemented: 239 of 241 (99.17%), but listed on the main page as complete
> * AVX Implemented: 296 of 298 (99.33%), but listed on the main page as complete

See https://github.com/simd-everywhere/simde?tab=readme-ov-file#caveats

Please create issues or PRs for the others, thanks!

PR #1172, opened 2024-04-12
arm: enable more intrinsic function for armv7

zengdage:
Hi, I run the arm-neon-tests (https://github.com/christophe-lyon/arm-neon-tests) for simde neon2rvv, and have set the flags `-DSIMDE_RISCV_V_NATIVE -DSIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES` to compiler. But I get some compiler reported errors as follow, it means that vld4_s6 is not supported  in simde neon2rvv.
```
ref_vldX.c:200:3: error: call to undeclared function 'vld4_s64'; ISO C99 and later do not support implicit function declarations 
```

As I know，vld4_s64 and the other intrinsic function which are mentioned in my patch are supported in armv7 platform.  But simde disable them in armv7 platform.  So I upload the patch to enable those intrinsic function in armv7 platform. Please help to review, thanks.

### DIFF
--- a/simde/arm/neon/ld2_lane.h
+++ b/simde/arm/neon/ld2_lane.h
@@ -285,10 +285,10 @@ simde_int16x8x2_t simde_vld2q_lane_s16(int16_t const ptr[HEDLEY_ARRAY_PARAM(2)],
   }
   return r;
 }
-#if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+#if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
   #define simde_vld2q_lane_s16(ptr, src, lane) vld2q_lane_s16(ptr, src, lane)
 #endif
-#if defined(SIMDE_ARM_NEON_A64V8_ENABLE_NATIVE_ALIASES)
+#if defined(SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES)
   #undef vld2q_lane_s16
   #define vld2q_lane_s16(ptr, src, lane) simde_vld2q_lane_s16((ptr), (src), (lane))
 #endif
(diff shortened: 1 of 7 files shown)


simde/arm/neon/ld2_lane.h
```diff
@@ -285,10 +285,10 @@ simde_int16x8x2_t simde_vld2q_lane_s16(int16_t const ptr[HEDLEY_ARRAY_PARAM(2)],
   }
   return r;
 }
-#if defined(SIMDE_ARM_NEON_A64V8_NATIVE)
+#if defined(SIMDE_ARM_NEON_A64V8_NATIVE) || defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```
mr-c, line 288:
```suggestion
#if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
```

If `SIMDE_ARM_NEON_A64V8_NATIVE` is defined, then `SIMDE_ARM_NEON_A32V8_NATIVE` is automaticall defined

https://github.com/simd-everywhere/simde/blob/fee149ae76155f05fbae921884afe68260f2ff83/simde/simde-features.h#L347-L349

and if `SIMDE_ARM_NEON_A32V8_NATIVE` is defined, then `SIMDE_ARM_NEON_A32V7_NATIVE` is automatically defined

https://github.com/simd-everywhere/simde/blob/fee149ae76155f05fbae921884afe68260f2ff83/simde/simde-features.h#L359-L361

So we can simplify the check here and elsewhere in your PR; likewise for `SIMDE_ARM_NEON_A32V7_ENABLE_NATIVE_ALIASES`

mr-c (commented):
Welcome @zengdage and thank you for running that test and for this contribution. I have one suggestion that needs addressing before merging.

simde/arm/neon/ld2_lane.h
mr-c: correct, no need for the `||` (or) , just go for the lowest ARM version requirement

mr-c (approved):

PR #1173, opened 2024-04-16
arm: fix some neon2rvv intrinsic function error

zengdage:
1. For vqdmlal_s16/s32, the doubling result maybe overflow, so need to use vqaddq_s16/32 to saturate it. As the same with vqdmlsl_s16/32.

2. The vrdmulh family function need to use vqadd saturating function to avoid the doubling result overflow. But now I just use rvv intrinsic function to replace the RISCV_V_NATIVE implementation.

3. The result of vrshl family function need to keep the sign bit of the origin data. If` a > 0 && b < 0,` the result of `(a + (1 << (-b - 1)))` maybe overflow into a negative value. And in gcc/clang, >> means the arithmetic shift left, so it will get the incorrect sign bit whithout unsigned extend value.

### DIFF
--- a/test/arm/neon/qrdmulh_n.c
+++ b/test/arm/neon/qrdmulh_n.c
@@ -35,6 +35,9 @@ test_simde_vqrdmulh_n_s16 (SIMDE_MUNIT_TEST_ARGS) {
     { { -INT16_C( 14394),  INT16_C( 28773),  INT16_C( 30122), -INT16_C(   574) },
       -INT16_C( 10708),
       {  INT16_C(  4704), -INT16_C(  9403), -INT16_C(  9843),  INT16_C(   188) } },
+    { {        INT16_MIN,        INT16_MIN,        INT16_MIN,        INT16_MIN },
+             INT16_MIN,
+      {        INT16_MAX,        INT16_MAX,        INT16_MAX,        INT16_MAX } },
 
   };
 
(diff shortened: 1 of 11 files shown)


mr-c (changes requested):
Thank you @zengdage . In addition to making changes so the CI passes, this also needs additional test cases that demonstrate why the changes were needed.

test/arm/neon/qrdmulh_n.c
```diff
@@ -212,6 +221,9 @@ test_simde_vqrdmulhq_n_s32 (SIMDE_MUNIT_TEST_ARGS) {
     { { -INT32_C(   219487289),  INT32_C(  1420994589),  INT32_C(   889110344), -INT32_C(  2103115347) },
        INT32_C(  1735639961),
       { -INT32_C(   177394091),  INT32_C(  1148476728),  INT32_C(   718597063), -INT32_C(  1699780598) } },
+    { {              INT32_MIN,              INT32_MIN,              INT32_MIN,              INT32_MIN },
+      {              INT32_MIN,              INT32_MIN,              INT32_MIN,              INT32_MIN },
+      {              INT32_MIN,              INT32_MIN,              INT32_MIN,              INT32_MIN } },
```
mr-c, line 227:
```suggestion
    { {              INT32_MIN,              INT32_MIN,              INT32_MIN,              INT32_MIN },
       INT32_MIN,
      {              INT32_MIN,              INT32_MIN,              INT32_MIN,              INT32_MIN } },

```

zengdage: I have changed the code. But how can I build and run the test cases  in my personal computer ? so I don't have to submit to a CI test every time.

howjmay:
@zengdage would you mind helping me to fix the same errors in neon2rvv too? Or I can do it?

howjmay:
That helps a lot thanks!

test/arm/neon/qrdmulh_n.c
mr-c: > But how can I build and run the test cases in my personal computer ? so I don't have to submit to a CI test every time.

Sorry for the late reply; See https://github.com/simd-everywhere/simde/tree/master/docker#simde-development-container for how to test using multiple architectures on a single system.

mr-c (approved):

mr-c:
Thank you @zengdage !

PR #1175, opened 2024-04-25
Add vhadd/vabd for rvv

zengdage:


mr-c (approved):
Thank you @zengdage !

PR #1179, opened 2024-05-06
neon: avoid warnings when "__ARM_NEON_FP" is not defined.

clopez:
WebKit project recently imported simde 0.8.2
Since then the build for Aarch64 targeting RPi4 boards has started giving lot of warnings related to the simde header:

`  warning: "__ARM_NEON_FP" is not defined, evaluates to 0 [-Wundef]`

Add a check to ensure that `__ARM_NEON_FP` is defined.

Related: https://bugs.webkit.org/show_bug.cgi?id=273789


mr-c:
Thank you @clopez and thanks for the nice news about webkit using SIMDe!

PR #1183, opened 2024-05-17
arm: improve performance in vqadd and vmvn in risc-v

zengdage:
use risc-v vector instrinsic to implement vqadd_xx and vmvn_xxx.


mr-c (approved):
Thank you @zengdage !

PR #1185, opened 2024-05-22
x86: Apply half tabular method in `_mm_crc32` family

Cuda-Chen:
Apply half tabular method in `_mm_crc32` family


mr-c (approved):
Thank you!

PR #1190, opened 2024-07-02
x86: remove redundant _mm_add_pd translation for WASM

wrv:
This PR removes a duplicated translation for WASM for the SSE2 `_mm_add_pd` intrinsic.


mr-c:
Thanks!

PR #1206, opened 2024-08-19
wasm NEON and AltiVec: add u16x8 and u8x16 avgr translations

wrv:
This PR provides WASM to NEON translations from `wasm_u16x8_avgr` to `vrhaddq_u16` and `wasm_u8x16_avgr` to `vrhaddq_u8`. 

I also tested these locally on a Mac Mini M1. 

Something worth noting here is that the AltiVec translation may also work, but I haven't tested it on a physical device.
Here is what that change looks like: https://github.com/wrv/simde/commit/ff4e85ac47e538aff820514ff728ca3fa6382a9a
And the GH Actions result on that change: https://github.com/wrv/simde/actions/runs/10461140123  

If that looks alright, I can add it to this PR.


mr-c (approved):
Thanks! Feel free to add the altivec commit as well

wrv:
Added

Issue #1211, opened 2024-08-30
arm-gnu-toolchain-11.3.rel1-x86_64-aarch64-none-linux-gnu  build failed


Qingwu-Li:
The issue was introduced from commit d08d67cf , which is gone after reverting this commit.
The issue was found with `arm-gnu-toolchain-11.3.rel1-x86_64-aarch64-none-linux-gnu`.  
the same issue was found with my yocto-generated arm64 gcc version 11.4.
but there is no issue with `arm-gnu-toolchain-12.3.rel1-x86_64-aarch64-none-linux-gnu`

Here is the detail
cross file
```
[binaries]
c = 'aarch64-none-linux-gnu-gcc'
cpp = 'aarch64-none-linux-gnu-g++'
ar = 'aarch64-none-linux-gnu-ar'
strip = 'aarch64-none-linux-gnu-strip'
objcopy = 'aarch64-none-linux-gnu-objcopy'
ld = 'aarch64-none-linux-gnu-ld'


[properties]
c_args = ['-march=armv8-a+simd', '-Wextra', '-Werror']
cpp_args = ['-march=armv8-a+simd', '-Wextra', '-Werror']

[host_machine]
system = 'linux'
cpu_family = 'arm'
cpu = 'armv8-a'
endian = 'little'
```
logs
```
aarch64-none-linux-gnu-gcc -Itest/arm/neon/cadd_rot270-emul-c.p -Itest/arm/neon -I../test/arm/neon -I. -I.. -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -g -march=armv8-a+simd -Wextra -Werror -fopenmp-simd -DSIMDE_CONSTRAINED_COMPILATION -DSIMDE_ENABLE_OPENMP -Wno-psabi -DSIMDE_TEST_BARE -DSIMDE_NO_NATIVE -MD -MQ test/arm/neon/cadd_rot270-emul-c.p/cadd_rot270.c.o -MF test/arm/neon/cadd_rot270-emul-c.p/cadd_rot270.c.o.d -o test/arm/neon/cadd_rot270-emul-c.p/cadd_rot270.c.o -c ../test/arm/neon/cadd_rot270.c
In file included from ../test/arm/neon/../../test.h:5,
                 from ../test/arm/neon/test-neon.h:4,
                 from ../test/arm/neon/cadd_rot270.c:3:
[...]

PR #1216, opened 2024-09-09
ARM NEON: fix arm64 gcc11 build excess elements in vector failure

Qingwu-Li:
Resolved a build failure on ARM64 with GCC 11 caused by excess elements in a vector initializer. The issue stemmed from the `__builtin_shuffle` function, where the number of elements in the argument vector(s) and the mask vector must match.

For more details, refer to issue #1211.

### DIFF
--- a/simde/arm/neon/cadd_rot270.h
+++ b/simde/arm/neon/cadd_rot270.h
@@ -54,7 +54,7 @@ simde_float16x4_t simde_vcadd_rot270_f16(simde_float16x4_t a, simde_float16x4_t 
       r_.sv64 = __riscv_vfadd_vv_f16m1(op1, a_.sv64, 4);
     #elif defined(SIMDE_SHUFFLE_VECTOR_) && !defined(SIMDE_BUG_GCC_100760) &&                                                     \
         ((SIMDE_FLOAT16_API == SIMDE_FLOAT16_API_FP16) || (SIMDE_FLOAT16_API == SIMDE_FLOAT16_API_FLOAT16))
-      b_.values = SIMDE_SHUFFLE_VECTOR_(16, 4, -b_.values, b_.values, 5, 0, 7, 2);
+      b_.values = SIMDE_SHUFFLE_VECTOR_(16, 8, -b_.values, b_.values, 5, 0, 7, 2);
       r_.values = b_.values + a_.values;
     #else
       SIMDE_VECTORIZE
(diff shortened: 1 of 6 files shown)


mr-c (changes requested):
Thank you @Qingwu-Li ; can you address the test failures? https://github.com/simd-everywhere/simde/actions/runs/10774988233/job/29878397301?pr=1216#step:10:1023

mr-c:
I can confirm this PR does fix some errors: https://github.com/simd-everywhere/simde/actions/runs/10775769592/job/29880998011?pr=1216#step:9:1271 (log of a gcc-11 aarch64 build without these fixes)

Qingwu-Li:
> Thank you @Qingwu-Li ; can you address the test failures? https://github.com/simd-everywhere/simde/actions/runs/10774988233/job/29878397301?pr=1216#step:10:1023

This is another issue related to gcc 11. While I'm new to SIMD and not fully confident in the fix, it provides a temporary solution to the problem.

mr-c:
> > Thank you @Qingwu-Li ; can you address the test failures? https://github.com/simd-everywhere/simde/actions/runs/10774988233/job/29878397301?pr=1216#step:10:1023
> 
> This is another issue related to gcc 11. While I'm new to SIMD and not fully confident in the fix, it provides a temporary solution to the problem.

Thanks for the fix! Is there a bug about ` vst1_*_x4` in the GCC bug tracker?

mr-c:
Yay, it worked! It would be nice to update the reference to include a specific GCC bug number before merging

Qingwu-Li:
> Yay, it worked! It would be nice to update the reference to include a specific GCC bug number before merging

Didn't found the related bug number , maybe fixed in this commit in releases/gcc-11.5.0，but not verifyed. 
https://gcc.gnu.org/git/?p=gcc.git;a=log;h=daee0409d195d346562e423da783d5d1cf8ea175

mr-c:
> > Yay, it worked! It would be nice to update the reference to include a specific GCC bug number before merging
> 
> Didn't found the related bug number , maybe fixed in this commit in releases/gcc-11.5.0，but not verifyed. https://gcc.gnu.org/git/?p=gcc.git;a=log;h=daee0409d195d346562e423da783d5d1cf8ea175

Okay, that seems to come from https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114521

mr-c (approved):

mr-c:
Thank you @Qingwu-Li !

PR #1223, opened 2024-09-18
x86: move definition of 'value' to correct branch in _mm_loadl_epi64

K-os:


mr-c (approved):
Good catch! Thank you @K-os

PR #1224, opened 2024-09-20
loongarch float16: use a portable version to avoid compilation errors

XiWeiGu:
Fixed the compilation issue on the Loongarch platform with clang:
` ../test/x86/avx512/../../../simde/simde-f16.h:100:11: error: _Float16 is not supported on this target.`


mr-c (approved):
That the new CI job catches the error is confirmed at https://github.com/simd-everywhere/simde/actions/runs/10972598018/job/30469053422#step:9:11

Thank you for the fix @XiWeiGu !

PR #1232, opened 2024-11-15
Fix clang type redef error

bd-jahn:
Clang has a `typedef unsigned long long __mask64` defined in `avx512bwintrin.h` ([link](https://clang.llvm.org/doxygen/avx512bwintrin_8h.html#a910ccf7d2141c4a3b8675e2480cc33ff)) and  causes `typedef redefinition with different types ('uint64_t' (aka 'unsigned long') vs 'unsigned long long')` error. 
This PR adds clang case in to the logic to fix this error.

See this for more detail: https://github.com/simd-everywhere/simde/commit/e8390a3bfb3bd9e5119a9883fb29b07169b2fe4d#r149134782


bd-jahn:
Thanks for pushing this! Please let me know if there is anything I can help with :+1:

PR #1238, opened 2024-11-28
should use INT64_C

jinboson:
I think using INT64_C may be better or please let me know if I missed something ?:)


mr-c:
Thank you @jinboson !

PR #1247, opened 2024-12-12
x86 sse2/avx2: rewrite loongarch immediate operand shift instruction

HecaiYuan:
The x86 instructions like _mm_srli_epi8() can accept variable or immediate operand as the second parm, however, the corresponding loongarch instructions like __lsx_vsrli_b only accept immediate operand as the second parm, so we need to rewirite them to avoid compilation error.

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -6158,12 +6158,12 @@ simde_mm_sll_epi16 (simde__m128i a, simde__m128i count) {
     if (count_.u64[0] > 15)
       return simde_mm_setzero_si128();
 
-    #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
+    #if defined(SIMDE_LOONGARCH_LSX_NATIVE)
+      r_.lsx_i64 = __lsx_vsll_h(a_.lsx_i64, __lsx_vreplgr2vr_h(count_.u64[0]));
+    #elif defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR)
       r_.u16 = (a_.u16 << count_.u64[0]);
     #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
       r_.neon_u16 = vshlq_u16(a_.neon_u16, vdupq_n_s16(HEDLEY_STATIC_CAST(int16_t, count_.u64[0])));
-    #elif defined(SIMDE_LOONGARCH_LSX_NATIVE)
-      r_.lsx_i64 = __lsx_vslli_h(a_.lsx_i64, count_.u64[0]);
     #elif defined(SIMDE_WASM_SIMD128_NATIVE)
       r_.wasm_v128 = ((wasm_i64x2_extract_lane(count_.wasm_v128, 0) < 16) ? wasm_i16x8_shl(a_.wasm_v128, HEDLEY_STATIC_CAST(int32_t, wasm_i64x2_extract_lane(count_.wasm_v128, 0))) : wasm_i16x8_const(0,0,0,0,0,0,0,0));
     #else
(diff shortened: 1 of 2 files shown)


jinboson:
Looks good to me.

jinboson:
> @jinboson @HecaiYuan Thanks. Do you know why the CI didn't catch this already?

Hi, @mr-c . In the CI test , for example, when running test for simde_mm_slli_epi16(), the second shift parm passed is immediate value, like below:

> test/x86/sse2.c:7866:test_simde_mm_slli_epi16(SIMDE_MUNIT_TEST_ARGS) {
> test/x86/sse2.c:7908:    simde__m128i r = simde_mm_slli_epi16(test_vec[i].a, 3);
> test/x86/sse2.c:7911:    r = simde_mm_slli_epi16(test_vec[i].a, 0);
> test/x86/sse2.c:7914:    r = simde_mm_slli_epi16(test_vec[i].a, 32);
> test/x86/sse2.c:7917:    r = simde_mm_slli_epi16(test_vec[i].a, 33);

however, we found that in some projects(like ,aom) that user would pass varibales to the second parm, like below:

> aom_dsp/x86/highbd_loopfilter_sse2.c:35:  *blt = _mm_slli_epi16(x, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:38:  *lt = _mm_slli_epi16(x, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:41:  *thr = _mm_slli_epi16(x, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:59:  *blt_out = _mm_slli_epi16(x0, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:64:  *lt_out = _mm_slli_epi16(x0, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:69:  *thr_out = _mm_slli_epi16(x0, shift);
> aom_dsp/x86/highbd_loopfilter_sse2.c:198:  th = _mm_slli_epi16(th, bd - 8);
> aom_dsp/x86/highbd_loopfilter_sse2.c:208:  th = _mm_slli_epi16(th, bd - 8);

In the case loongarch implementation for simde_mm_slli_epi16() would emit error at compile time before the PR.

BTW, I just find another porblem that we need to adujust the implementation location of loongarch for simde_mm_sll_epi16 and simde_mm_sll_epi32 , otherwise the loongarch implementaion will go into the #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR) codepath. We will update it as soon as possible. Thanks.

simde/x86/sse2.h
```diff
@@ -6163,7 +6163,7 @@ simde_mm_sll_epi16 (simde__m128i a, simde__m128i count) {
     #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
       r_.neon_u16 = vshlq_u16(a_.neon_u16, vdupq_n_s16(HEDLEY_STATIC_CAST(int16_t, count_.u64[0])));
     #elif defined(SIMDE_LOONGARCH_LSX_NATIVE)
-      r_.lsx_i64 = __lsx_vslli_h(a_.lsx_i64, count_.u64[0]);
```
jinboson, line 6166: Move this to the line 6160, Othewise the codepath will step into #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR) branch.

simde/x86/sse2.h
```diff
@@ -6163,7 +6163,7 @@ simde_mm_sll_epi16 (simde__m128i a, simde__m128i count) {
...
+      r_.lsx_i64 = __lsx_vsll_h(a_.lsx_i64, __lsx_vreplgr2vr_h(count_.u64[0]));
```
jinboson, line 6166: Move the implementation to the line 6160, Othewise the codepath will step into #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR) branch.

simde/x86/sse2.h
```diff
@@ -6199,7 +6199,7 @@ simde_mm_sll_epi32 (simde__m128i a, simde__m128i count) {
     #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
       r_.neon_u32 = vshlq_u32(a_.neon_u32, vdupq_n_s32(HEDLEY_STATIC_CAST(int32_t, count_.u64[0])));
     #elif defined(SIMDE_LOONGARCH_LSX_NATIVE)
-      r_.lsx_i64 = __lsx_vslli_w(a_.lsx_i64, count_.u64[0]);
+      r_.lsx_i64 = __lsx_vsll_w(a_.lsx_i64, __lsx_vreplgr2vr_w(count_.u64[0]));
```
jinboson, line 6202: Move the implementation to the line 6196, Othewise the codepath will step into #if defined(SIMDE_VECTOR_SUBSCRIPT_SCALAR) branch.

rosbif:
I haven't looked at SIMDe recently but I feel that I should chip in on this.
The initial premise is false: the Intel `s[rl][al]i` intrinsics should only accept an immediate constant for the shift count.
This is clear from the Intel Intrinsics Guide and the 'i' in the intrinsic stands for immediate.
It is even clearer from the assembler instruction which should be generated.
The problem is that unfortunately some compilers accept a variable.
This presents a dilemma for SIMDe: a variable shift count should be refused but in this case users will complain that their compiler accepts a variable bur SIMDe doesn't.

jinboson:
> dilemma

That's really a dilemma, thank you for setting me straight.

jinboson:
> I haven't looked at SIMDe recently but I feel that I should chip in on this. The initial premise is false: the Intel `s[rl][al]i` intrinsics should only accept an immediate constant for the shift count. This is clear from the Intel Intrinsics Guide and the 'i' in the intrinsic stands for immediate. It is even clearer from the assembler instruction which should be generated. The problem is that unfortunately some compilers accept a variable. This presents a dilemma for SIMDe: a variable shift count should be refused but in this case users will complain that their compiler accepts a variable but SIMDe doesn't.

That's really a dilemma, thank you for setting me straight.

jinboson:
Hi, @mr-c . This is ready for review, then is there anything else needed to move this forward ?

HecaiYuan:
Hi, @mr-c . Regarding this issue, do you have any good suggestions? What should we do next?

yinshiyou:
I am discussing with the compiler team about allowing the slli instruction to accept variables as parameters, as well as allowing  immediate values expanding the range(b 8, h 16, w 32) .
Currently, it is still recommended to merge this PR, at least  it eliminates the need  to check if the immediate value exceeds the limit.

jinboson:
This can be closed now, see #1263 for details. Thank you every one.

PR #1255, opened 2025-01-03
Fix SQDMULL implementation for 32-bit inputs.

Syonyk:
The non-vectorized SQDMULL implementation was wrong for 32-bit inputs.  It incorrectly checked one of the operands to see if the value would overflow before doubling it, not the result of the initial multiplication.  It now matches the 16-bit operand version in flow (which was correct), and also matches hardware.  A test has been added for the scalar form of the function, testing a range of values that will saturate when multiplied and doubled.  This set of test vectors was produced on an ARMv9 machine (Google Cloud box), failed tests on x86 with the existing code, and passes with the modified code.

I have not tested the vectorized versions of this with WASM and RISC-V to see if they're correct, only the fallback code.


mr-c (approved):

mr-c:
Thank you @Syonyk !

PR #1256, opened 2025-01-03
Fix UQSHL to match hardware.  Add extensive test vectors.

Syonyk:
UQSHL was wrong in a variety of ways.  Most importantly, it did not trim off the low byte for the shift as per the instruction spec: "... shifts the element by a value from the least significant byte of the corresponding element of the second source register..." This was not detected in the tests, as the test vectors were limited to small shift values that did not demonstrate the fault.

Behavior is fixed to be correct, per hardware.  Test vectors are dramatically expanded, using both the previous small-range values as well as additional full-range values.  Vectors generated on an ARMv9 system (Google Compute Engine C4A).


mr-c:
These are great, thank you @Syonyk !

PR #1258, opened 2025-01-03
Remove incorrect qrdmulh SSE code.

Syonyk:
The only reason the SSE code for qrdmulh passed is because the edge cases were not included in the tests unless SSE was disabled.  The `INT16_MIN * INT16_MIN` case ought result in `INT16_MAX` - and it does, in the fallback code.  It does not, in the SSE code, which is what will typically be used on x86 hardware.  Saturating code not handling edge cases is simply wrong.

Unfortunately, my SSE skills are not sufficient to sort out how to fix this problem.  I'd also be fine commenting out the code with a note about the problem, but I prefer correct results in all cases to "commonly correct results except in the case where the result matters."

Tests are modified to remove the conditional test.  The final test cases passes cleanly on ARMv9 hardware, but not on x86 with SSE.  It passes in the fallback code path.

### DIFF
--- a/simde/arm/neon/qrdmulh.h
+++ b/simde/arm/neon/qrdmulh.h
@@ -128,29 +128,8 @@ simde_vqrdmulhq_s16(simde_int16x8_t a, simde_int16x8_t b) {
       a_ = simde_int16x8_to_private(a),
       b_ = simde_int16x8_to_private(b);
 
-    /* https://github.com/WebAssembly/simd/pull/365 */
     #if defined(SIMDE_ARM_NEON_A32V7_NATIVE)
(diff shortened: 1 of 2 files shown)


mr-c:
Checking the git blame, I'm pinging @nemequ about ccacf9428b4b28ad6a678da167ea7d6fc628daac and  @zengdage
 about 2a548e591c74fb26b88c5c25dd295f58dee22069 in the context of this PR

Syonyk:
Sounds good - if they're able to fix it properly, to cover all the test cases, I have no objections.  Just so long as all cases are correct.

mr-c (approved):

PR #1259, opened 2025-01-03
Fix vqshlud_n_s64 implementation to be 64-bit.

Syonyk:
The scalar vqshlud_n_s64 function was wrong, and there was no test case for it (or the scalar 32 bit form).  Add test cases for the scalar functions (test vectors generated on GCE ARMv9 system), validate the fix for the 64-bit scalar form.  Tweak test case generation to generate more "valid" test cases (cases that don't simply saturate to 0 or rapidly saturate to INT_MAX for the function size).


mr-c (approved):
Thank you! Please don't push to this branch, I'll be merging it next.

PR #1261, opened 2025-01-06
Change generation of 2^n values for fixed point conversions.

Syonyk:
As demonstrated by test code in https://github.com/simd-everywhere/simde/issues/1260 the behavior of pow() in non-round-to-nearest rounding modes is not exact.  This causes behavior divergence from ARMv8 hardware when not using round-to-nearest.  The updated forms match hardware properly across a range of values.  The tests are not updated to handle rounding modes, as doing this in a cross-platform way is not trivial. However, all existing test vectors pass properly, and in more detailed testing, these changes are closer to hardware.


Syonyk:
Okeday, needs a few tweaks...

mr-c:
> Okeday, needs a few tweaks...

Let me know if you need help dealing with the compiler complaints :-)

Syonyk:
Thanks, I'll poke at this more and see what I can come up with.

Issue #1177, opened 2024-04-30
reimplement the removed FCVTZS/FCVTMS/FCVTPS/FCVTNS family intrinsics 


mr-c:
https://github.com/simd-everywhere/simde/commit/339ffe4ca89bf6698cab0c8420fe813a4e293aec

Please test using real hardware ; see https://github.com/simd-everywhere/simde/issues/1099 for more details

PR #1263, opened 2025-01-13
Fixes for loongarch compilers

jinboson:
Happy New Year, Michael @mr-c. 

We rewrited the PR https://github.com/simd-everywhere/simde/pull/1247 (can be closed now) and seperated it into two patches. We did two things on them.

1）The first thing is to fix a typo for lasx implemention in x86 avx2 that modified simde_mm256_bslli_epi128 to simde_mm256_bsrli_epi128, very sorry for this typo error.
2）The second thing is to work around compiler errors in use-cases that users would pass variables for shift count for slli/srli/srai instrutions, which can not support for loongarch at present.

We hope you could take some time to review this for us, if you have any questions please feel free to let us know, thank you Michael， thank you.

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -6163,7 +6163,7 @@ simde_mm_sll_epi16 (simde__m128i a, simde__m128i count) {
     #elif defined(SIMDE_ARM_NEON_A32V7_NATIVE)
       r_.neon_u16 = vshlq_u16(a_.neon_u16, vdupq_n_s16(HEDLEY_STATIC_CAST(int16_t, count_.u64[0])));
     #elif defined(SIMDE_LOONGARCH_LSX_NATIVE)
-      r_.lsx_i64 = __lsx_vslli_h(a_.lsx_i64, count_.u64[0]);
+      r_.lsx_i64 = __lsx_vsll_h(a_.lsx_i64, __lsx_vreplgr2vr_h(count_.u64[0]));
     #elif defined(SIMDE_WASM_SIMD128_NATIVE)
       r_.wasm_v128 = ((wasm_i64x2_extract_lane(count_.wasm_v128, 0) < 16) ? wasm_i16x8_shl(a_.wasm_v128, HEDLEY_STATIC_CAST(int32_t, wasm_i64x2_extract_lane(count_.wasm_v128, 0))) : wasm_i16x8_const(0,0,0,0,0,0,0,0));
     #else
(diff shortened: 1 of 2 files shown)


simde/x86/sse2.h
```diff
@@ -6590,7 +6590,7 @@ simde_mm_slli_epi16 (simde__m128i a, const int imm8)
   #define simde_mm_slli_epi16(a, imm8) \
     ((imm8 & ~15) ? simde_mm_setzero_si128() : simde__m128i_from_altivec_i16(vec_sl(simde__m128i_to_altivec_i16(a), vec_splat_u16(HEDLEY_STATIC_CAST(unsigned short, imm8)))))
 #elif defined(SIMDE_LOONGARCH_LSX_NATIVE)
-  #define simde_mm_slli_epi16(a, imm8) ((imm8 & ~15) ? simde_mm_setzero_si128() : simde__m128i_from_lsx_i64(__lsx_vslli_h(simde__m128i_to_private(a).lsx_i64, ((imm8) & 15))))
+  #define simde_mm_slli_epi16(a, imm8) ((imm8 & ~15) ? simde_mm_setzero_si128() : simde__m128i_from_lsx_i64(__lsx_vsll_h(simde__m128i_to_private(a).lsx_i64, __lsx_vreplgr2vr_h((imm8) & 15))))
```
mr-c, line 6593: Why are you changing this? For `simde_mm_slli_epi16` we require `imm8` to be a constant value.

See https://github.com/simd-everywhere/simde/pull/1263/commits/1715c161b794c5e4aef960c94647fa4f4ac5ab3d#diff-d5773576112e65731ce8f01f5beee4f665e6fcffd1017ee76c583fa0988a40b4R6554-R6555

Same for `simde_mm_slli_epi32`, `simde_mm_slli_epi64`, `simde_mm_srli_epi16`, `simde_mm_srli_epi32` and `simde_mm_srli_epi64`

mr-c (changes requested):

simde/x86/sse2.h
mr-c: Wacky. Sure, it doesn't bother me either way

mr-c (changes requested):
if the `LASX_NATIVE` versions don't require the use of immediate values, then please move those optimized implementaions into the function definitions instead of using macros.

mr-c:
@jinboson if SIMDe is important to the LOONGARCH ecosystem, it would be helpful if I had some assistance with https://github.com/simd-everywhere/simde/issues/1177 and similar ; I'm the only maintainer these days..

simde/x86/sse2.h
jinboson:
```
#include <stdio.h>
#include "/data/home/Jin/data/community/simde-loong64/simde/x86/sse2.h"

int main()
{
        simde__m128i vec0;
        const int imm8 = 1;     //emulate use-cases that end-users pass variable for shift count.
        simde_mm_slli_epi16(vec0, imm8);
	return 0;
}

```
I built above demo on real loongarch hardware with simde git master, and the compiler output errors:
image
Same for simde_mm_slli_epi32, simde_mm_slli_epi64, simde_mm_srli_epi16, simde_mm_srli_epi32 and simde_mm_srli_epi64

jinboson:
```
#include <stdio.h>
#include "/data/home/Jin/data/community/simde-loong64/simde/x86/sse2.h"

int main()
{
        simde__m128i vec0;
        //const int imm8 = 1;     //emulate use-cases that end-users pass variable for shift count.
        simde_mm_slli_epi16(vec0, 1);
	return 0;
}
```
If user pass immediate values for shift count just like above demo,  it's ok for loongarch compiler.

jinboson: So, at present, we need to load imm8 into vectors and then use sll version instruction instead. If in the feature, the loongarch compiler can accept a variable for shift count for slli instruction then we can solve this dilemma.

jinboson:
> @jinboson if SIMDe is important to the LOONGARCH ecosystem, it would be helpful if I had some assistance with #1177 and similar ; I'm the only maintainer these days..

Understood, thank you @mr-c , I have posted the test result based on the real loongarch hardware and made explanations for you.  Thanks for you reply!

jinboson:
> if the `LASX_NATIVE` versions don't require the use of immediate values, then please move those optimized implementaions into the function definitions instead of using macros.

It depens on how user uses them.  We do have slli version instructions, but you must pass immediate values for shift count when using them.

jinboson:
> move those optimized implementaions into the function definitions instead of using macros. 

Done by the latest update, thank you Michael.

mr-c (approved):

PR #1269, opened 2025-01-28
arm: replace ARMv8 sub-architecture detection with feature detection

mr-c:


mr-c:
@wewe5215 I copied your work from #1270 into this PR so we can get extra testing

PR #1272, opened 2025-01-30
Fix incorrect UQRSHL implementation.

Syonyk:
UQRSHL was incorrect in several ways, and did not match hardware. Shift checks for saturating were not modified from the signed version, leading to incorrect behavior at (untested) edge cases.

Further, for the 32-bit and 64-bit versions, the rounding math was incorrect with a -32 or -64 shift, and did not match hardware.  This has been special cased to do the proper thing at the edge case.  The core issue is overflowing the stock 32-bit or 64-bit types used at the limit.

Tests have been modified with substantially increased test cases to properly exercise these changes.  The new tests should cause failures in the unsigned cases without the modified code, and should pass with the changes.  Additional signed tests were added to handle the edge cases, though no incorrect behavior was observed on those.

Test cases generated on a Google C4a ARMv8.4 machine.


mr-c (approved):
Thank you @Syonyk !

PR #1282, opened 2025-02-23
[Neon][RVV] Fallback to autovec without mrvv-vector-bits flag.

Ruhung:
Fallback to auto vectorization if there is no mvv-vector-bits flag.


mr-c (approved):

mr-c:
Thank you @Ruhung for your many contributions!

PR #1284, opened 2025-02-25
[Neon][RVV] Enable RVV when there are zve64d and zvl128b flags.

Ruhung:
Enable RVV when there are zve64d and zvl128b flags.


mr-c (approved):
Thanks. Can you look into the CI failures?

PR #1285, opened 2025-02-25
[Neon][RVV] Enable RVV segment load/store only when we have __riscv_zvlsseg flag.

Ruhung:
Enable RVV segment load/store only when we have __riscv_zvlsseg flag.

### DIFF
--- a/simde/simde-arch.h
+++ b/simde/simde-arch.h
@@ -550,6 +550,9 @@
 #if defined(__riscv_zvfhmin)
 #  define SIMDE_ARCH_RISCV_ZVFHMIN 1
 #endif
+#if defined(__riscv_zvlsseg)
+#  define SIMDE_ARCH_RISCV_ZVLSSEG 1
+#endif
 
 /* SPARC
    <https://en.wikipedia.org/wiki/SPARC> */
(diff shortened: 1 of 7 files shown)


simde/simde-arch.h
```diff
@@ -550,6 +550,9 @@
 #if defined(__riscv_zvfhmin)
 #  define SIMDE_ARCH_RISCV_ZVFHMIN 1
 #endif
+#if defined(__riscv_zvlsseg)
+#  define SIMDE_ARCH_RISCV_ZVLSSEG 1
+#endif
```
mr-c, line 555:
```suggestion
#if defined(__riscv_zvlsseg) || defined(__riscv_v)
#  define SIMDE_ARCH_RISCV_ZVLSSEG 1
#endif
```

For example, llvm 14+ removed the `__riscv_zvlsse` flag https://github.com/llvm/llvm-project/commit/7a275dc35411b8c3f510166f40c225cd10dc5eec

mr-c:
@Ruhung I presume this is to be merged after https://github.com/simd-everywhere/simde/pull/1284 ?

simde/simde-arch.h
Ruhung: You're right. I applied this.

Ruhung:
> @Ruhung I presume this is to be merged after #1284 ?

It can be merged after #1284.

mr-c (approved):

topperc:
Sorry for the archaelogy here. Zvlsseg is an old extension name from pre-1.0 RVV. It was removed from RVV 1.0 and the last several compiler releases never define `__riscv_zvlsseg`.

PR #1292, opened 2025-04-02
prefer NATIVE implementations over SIMDE_VECTOR_SUBSCRIPT_OPS, remove some broken optimized implementations

Ruhung:
Misuse RVV intrinsics vnot. We didn't find this error cause it always use builtin vector operation in the testing.


mr-c:
@Ruhung Can you suggest a change to our CI setup that would have caught this bug earlier?

PR #1296, opened 2025-06-24
Avoid undefined behaviour with signed integer multiplication

### LINKED ISSUE #1295: Many intrinsics exhibit signed integer overflow UB
For example simde_vmul_laneq_s32:
https://github.com/simd-everywhere/simde/blob/51743e7920b6e867678cb50e9c62effe28f70b33/simde/arm/neon/mul_lane.h#L389

The multiplication here is UB if signed overflow occurs. The ARM documentation does not specify such UB: https://developer.arm.com/architectures/instruction-sets/intrinsics/vmul_laneq_s32

The correct way to perform the multiplication would be to cast to `(uint32_t)` first.

This causes test failures with clang 20.

rathann:
This is the minimal patch that makes tests pass with clang 20.1.x on Fedora rawhide x86_64 and aarch64.

Fixes #1295 .

### DIFF
--- a/simde/arm/neon/mul_lane.h
+++ b/simde/arm/neon/mul_lane.h
@@ -386,7 +386,7 @@ simde_vmul_laneq_s32(simde_int32x2_t a, simde_int32x4_t b, const int lane)
   #else
     SIMDE_VECTORIZE
     for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-      r_.values[i] = (uint32_t)a_.values[i] * (uint32_t)b_.values[lane];
+      r_.values[i] = (int32_t)((uint32_t)a_.values[i] * (uint32_t)b_.values[lane]);
     }
   #endif
 
@@ -772,7 +772,7 @@ simde_vmulq_laneq_s32(simde_int32x4_t a, simde_int32x4_t b, const int lane)
   #else
     SIMDE_VECTORIZE
     for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-      r_.values[i] = (uint32_t)a_.values[i] * (uint32_t)b_.values[lane];
+      r_.values[i] = (int32_t)((uint32_t)a_.values[i] * (uint32_t)b_.values[lane]);
     }
   #endif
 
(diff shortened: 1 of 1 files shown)


simde/arm/neon/mul_lane.h
```diff
@@ -386,7 +386,7 @@ simde_vmul_laneq_s32(simde_int32x2_t a, simde_int32x4_t b, const int lane)
   #else
     SIMDE_VECTORIZE
     for (size_t i = 0 ; i < (sizeof(r_.values) / sizeof(r_.values[0])) ; i++) {
-      r_.values[i] = a_.values[i] * b_.values[lane];
+      r_.values[i] = (uint32_t)a_.values[i] * (uint32_t)b_.values[lane];
```
mr-c, line 389: Thank you, but this generated clang errors:

https://github.com/simd-everywhere/simde/actions/runs/15861220962/job/44718601505?pr=1296#step:10:666
```suggestion
      r_.values[i] = (int32_t)((uint32_t)a_.values[i] * (uint32_t)b_.values[lane]);
```

aqrit:
> Fixes https://github.com/simd-everywhere/simde/issues/1295 .

`simde_vmul_laneq_s32` was for an example.
 The "many intrinsics" statement is key.

related:

>Undefined behavior stemming from signed overflow seems to be everywhere.
At this time, I'd say that SIMDE should never be used without -fwrapv.
_https://github.com/simd-everywhere/simde/issues/901#issuecomment-910662042_

simde/arm/neon/mul_lane.h
```diff
@@ -772,7 +772,7 @@ simde_vmulq_laneq_s32(simde_int32x4_t a, simde_int32x4_t b, const int lane)
...
```
mr-c, line 775: Probably also here as well?

rathann:
It looks like this is now failing when compiled as C++:
```
../test/arm/neon/../../../simde/arm/neon/mul_lane.h:389:32: error: use of old-style cast [-Werror,-Wold-style-cast]
```
What is the recommended way of keeping the code compatible with both C and C++? I'm not very good with C++.

mr-c:
> It looks like this is now failing when compiled as C++:
> 
> ```
> ../test/arm/neon/../../../simde/arm/neon/mul_lane.h:389:32: error: use of old-style cast [-Werror,-Wold-style-cast]
> ```
> 
> What is the recommended way of keeping the code compatible with both C and C++? I'm not very good with C++.

My mistake, sorry

Try one of `HEDLEY_STATIC_CAST` or `HEDLEY_REINTERPRET_CAST` 
https://github.com/simd-everywhere/simde/blob/51743e7920b6e867678cb50e9c62effe28f70b33/simde/hedley.h#L842-L858

mr-c (approved):

PR #1319, opened 2025-09-09
x86 sse2: add powerpc*-darwin support for `simde_mm_pause`

barracuda156:
See: https://github.com/simd-everywhere/simde/issues/1313

### DIFF
--- a/simde/x86/sse2.h
+++ b/simde/x86/sse2.h
@@ -5245,7 +5245,11 @@ simde_mm_pause (void) {
       __asm__ __volatile__("isb\n");
     #endif
   #elif defined(SIMDE_ARCH_POWER)
-    __asm__ __volatile__ ("or 27,27,27" ::: "memory");
+    #if defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


simde/x86/sse2.h
```diff
@@ -5245,7 +5245,11 @@ simde_mm_pause (void) {
       __asm__ __volatile__("isb\n");
     #endif
   #elif defined(SIMDE_ARCH_POWER)
-    __asm__ __volatile__ ("or 27,27,27" ::: "memory");
+    #if defined(__APPLE__)
+      __asm__ __volatile__ ("or r27,r27,r27" ::: "memory");
+    #else
+      __asm__ __volatile__ ("or 27,27,27" ::: "memory");
+    #endif
```
mr-c, line 5252: Good to know! If you are looking for other software that will also need patching this way, look at https://codesearch.debian.net/search?q=%22or+27%2C27%2C27%22&literal=1&perpkg=1

PR #1331, opened 2025-09-19
x86 axv512 storeu: fix mask storeu zeroing out original memory

michael-chuh:
The original codes will overwrite the umasked memory location with zeros, causing data loss.


mr-c (approved):
Thank you @michael-chuh !

I've updated the tests to prove your fixes.

Issue #1337, opened 2025-09-23
simde_mm256_permute2f128_si256 does not map to the native asm instruction when compiling for AVX2


ethomag:
I believe this is a simple type in x86/avx.h.  Around line 5431, it reads:

```C++
#if defined(SIMDE_X86_AVX_NATIVE)
#  define simde_mm256_permute2f128_si128(a, b, imm8) _mm256_permute2f128_si128(a, b, imm8)
#endif
```

I think is should be:

```C++
#if defined(SIMDE_X86_AVX_NATIVE)
#  define simde_mm256_permute2f128_si256(a, b, imm8) _mm256_permute2f128_si256(a, b, imm8)
#endif
```

`_mm256_permute2f128_si128` is not a valid AVX intrinsic

mr-c:
Thank you @ethomag , you are correct. ~~While we test the native aliases, that only works when we have a test case for that aliased function, which is missing for `_mm256_permute2f128_si256` and probably others in the `_mm{256,}_permute` family of AVX intrinsics, whoops!~~

Feel free to send a PR to fix the typo ~~bonus if it adds the missing test~~

mr-c:
Ah, I see why the testing didn't find this; it isn't an error in the native aliases (setting `SIMDE_ENABLE_NATIVE_ALIASES` and then using `_mm256_permute2f128_si256`  to mean `simde_mm256_permute2f128_si256`), but an error of calling the real AVX intrinsic on AVX capable systems

mr-c:
Fixed in #1338

PR #1338, opened 2025-09-23
x86 avx: fix native implementation of `simde_mm256_permute2f128_si256`

ethomag:
Hi @mr-c and thanks for a great library!  

This is to correct the typo that caused: #1337 "simde_mm256_permute2f128_si256 does not map to the native asm instruction when compiling for AVX2"

I am happy to take a shot at the tests at a later stage.

### DIFF
--- a/test/x86/avx.c
+++ b/test/x86/avx.c
@@ -12451,74 +12451,98 @@ test_simde_mm256_permute2f128_si256(SIMDE_MUNIT_TEST_ARGS) {
   const struct {
     const int16_t a[16];
     const int16_t b[16];
-    const int imm8;
     const int16_t r[16];
   } test_vec[] = {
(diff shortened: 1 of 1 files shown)


mr-c:
Thank you @ethomag ! Here's a PR to your PR to fix the appveyor CI failure https://github.com/ethomag/simde/pull/1

ethomag:
Thanks again @mr-c. Unfortunately I don't know where to begin locating the error. It seems to originate in a file called avx.cpp that is generated ? And only for MVSC (which I don't have access to).  Any pointers how to attack this would be appreciated.

mr-c:
> Thanks again @mr-c. Unfortunately I don't know where to begin locating the error. It seems to originate in a file called avx.cpp that is generated ? And only for MVSC (which I don't have access to). Any pointers how to attack this would be appreciated.

@ethomag I fixed it for you already; just click the "merge" button at https://github.com/ethomag/simde/pull/1

PR #1341, opened 2025-09-25
_mm256_permute4x64_epi64: Avoid "may be used uninitialized" warning

ethomag:


ethomag:
I can't see how these CI failures relate to the change I made.

mr-c:
I'm going to wait for them to finish and then I'll retry those two failing runs for you

PR #1345, opened 2025-10-01
Two more cases when g++ reports a warning about 'maybe uninitialized'

ethomag:


ethomag:
Seems the "clang 13" test timed out during installation and the other one did not even start.

mr-c:
Thank you @ethomag !

PR #1356, opened 2025-12-18
x86 avx512: add mask compress and compressstore functions for 128bit

AdrianRiedl:
This pull request adds support for AVX-512 compress and compress-store intrinsics for 128-bit vectors. It introduces portable implementations for these intrinsics for double, float, int32, and int64 types, along with their corresponding mask and zero-masked variants, and sets up native aliasing macros for seamless integration with existing code.

Code style is kept similar to the already existing implementations for m256/512.

### DIFF
--- a/simde/x86/avx512/compress.h
+++ b/simde/x86/avx512/compress.h
@@ -8,6 +8,362 @@ SIMDE_DISABLE_UNWANTED_DIAGNOSTICS
 SIMDE_BEGIN_DECLS_
 
 SIMDE_FUNCTION_ATTRIBUTES
+simde__m128d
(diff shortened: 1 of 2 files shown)


mr-c:
Hello @AdrianRiedl and thank you for this PR! I'll fix the CI issues, they appear unrelated. 

From a brief glance everything looks okay. Did you generate the test vectors on real AVX512VL hardware or use an emulator?

AdrianRiedl:
Hi @mr-c, thanks for taking a look — and especially for dealing with the CI issues.

The test vectors were generated on real AVX512VL hardware (AMD Ryzen 9 7950X 16-Core Processor).

PR #1360, opened 2026-01-07
Fix some SSE on NEON bugs

TheNumbat:
- `_mm_min_pd`/`_mm_max_pd` had incorrect NaN behavior. The implementations of the four min/max functions were all slightly different; they now use the same logic. The NaN bug also applied to other architectures.

- `_mm_round_pd` wasn't doing half-to-even rounding in `_MM_FROUND_TO_NEAREST_INT` mode. It now uses the correct NEON intrinsic. This also applied to ppc.

- `_mm_cvtpd_pi32` /`_mm_cvtpd_epi32` wasn't doing half-to-even rounding. It now uses the corrected implementation of `simde_x_mm_round_pd` (moved over from `sse4.1.h`).

  Technically, the conversions (along with `_mm_cvtps_epi32`) should use the current rounding mode rather than specifically half-to-even, but it looked like always using half-to-even is intentional.

Adds tests that fail without each of the fixes.

### DIFF
--- a/test/x86/sse2.c
+++ b/test/x86/sse2.c
@@ -3279,10 +3279,6 @@ test_simde_mm_cvtpd_epi32 (SIMDE_MUNIT_TEST_ARGS) {
       { -INT32_C(           4),  INT32_C(          4) } },
     { { SIMDE_FLOAT64_C(-4.50), SIMDE_FLOAT64_C(4.50) },
       { -INT32_C(           4),  INT32_C(          4) } },
-    { { SIMDE_FLOAT64_C(-2147483650.0), SIMDE_FLOAT64_C(-2147483650.0) },
-      { INT32_MIN, INT32_MIN } },
-    { { SIMDE_FLOAT64_C( 2147483649.0), SIMDE_FLOAT64_C( 2147483649.0) },
-      { INT32_MIN, INT32_MIN } },
     #endif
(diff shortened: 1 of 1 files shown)


simde/x86/sse.h
```diff
@@ -3236,27 +3236,20 @@ simde_mm_min_ps (simde__m128 a, simde__m128 b) {
...
       r_.lsx_f32 = __lsx_vfmin_s(a_.lsx_f32, b_.lsx_f32);
-    #elif defined(SIMDE_VECTOR_SUBSCRIPT_OPS)
-      uint32_t SIMDE_VECTOR(16) m = HEDLEY_REINTERPRET_CAST(__typeof__(m), a_.f32 < b_.f32);
-      r_.f32 =
-        HEDLEY_REINTERPRET_CAST(
-          __typeof__(r_.f32),
-          ( (HEDLEY_REINTERPRET_CAST(__typeof__(m), a_.f32) &  m) |
-            (HEDLEY_REINTERPRET_CAST(__typeof__(m), b_.f32) & ~m)
-          )
-        );
```
mr-c: Must this be removed?

TheNumbat: Added these back

mr-c:
Thank you @TheNumbat !

PR #1373, opened 2026-01-21
test: Avoid -ffinite-math-only on floating point comparisons

iv-m:
With `-ffinite-math-only` (implied by `-ffast-math` and -Ofast), GCC considers all comparisons against infinities to be false and can optimize them away. This breaks several tests that rely on infinities to be compared correctly. To avoid this, we add GCC-specific optimize attribute that disables `-ffinite-math-only` optimization for floating point comarisions used in assertions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,12 +299,12 @@ jobs:
           distro: ubuntu-24.04
         - version: 10
           distro: ubuntu-24.04
-          arch_flags: -ffast-math
+          arch_flags: -Ofast
         - version: 11
           distro: ubuntu-24.04
         - version: 11
           distro: ubuntu-24.04
-          arch_flags: -ffast-math
+          arch_flags: -Ofast
         - version: 11
           distro:  ubuntu-24.04-arm
         - version: 11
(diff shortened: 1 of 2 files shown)


mr-c (approved):

mr-c (changes requested):
`-Ofast` produces errors with various GCC versions; perhaps they are compiler errors, perhaps they are issues that SIMDe can workaround.

Regardless, they need addressing before we merge; unless `-ffast-math` is sufficient?

iv-m:
The idea of commit `test: Avoid -ffinite-math-only on floating point comparisons` was to fix a very specific case of failing tests: when an assertion needs to compare two infinities (one returned from the function under test, and one from the expected response), but GCC assumed such comparison to be `false` because of `-ffinite-math-only`.

Since `-ffast-math` implies `-ffinite-math-only`,  `-ffast-math` is enough to trigger this failure mode. Also, since this commit addresses a specific issue triggered by  `-ffinite-math-only`, it probably won't help with any tests that are passing with `-ffast-math` but failing with `-Ofast` or when `-O3` is added.

What it can do is to allow to unskip some tests that are currently skipped when `SIMDE_FAST_MATH` is set, which seems to be quite a lot of tests.

mr-c:
@iv-m Great! Can you revert the change to use `-Ofast` where that failed, and then adjust the test skips where we can benefit?

iv-m:
> @iv-m Great! Can you revert the change to use `-Ofast` where that failed, and then adjust the test skips where we can benefit?

Well, I won't promise I'll review every entrance of `SIMDE_FAST_MATH` in the tests, but I think I can easily enable several tests (a couple of hundred `#if`s from my rough initial estimate) to demonstrate that this approach works.

And I think since this change affects only some GCC-like compilers, these tests should stay disabled for all the rest of the compilers, and so  another feature-test macro is needed for use in place of `SIMDE_FAST_MATH`. For example:

```c
#if defined(SIMDE_FAST_MATH)
#define SIMDE_TEST_EQUALTITY_WORKS_FOR_INFINITIES
#endif
#if defined(SIMDE_FAST_MATH) && defined(HEDLEY_GCC_VERSION) && HEDLEY_GCC_VERSION_CHECK(4,4,0)
/* For these GCC versions, `-ffinite-math-only` is disabled for
 * the functions that are used in assertions */
#define SIMDE_TEST_EQUALTITY_WORKS_FOR_INFINITIES
#endif
```

`SIMDE_TEST_EQUALTITY_WORKS_FOR_INFINITIES` is somewhat long name. Maybe a negative macro, e.g. `SIMDE_TEST_INFINITE_EQ_BROKEN`, is better?  Maybe you have better naming suggestions?

PR #1379, opened 2026-01-30
x86 avx: adjust rounding modes & other fixes

TheNumbat:
- `_mm256_cvtpd_epi32`/`_mm256_cvtps_epi32` didn't respect out-of-range inputs, and used a different rounding mode than their 128-bit counterparts. They now call the 128-bit versions when available, and otherwise use a C implementation copied from the 128-bit version.

- `_mm256_testnzc_si256` used `rc & rz`, meaning it returns `0` if they have non-overlapping set bits. `vptest` examines all bits, so this should be using a logical and. I've also slightly refactored the other `test_si256` implementations to call their 128-bit versions when available.

Adds tests that fail without the fixes.


mr-c:
Thank you @TheNumbat !

PR #1400, opened 2026-03-30
math: extend non-glibc guard to Clang and add exp10 guard

### LINKED ISSUE #1014: Several tests fail
See the log [here](https://people.freebsd.org/~yuri/simde-0.7.4-tests.log).

Version: 0.7.4
clang-15
FreeBSD 13.2

kjg0724:
Follow-up to #1398, addressing two remaining issues reported in #1396.

**1. Clang on non-AArch64 targets**

The previous fix guarded `__builtin_roundeven` for GCC 10+ only, based
on the assumption that Clang always lowers it to an inline
`llvm.roundeven.*` intrinsic. That is only true on AArch64 (FRINTN).
On x86 without SSE4.1, powerpc, sparc, i386, and others, Clang emits
a `roundeven()`/`roundevenf()` libm call — same failure mode as GCC.

Fix: drop the compiler-specific check and apply the glibc >= 2.25 guard
to all compilers. Non-glibc platforms fall through to the existing
portable inline fallback.

**2. `__builtin_exp10` / `__builtin_exp10f`**

`exp10()` is a GNU extension present only in glibc. On musl and other
non-glibc libcs the symbol is absent, producing the same class of link
error. The existing fallback (`simde_math_pow(10.0, v)`) is correct;
it just needs to be activated on non-glibc platforms.

Closes #1396.

### DIFF
--- a/simde/simde-math.h
+++ b/simde/simde-math.h
@@ -973,13 +973,18 @@ simde_math_fpclass(double v, const int imm8) {
   #endif
 #endif
 
-#if HEDLEY_HAS_BUILTIN(__builtin_exp10) ||  HEDLEY_GCC_VERSION_CHECK(3,4,0)
+/* __builtin_exp10 lowers to exp10() which is a GNU extension available
+ * only in glibc.  Other libcs (musl, OpenBSD, FreeBSD, MinGW, etc.)
+ * lack the symbol and produce a link error.  Fall back to pow(10, v). */
+#if (HEDLEY_HAS_BUILTIN(__builtin_exp10) || HEDLEY_GCC_VERSION_CHECK(3,4,0)) && \
+    defined(__GLIBC__)
   #  define simde_math_exp10(v) __builtin_exp10(v)
 #else
 #  define simde_math_exp10(v) simde_math_pow(10.0, (v))
 #endif
 
-#if HEDLEY_HAS_BUILTIN(__builtin_exp10f) ||  HEDLEY_GCC_VERSION_CHECK(3,4,0)
+#if (HEDLEY_HAS_BUILTIN(__builtin_exp10f) || HEDLEY_GCC_VERSION_CHECK(3,4,0)) && \
+    defined(__GLIBC__)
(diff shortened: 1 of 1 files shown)


brad0:
```
arm64$ cat /tmp/test.c
double foo1(double x) {
    return __builtin_roundeven(x);
}

float foo2(float x) {
    return __builtin_roundevenf(x);
}

long foo3(long x) {
    return __builtin_roundevenl(x);
}
arm64$ /usr/local/bin/clang-19 -target aarch64-unknown-openbsd -O2 -c test.c -o test.o
arm64$ nm -u test.o
arm64$ /usr/local/bin/clang-19 -target powerpc-unknown-openbsd -O2 -c test.c -o test.o
arm64$ nm -u test.o
roundeven
roundevenf
arm64$ /usr/local/bin/clang-19 -target powerpc64-unknown-openbsd -O2 -c test.c -o test.o
arm64$ nm -u test.o
.TOC.
roundeven
roundevenf
arm64$ /usr/local/bin/clang-19 -target mips64-unknown-openbsd -O2 -c test.c -o test.o
arm64$ nm -u test.o
roundeven
roundevenf
arm64$ /usr/local/bin/clang-21 -target amd64-unknown-openbsd -O2 -c test.c -o test.o
arm64$ nm -u test.o
roundeven
roundevenf

brad0:
Thanks. I'll take this for a spin and test on the various archs OpenBSD / FreeBSD.

brad0:
Build tested and tests run on OpenBSD/ FreeBSD amd64 with GCC and Clang. Build tests and tests run on OpenBSD aarch64 with Clang.

This also closes https://github.com/simd-everywhere/simde/issues/1014

mr-c (approved):
Thank you @kjg0724 !

mr-c:
@brad0 @kjg0724 PR(s) with additional CI configuration (non-Github Actions preferred) that would have caught some of these issues would be very welcome. 

@brad0 thank you for the manual testing!

PR #1415, opened 2026-07-18
x86: add mm512 epi8 compress intrinsics

fo40225:
#### Summary

avx512/compress.h: add `simde_mm512_maskz_compress_epi8` and `simde_mm512_mask_compressstoreu_epi8` (AVX512VBMI2)

Portable fallbacks included. On Zen 4 (`__znver4__`), `vpcompressb` with a memory destination is microcoded and extremely slow, so on that target `mask_compressstoreu_epi8` compresses to a register and then does a masked store, deriving the store mask via `_pext_u64` (BMI2 is implied by AVX512VBMI2-capable CPUs).

#### Tests

Tests for all three new functions are added in `test/x86/avx512/compress.c` and registered in the munit test lists.


mr-c:
Thank you @fo40225 !

PR #1418, opened 2026-07-18
x86/gfni: fix AltiVec paths on big-endian POWER

fo40225:
## Problem

The POWER AltiVec (P6 and P8) implementations of `simde_x_mm_gf2p8matrix_multiply_epi64_epi8` the helper used by the `gf2p8affine` / `gf2p8affineinv` family return wrong results on big-endian POWER. The code was correct only for little-endian.

## Root cause

Three endianness assumptions baked into vector constants and element indices:

1. **P8, `vec_perm` index tables** (`byte_interleave`, `byte_deinterleave`): `vec_perm` byte indices are subject to the compiler's little-endian adjustment, so the same table selects different bytes on big-endian.
   The BE variants are the reversed index vectors.
   `bit_select` is *not* affected: it feeds `vec_bperm`, whose bit numbering is big-endian on both targets, so it stays shared.

2. **P8, `vec_bperm` result position**: the gathered bits land in bits 48:63 of the vector unsigned-short element 4 in little-endian element numbering, but element 3 in big-endian.
   The `vec_splat` index must follow.

3. **P6, `byte_select` table**: `vec_sum2s` leaves the 8-bit gather in the low byte of words 1 and 3, i.e. byte offset 4/12 on little-endian but 7/15 on big-endian.

## Changes

`simde/x86/gfni.h` only: the three constants/indices above become endianness-specific via `SIMDE_ENDIAN_ORDER`.

No change on little-endian; no functional change to any other backend.

## Testing

Existing gfni munit tests exercise these paths; verified on ppc64 (big-endian) and ppc64le.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -439,6 +439,11 @@ jobs:
           arch_deb: ppc64el
           distro: ubuntu-24.04
         - version: 14
+          cross: power9be
+          arch_gnu: powerpc64
+          arch_deb: ppc64
+          distro: ubuntu-24.04
+        - version: 14
           cross: power7
           arch_gnu: powerpc64le
           arch_deb: ppc64el
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -440,6 +440,11 @@ jobs:
           arch_deb: ppc64el
           distro: ubuntu-24.04
         - version: 14
+          cross: power9be
+          arch_gnu: powerpc64
+          arch_deb: ppc64
+          distro: ubuntu-24.04
+        - version: 14
           cross: loongarch64
           arch_gnu: loongarch64
           arch_deb: loong64
--- a/docker/cross-files/power9be-gcc-14-ccache.cross
+++ b/docker/cross-files/power9be-gcc-14-ccache.cross
@@ -1,0 +1,18 @@
+[binaries]
+c = ['ccache', 'powerpc64-linux-gnu-gcc-14']
+cpp = ['ccache', 'powerpc64-linux-gnu-g++-14']
+ar = 'powerpc64-linux-gnu-ar'
+strip = 'powerpc64-linux-gnu-strip'
+objcopy = 'powerpc64-linux-gnu-objcopy'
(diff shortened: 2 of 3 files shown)


rosbif:
Good catch, congratulations for finding and correcting the problem.
I am Christopher Moore (aka rosbif) the principal author of gfni.h.
When testing my code I use the big-endian s390x to check for endian issues but (retrospectively) it is obvious that it cannot test for them in code paths for other platforms.
There are potentially problems with ARM too which theoretically is (or used to be ?) bi-endian.
However the only big-endian ARM system I have encountered is the venerable Intel IXP420 in the Linksys NSLU2 (aka slug).
I haven't heard of any big-endian ARM system with NEON.
Does anyone know of one ?

fo40225:
The project homepage does not clarify whether bi-endian architectures are supported, or if both little-endian (LE) and big-endian (BE) are fully covered (https://github.com/simd-everywhere/simde#hardware).

In my opinion, if there is a real user demand for an architecture, it should be added to the CI testing.

Endianness configurations without CI tests should be explicitly marked as "untested" in the documentation to prevent the test matrix from bloating with unused setups.

Possible bi-endian candidates include:

- ARM: aarch64 / aarch64_be, arm / armeb

- PowerPC: powerpc64le / powerpc64, powerpcle (theoretical) / powerpc

- MIPS: mips64el / mips64, mipsel / mips

While the RISC-V specification includes bi-endian support, practically no silicon or toolchain currently supports it?

Regarding the endian-safety of scalar implementations: if they have already been tested on x86-64 (LE) + aarch64 (LE) and s390x (BE), both endian types should theoretically be covered.

mr-c (commented):
Dear @fo40225 ; thank you for your contribution including the big-endian Power9 CI job.

Do you want to fix the other big-endian errors seen in the CI results, or just skip the power* implementations for those function on big-endian systems and fallback to the portable code?

https://github.com/simd-everywhere/simde/actions/runs/31862512619/job/94958281139?pr=1418#step:10:62

fo40225:
Too many functions need to be modified. Bi-endian support should be a separate PR, if anyone actually needs it.

PR #1429, opened 2026-08-16
Enhance simde/x86/gfni.h

fo40225:
#### Summary
This change makes the GFNI emulation faster.
The changed files are `simde/x86/gfni.h` and `test/x86/gfni.c`.

#### New feature detection.

The macro `SIMDE_X_GFNI_HAVE_AES` detects hardware AES support.
The support comes from AES-NI on x86 or the crypto extension on ARMv8.
The macro `SIMDE_X_GFNI_HAVE_SHUFFLE` detects byte-shuffle support.

#### New AES path.

When the hardware has AES, the function `simde_x_mm_gf2p8inverse_epi8` uses the AES S-box to compute the GF(2⁸) inverse.
New lookup tables undo the ShiftRows and affine steps of AES.
When the matrix is A_aes and the constant is 0x63, the affineinv operation becomes one SubBytes operation.

#### New tower-field path.

When the hardware has shuffle but no AES, the code computes the inverse and the multiply in the tower field GF((2⁴)²).
New lookup tables hold the log, exp, square, and basis-change values of GF(2⁴).
The multiply uses the Karatsuba method over GF(2⁴).

#### New constant-matrix path.

When the matrix is a compile-time constant, the code turns the matrix into two nibble lookup tables.
The affine transform then becomes two shuffles and one XOR at run time.
The new 256-entry table `simde_x_gf2p8_mul_matrix_lut` turns a multiply by a GF(2⁸) constant into an affine transform.
This path needs the two conditions `SIMDE_CHECK_CONSTANT_` and `SIMDE_X_GFNI_HAVE_SHUFFLE`.
Without shuffle, the matrix multiply uses a slower movemask loop.
In that condition, the usual schoolbook multiply is faster.
Thus the code does not use the constant-multiply path without shuffle.

#### Removed old code.

This change removes the old lookup-table inverse implementations for SSE4.1, AVX2, and AVX512BW; the new methods are faster than a large table lookup.
CPUs that support AVX512 but not GFNI become much slower with 512-bit operations; thus the 512-bit functions now use the 256-bit fallback path.

#### New tests.
The file `test/x86/gfni.c` gets new test functions.

### DIFF
--- a/test/x86/gfni.c
+++ b/test/x86/gfni.c
@@ -5070,6 +5070,30 @@ test_simde_mm_gf2p8mul_epi8(SIMDE_MUNIT_TEST_ARGS) {
 }
 
 static int
+test_simde_mm_gf2p8mul_epi8_const(SIMDE_MUNIT_TEST_ARGS) {
+  /* Multiplying by a compile-time-constant broadcast degenerates to an affine
+   * transform; check it matches the run-time (general) multiply. */
+  #define SIMDE_TEST_GFNI_MUL_CONST(K) \
+    do { \
+      volatile int vk_ = (K); \
+      simde__m128i rc = simde_mm_gf2p8mul_epi8(a, simde_mm_set1_epi8(HEDLEY_STATIC_CAST(int8_t, (K)))); \
+      simde__m128i rr = simde_mm_gf2p8mul_epi8(a, simde_mm_set1_epi8(HEDLEY_STATIC_CAST(int8_t, vk_))); \
+      simde_assert_m128i_i8(rc, ==, rr); \
+    } while (0)
+
+  for (int t = 0 ; t < 16 ; t++) {
+    simde__m128i a = simde_test_x86_random_i8x16();
+    SIMDE_TEST_GFNI_MUL_CONST(0x00); SIMDE_TEST_GFNI_MUL_CONST(0x01);
+    SIMDE_TEST_GFNI_MUL_CONST(0x02); SIMDE_TEST_GFNI_MUL_CONST(0x1b);
+    SIMDE_TEST_GFNI_MUL_CONST(0x53); SIMDE_TEST_GFNI_MUL_CONST(0x57);
+    SIMDE_TEST_GFNI_MUL_CONST(0x80); SIMDE_TEST_GFNI_MUL_CONST(0xff);
+  }
+
+  #undef SIMDE_TEST_GFNI_MUL_CONST
+  return 0;
+}
+
+static int
 test_simde_mm256_gf2p8mul_epi8(SIMDE_MUNIT_TEST_ARGS) {
   const struct {
     simde__m256i a;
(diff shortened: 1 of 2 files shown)


fo40225:
@rosbif

test/x86/gfni.c
```diff
@@ -5044,6 +5094,30 @@ test_simde_mm_gf2p8mul_epi8(SIMDE_MUNIT_TEST_ARGS) {
...
+      simde__m128i rr = simde_mm_gf2p8mul_epi8(a, simde_mm_set1_epi8(HEDLEY_STATIC_CAST(int8_t, vk_))); \
+      simde_assert_m128i_i8(rc, ==, rr); \
+    } while (0)
+
+  for (int t = 0 ; t < 16 ; t++) {
+    simde__m128i a = simde_test_x86_random_i8x16();
+    SIMDE_TEST_GFNI_MUL_CONST(0x00); SIMDE_TEST_GFNI_MUL_CONST(0x01);
+    SIMDE_TEST_GFNI_MUL_CONST(0x02); SIMDE_TEST_GFNI_MUL_CONST(0x1b);
+    SIMDE_TEST_GFNI_MUL_CONST(0x53); SIMDE_TEST_GFNI_MUL_CONST(0x57);
+    SIMDE_TEST_GFNI_MUL_CONST(0x80); SIMDE_TEST_GFNI_MUL_CONST(0xff);
```
mr-c, line 5116: > `warning C4309: 'static_cast': truncation of constant value`

https://ci.appveyor.com/project/nemequ/simde/builds/54552506/job/hlmksf28wd24vg2y#L1168

mr-c:
Thank you for this PR @fo40225 

> This change makes the GFNI emulation faster.

Do you have timing data to share? Can you explain a bit about your use case and motivation? We love to hear how people are using SIMDe, it is helpful to us!

fo40225:
### SIMDe GFNI old vs new speedup (spd = old ÷ new, >1 = new is faster)

**With GFNI hardware** (rocketlake, alderlake, raptorlake, meteorlake, znver4):
both versions compile to a single native instruction **all 1.00×**, no difference.

**Without GFNI hardware (emulated path):**

| Op | Real x86 Zen4 (all march) | Real x86 Nehalem | ARM X925 (big core) | ARM A725 (little core) |
|---|--:|--:|--:|--:|
| mul var | 0.92~1.09× | 1.15× | ~1.00× | ~1.00× |
| mul const | **8.1~9.5×** | **6.7×** | ~1.00× | 1.33× |
| affine var | ~1.00× | ~1.00× | ~1.00× | ~1.00× |
| affine const | **9.7~10.4×** | **7.0×** | **4.75×** | **9.50×** |
| affineinv var | 1.1× (no AES) / 1.7~1.9× (AES) | 1.57× | 1.43× | 0.66× |
| affineinv const | 2.1~2.4× (no AES) / **~30×** (AES) | 2.48× | **2.58×** | 1.49× |

**Summary:** the new version wins big on **constant-operand paths** (folded at compile
time into nibble shuffles / AES fusion); variable paths are equal or slightly faster.
The only regression is affineinv var on the ARM A725 little core (0.66×, where a LUT
beats the tower-field approach on narrow cores); the tower implementation was kept
anyway no special case added for a single little core.

fo40225:
I am trying to write a faster zlib compression and decompression library based on GFNI instructions (although it haven't succeeded yet).

Recent PRs was just a quick fix for issues I ran into while using simde.

rosbif (commented):
Sorry for this late reply.
These improvements seem excellent ideas to me, congratulations.
I only regret not having found them myself ;-).
I am away on holiday so I cannot easily test them but obviously I approve these improvements.

PR #1441, opened 2026-09-04
gh-actions gcc-16: ubuntu 26.04 with newer qemu (10.2.1+)

mr-c:
- **gh-actions gcc-16: ubuntu 26.04 with newer qemu (10.2.1+)**
- **gcc-16 -O3 silence uninitialized warning**


mr-c:
Manual trigger of the gcc-16 workflow is at https://github.com/simd-everywhere/simde/actions/runs/33846529019

PR #1450, opened 2026-09-14
wasm: use vqtbl1q_s8 for swizzle on A64V8 for SIMD and relaxed SIMD

jawj:
This patch has `simde_wasm_i8x16_swizzle` and `simde_wasm_i8x16_relaxed_swizzle` use `vqtbl1q_s8` for `SIMDE_ARM_NEON_A64V8_NATIVE`.

I haven't done comprehensive benchmarking, but some arguments in favour:

* It compiles to one instruction, not several (see below).
* There's a precedent in SIMDe: `_mm_shuffle_epi8` in ssse3.h already uses `vqtbl1q_s8`.
* Results are identical: the native tests pass in both C and C++, and the instruction returns 0 for indices ≥ 16, as WASM's swizzle requires.
* Real-world performance: some hex encoding/decoding functions with wasm2c on my M3 Pro Mac show 25 – 45% higher throughput (that's the whole functions: the swizzle bump is presumably larger).

In terms of what's emitted, here's Apple Clang 17.0.0 (clang-1700.6.4.2) with `-O3` before —

```
       0:      	tbl.8b	v2, { v0 }, v1
       4:      	ext.16b	v1, v1, v1, #0x8
       8:      	tbl.8b	v0, { v0 }, v1
       c:      	mov.d	v2[1], v0[0]
      10:      	mov.16b	v0, v2
      14:      	ret
```

— and after —

```
       0:      	tbl.16b	v0, { v0 }, v1
       4:      	ret
```


mr-c (approved):
LGTM; this is also what v8 does; thank you @jawj !

https://github.com/v8/v8/blob/9f9b3ebe0a8137c2a344d1a84536068231af632f/src/compiler/backend/arm/code-generator-arm.cc#L3356

and likewise for wasmtime (used by Firefox) https://github.com/bytecodealliance/wasmtime/blob/817c58787f432bcdbbb87679011f72c5bc80dbda/winch/codegen/src/isa/aarch64/masm.rs#L1533

